PR #1, opened 2015-04-14
Modify existing kwargs dict instead of creating a new one

jasongrout:
Creating a new dict throws away any kwargs passed into the function.  For example, the working directory for the kernel is passed in, and then is ignored if we create a new kwargs dict.  The end result is that we can't set the current working directory for a kernel.


minrk:
Thanks! Not sure how I did that one.

jasongrout:
I bet you've had a *lot* of code flying around the last few weeks.

PR #3, opened 2015-04-15
Make print function python2-compatible by importing from __future__

jasongrout:
Without this, we get a syntax error on python2 in the print around line 111.


Carreau:
Thanks.

PR #8, opened 2015-04-29
Add prefix to names that get formatted into kernel cmd

takluyver:
This makes it easier to make a portable Jupyter+kernels setup, because you can specify executable paths for kernels relative to the parent process' prefix, e.g. `argv: ['{prefix}/bin/R', ...]`

Closes jupyter/jupyter_notebook#53

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -166,7 +166,9 @@ class KernelManager(ConnectionFileMixin):
         else:
             cmd = self.kernel_spec.argv + extra_arguments
 
-        ns = dict(connection_file=self.connection_file)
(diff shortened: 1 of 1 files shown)


Carreau:
should we then also re-consider expanding `~` and other things like that ? Should we document that somewhere ?

takluyver:
Given that this mechanism is already in place, I think I'd support it as `{home}` rather than `~`. I'm happy to consider it, but I think this set of names should be pretty minimal, with each entry justified.

It should absolutely be documented. We're still considering what documentation should live where, though.

Carreau:
fair enough. lets try that.

PR #10, opened 2015-04-30
remove SYSTEM_KERNEL_DIRS

minrk:
use SYSTEM_JUPYTER_PATH from core instead

sibling to jupyter/jupyter_notebook#64

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -1,8 +1,12 @@
+"""Tools for managing kernel specs"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 import io
 import json
 import os
 import shutil
-import sys
 import warnings
 
 pjoin = os.path.join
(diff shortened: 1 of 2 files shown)


jupyter_client/kernelspec.py
```diff
@@ -136,8 +122,8 @@ def _get_destination_dir(self, kernel_name, user=False):
         if user:
             return os.path.join(self.user_kernel_dir, kernel_name)
         else:
-            if SYSTEM_KERNEL_DIRS:
-                return os.path.join(SYSTEM_KERNEL_DIRS[-1], kernel_name)
+            if SYSTEM_JUPYTER_PATH:
+                return os.path.join(SYSTEM_JUPYTER_PATH[0], 'kernels', kernel_name)
             else:
                 raise EnvironmentError("No system kernel directory is available")
```
takluyver: This case shouldn't be necessary now, right? We're ensuring that there's always at least 1 path in `SYSTEM_JUPYTER_PATH`.

minrk: Right.

Carreau:
same, check for `SYSTEM_KERNEL_DIRS` and warn if set ?

minrk:
same as before - how would it be set? It's a constant defined in this module.

PR #17, opened 2015-05-13
fix inverted kernelspec priority

minrk:
kernel_dis is a search path, which means that the first items have highest priority. The priority of the kernel loading was backward, prioritizing system kernels over user kernels.


Carreau:
Make sens.

PR #21, opened 2015-05-21
localization fixes for local IP detection

minrk:
- prefer `ip` to `ifconfig`
- use regex to reduce localization sensitivity in ifconfig output

### DIFF
--- a/jupyter_client/localinterfaces.py
+++ b/jupyter_client/localinterfaces.py
@@ -225,11 +225,11 @@ def _load_ips(suppress_exceptions=True):
                 pass
         else:
             try:
-                return _load_ips_ifconfig()
+                return _load_ips_ip()
             except (IOError, NoIPAddresses):
                 pass
             try:
-                return _load_ips_ip()
+                return _load_ips_ifconfig()
             except (IOError, NoIPAddresses):
                 pass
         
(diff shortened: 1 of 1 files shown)


jupyter_client/localinterfaces.py
```diff
@@ -95,18 +97,15 @@ def _load_ips_ifconfig():
     lines = out.splitlines()
     addrs = []
     for line in lines:
-        blocks = line.lower().split()
-        if (len(blocks) >= 2) and (blocks[0] == 'inet'):
-            if  blocks[1].startswith("addr:"):
-                addrs.append(blocks[1].split(":")[1])
-            else:
-                addrs.append(blocks[1])
+        m = _ifconfig_ipv4_pat.match(line.strip())
```
takluyver: Do we want to rely on the ip address being the first match for this pattern? When I run `ifconfig`, I get the IP address, broadcast address and subnet mask on the same line - all of them in the same N.N.N.N format.

minrk: I believe it is reliably the first item.

takluyver: OK, will merge tomorrow unless anyone else wants to review.

PR #22, opened 2015-05-27
Make a start on docs

takluyver:
Narrative docs moved in from IPython repo and updated, API docs created anew.


takluyver:
Docs now up at https://jupyter-client.readthedocs.org/

rgbkrk:
This is great. Thank you so much @takluyver.

docs/api/client.rst
```diff
@@ -0,0 +1,31 @@
+client - communicating with kernels
+===================================
+
+.. module:: jupyter_client.client
```
minrk: I would put this as top-level in `jupyter_client`, so it's documented as `jupyter_client.KernelClient`. Same for Manager.

takluyver: Done in #23.

PR #26, opened 2015-05-29
always replace existing installations with `kernelspec.install`

minrk:
deprecate replace arg with a DeprecationWarning that nobody will ever see...

Also return the install destination for use by callers.

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -162,6 +162,8 @@ class KernelSpecManager(LoggingConfigurable):
             shutil.rmtree(destination)
 
         shutil.copytree(source_dir, destination)
+        self.log.info('Installed kernelspec %s in %s', kernel_name, destination)
+        return destination
 
     def install_native_kernel_spec(self, user=False):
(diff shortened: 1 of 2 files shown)


takluyver:
Seems fine

takluyver:
There's a test that OSError is raised when you try to install a kernelspec that already exists.

jupyter_client/kernelspec.py
```diff
@@ -143,22 +143,27 @@ def install_kernel_spec(self, source_dir, kernel_name=None, user=False,
...
         if not kernel_name:
             kernel_name = os.path.basename(source_dir)
         kernel_name = kernel_name.lower()
-
+        
+        if replace is not None:
+            warnings.warn(
+                "replace is ignored. Installing a kernelspec always replaces an existing installation",
+                DeprecationWarning,
+            )
```
Carreau: stacklevel=2 to actually have the error message pointing at the function call ?

Carreau:
Test need fixing. 

+1

minrk:
test should be fixed, warning stack level changed

PR #28, opened 2015-06-10
Don't allow PYTHONEXECUTABLE env to be inherited by kernel process

minrk:
it can break Python.

Server in one conda env with kernel in another illustrates this problem.

Symptomatic kernel output:

```
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
ImportError: No module named site
```


takluyver:
[Huh](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONEXECUTABLE). I didn't even know that was a thing. I think this makes sense.

PR #29, opened 2015-06-11
match pep 440 version scheme

Carreau:


minrk:
I'm :-1: on that, actually.

Carreau:
Any reason why ? Seem like following a standard might be a good thing...

minrk:
If we are doing it properly, we should be revving the number on every build / install, not just using 0 all the time. This part of PEP 440 bugs me a lot, actually. I think it was a mistake.

minrk:
Also, I assume you are doing this because of the message from pip 6:

> UserWarning: The version specified requires normalization, consider using '4.0.0.dev0' instead of '4.0.0-dev'.

They realized this message was a bad idea, and no longer show it in pip 7.

Carreau:
Ok,fair enough. I still feel sad that we don't match the pep. I'm not sure where in the pep is stated that we should update the N on each build. I see that more as a number if we had nightly build

PR #30, opened 2015-06-13
add ignored profile kwarg

minrk:
and warn about it if specified. For signature compatibility with pre-4.0.


jupyter_client/connect.py
```diff
@@ -158,6 +155,8 @@ def find_connection_file(filename='kernel-*.json', path=None):
     -------
     str : The absolute path of the connection file.
     """
+    if profile is not None:
+        warnings.warn("Jupyter has no profiles. profile=%s has been ignored." % profile)
```
takluyver: I'd probably use `%r`, so that it shows up as `profile='foo'` rather than `profile=foo`.

takluyver: Oh, I missed that @Carreau already merged this. Never mind.

Issue #35, opened 2015-06-24
kernelspec kernel.json from deleted virtualenv will be used leading to kernel error


tonysimpson:
Steps to reproduce
-----
 1. Install jupyter in a python virtualenv
 2. start jupyter notebook and create a new python notebook
 3. stop jupyter and delete the virtualenv
 4. create a new virtualenv with a different name to 1. and install jupyter
 5. start jupyter and create a python notebook or open the one from 2.
 6. "Kernel Error" appears in notebook

Notes
-----
I think this is probably WIP but I can't find any documentation on how it's supposed to work, closest I can find is https://ipython.org/ipython-doc/dev/development/kernels.html#kernelspecs but it's unclear if in the future we should manually configure and manage kernelspecs or some part of jupyter will do this for us. I would be happy to help fix this if the expected behaviour was specified.

Example kernel.json
-----

```json
{
 "language": "python",
 "display_name": "Python 3",
 "argv": [
  "C:\\Users\\tsimpson\\Envs\\jupyter\\Scripts\\python.exe",
  "-m",
  "ipykernel",
  "-f",
  "{connection_file}"
 ]
}
```

Workaround
-----
You can fix this by manually deleting the kernel config director containing the offending kernel.json and jupyter will then regenerate one using the current virtualenv.

System info
=========

```python
{'commit_hash': 'b754d5e',
 'commit_source': 'repository',
 'default_encoding': 'cp850',
 'ipython_path': 'c:\\users\\tsimpson\\envs\\jupyter\\src\\ipython\\IPython',
 'ipython_version': '4.0.0-dev',
 'os_name': 'nt',
 'platform': 'Windows-8-6.2.9200',
 'sys_executable': 'C:\\Users\\tsimpson\\Envs\\jupyter\\Scripts\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '3.4.1 (v3.4.1:c0e311e010fc, May 18 2014, 10:45:13) [MSC '
                'v.1600 64 bit (AMD64)]'}
```

takluyver:
This is a problem because in the Jupyter split, we switched from adding the native kernelspec to the list internally, to installing it if it doesn't exist. That assumes that the interpreter with which you first run the notebook is the standard Python interpreter you want. I don't think that's an assumption we should be making.

minrk:
@takluyver do you mean we shouldn't register a Python kernelspec? That the IPython kernelspec should only be installed when explicitly requested by the user?

[...]

PR #37, opened 2015-06-24
Make the native kernel available without needing to install it

takluyver:
Closes gh-35, in concert with jupyter/notebook#168


minrk:
:+1:

PR #42, opened 2015-07-03
allow installing kernelspecs with a prefix

minrk:
Makes it a little easier to put kernelspecs in virtualenvs or conda envs, and build packages for kernels.

### DIFF
--- a/jupyter_client/tests/test_kernelspec.py
+++ b/jupyter_client/tests/test_kernelspec.py
@@ -9,7 +9,12 @@ try:
     from unittest.mock import patch
 except ImportError:
     from mock import patch
-    
+
+if str is bytes: # py2
+    StringIO = io.BytesIO
+else:
+    StringIO = io.StringIO
+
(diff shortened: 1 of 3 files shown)


jupyter_client/tests/test_kernelspec.py
```diff
@@ -83,6 +85,35 @@ def test_install_kernel_spec(self):
...
+        self.addCleanup(td.cleanup)
+        capture = io.StringIO()
+        handler = StreamHandler(capture)
+        self.ksm.log.addHandler(handler)
+        self.ksm.install_kernel_spec(self.installable_kernel,
+                                     kernel_name='tstinstalled',
+                                     prefix=td.name)
+        captured = capture.getvalue()
+        self.ksm.log.removeHandler(handler)
+        self.assertIn("may not be found", captured)
```
takluyver: This assert is failing on Python 2.

minrk: Python 2 appeased

Issue #31, opened 2015-06-16
Provide JSON output for kernelspec lists


rgbkrk:
Similar to `jupyter --paths --json` from jupyter_core, it would be great to get the full set of kernelspecs using:

```
jupyter kernelspecs list --json
```

or something similar. This is for the use case of alternative frontends, especially those written in other languages. The driving use case is those building Atom packages or Electron plugins which is JavaScript all the way down.

/cc @willwhitney @karissa

minrk:
And I presume a similar static-path output on jupyter/notebook?

rgbkrk:
> And I presume a similar static-path output on jupyter/notebook?

I'll let someone else chime in there, as I think those should be packaged on `bower` or `npm`. It couldn't hurt to have a static path output for the notebook, in fact that would be really helpful for our own CDN use across try and the releases.

minrk:
Ah, that makes sense (bower at least, npm less so).

takluyver:
This now exists

PR #44, opened 2015-07-03
provide a way to output kernelspec as json

Carreau:
Path toward fixing #31

### DIFF
--- a/jupyter_client/kernelspecapp.py
+++ b/jupyter_client/kernelspecapp.py
@@ -32,19 +32,15 @@ class ListKernelSpecs(JupyterApp):
         return KernelSpecManager(parent=self, data_dir=self.data_dir)
 
     def start(self):
-        specs = sorted(self.kernel_spec_manager.find_kernel_specs().items())
+        specs = self.kernel_spec_manager.find_kernel_specs()
         if not self.json_output:
             print("Available kernels:")
-            for kernelname, spec in specs:
+            for kernelname, spec in sorted(specs.items()):
                 print("  %s" % kernelname)
         else:
             print(json.dumps({
-                "kernelspecs":
-                    [{'name':k, 'path':v} for k,v in specs]
-                },
-                indent=2
-                )
-            )
+                'kernelspecs': specs
+            }, indent=2))
 
 
 
--- a/jupyter_client/kernelspecapp.py
+++ b/jupyter_client/kernelspecapp.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import errno
 import os.path
 import sys
+import json
 
 from traitlets.config.application import Application
 from jupyter_core.application import (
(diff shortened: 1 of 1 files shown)


jupyter_client/kernelspecapp.py
```diff
@@ -19,14 +19,28 @@
...
 
     def start(self):
-        print("Available kernels:")
-        for kernelname in sorted(self.kernel_spec_manager.find_kernel_specs()):
-            print("  %s" % kernelname)
+        specs = sorted(self.kernel_spec_manager.find_kernel_specs().items())
+        if not self.json_output:
+            print("Available kernels:")
+            for kernelname,spec in specs:
+                print("  %s, %s" % (kernelname, spec))
```
minrk: I'm not sure printing the spec in this case makes sense, does it?

Carreau: Not untrue. Fixed.

jupyter_client/kernelspecapp.py
```diff
@@ -19,14 +19,28 @@
...
-        print("Available kernels:")
-        for kernelname in sorted(self.kernel_spec_manager.find_kernel_specs()):
-            print("  %s" % kernelname)
+        specs = sorted(self.kernel_spec_manager.find_kernel_specs().items())
+        if not self.json_output:
+            print("Available kernels:")
+            for kernelname, spec in specs:
+                print("  %s" % kernelname)
+        else:
+            import json
```
minrk: import json at module level

jupyter_client/kernelspecapp.py
```diff
@@ -60,6 +74,8 @@ def _kernel_name_default(self):
                 "Install to the per-user kernel registry"),
              'replace': ({'InstallKernelSpec': {'replace': True}},
                 "Replace any existing kernel spec with this name."),
+             'replace': ({'InstallKernelSpec': {'replace': True}},
```
minrk: did you mean to duplicate replace?

Carreau: WTF ??

Carreau:
Side question, do we want array as top level ? or [wrap it in an object](http://stackoverflow.com/questions/3833299/can-an-array-be-top-level-json-text)

minrk:
Wrapping it in an object with one key `kernelspecs` is probably a good idea. It will give us a better way to add new information without breaking things, if we find any we should add in the future.

Carreau:
Done. 

We can maybe also still change "find_kernel_spec()" api to be consistent with that. so that we just have to `json.dumps({kernelspecs:find_kernel_specs()})`, or do we stick with find_kernel_spec api of 3.x ?

minrk:
If anything, I probably like the find_kernel_specs returning a dict better, so maybe it should do that.

jupyter_client/kernelspecapp.py
```diff
@@ -19,14 +20,32 @@
...
-            print("  %s" % kernelname)
+        specs = sorted(self.kernel_spec_manager.find_kernel_specs().items())
+        if not self.json_output:
+            print("Available kernels:")
+            for kernelname, spec in specs:
+                print("  %s" % kernelname)
+        else:
+            print(json.dumps({
+                "kernelspecs":
+                    [{'name':k, 'path':v} for k,v in specs]
```
takluyver: Should this also include the info from the kernelspecs? In the particular case of the native kernelspec, the `kernel.json` file may not actually be present in the resource directory, because that dictionary is generated without writing it somewhere. Alternatively, we could do something like writing it to a temporary directory when we look it up.

Carreau:
> If anything, I probably like the find_kernel_specs returning a dict better, so maybe it should do that.

it does already, but not the same way it's `{name:spec}` digging around I think it is painful to change the lower lever API now, and I dont' knowhow much it is used. So I change to that : 

```bash
~/d/jupyter_client (master) $ jupyter kernelspec list
Available kernels:
  julia-0.3
  julia-0.4
  python3

~/d/jupyter_client (master) $ jupyter kernelspec list --json
{
  "kernelspecs": {
    "python3": "/usr/local/share/jupyter/kernels/python3",
    "julia-0.4": "/Users/bussonniermatthias/.ipython/kernels/julia-0.4",
    "julia-0.3": "/Users/bussonniermatthias/.ipython/kernels/julia-0.3"
  }
}
```

minrk:
Thanks, I think that works.

takluyver:
I think that if the special case for the native kernelspec kicks in (i.e. if there isn't a kernel installed called python3 or python2 as appropriate), the code that called it will run into problems when it tries to load kernel.json for the native kernel.

PR #48, opened 2015-07-11
add kernel spec to json kernelspec list output

emilienschultz:
pure json display of kernels informations

### DIFF
--- a/jupyter_client/kernelspecapp.py
+++ b/jupyter_client/kernelspecapp.py
@@ -33,7 +33,10 @@ class ListKernelSpecs(JupyterApp):
 
     def start(self):
         specs = self.kernel_spec_manager.find_kernel_specs()
-        specs = {kname:{"dir":specs[kname],"info":self.kernel_spec_manager.get_kernel_spec(kname).to_dict()} for kname in specs}
+        specs = {kname: {
+                "resources_dir": specs[kname],
+                "spec": self.kernel_spec_manager.get_kernel_spec(kname).to_dict() 
+            } for kname in specs}
         if not self.json_output:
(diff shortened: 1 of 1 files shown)


Carreau:
+1

jupyter_client/kernelspecapp.py
```diff
@@ -33,6 +33,7 @@ def _kernel_spec_manager_default(self):
 
     def start(self):
         specs = self.kernel_spec_manager.find_kernel_specs()
+        specs = {kname:{"dir":specs[kname],"info":self.kernel_spec_manager.get_kernel_spec(kname).to_dict()} for kname in specs}
```
takluyver: We're not strict about keeping lines to 80 characters, but I think this is a bit too long. Can you split it onto two (or more) lines, and add a few spaces around the punctuation to make it  a bit more readable?

Carreau: FYI, pep8 we where talking about this morning, that might be relevant: https://www.python.org/dev/peps/pep-0008/

minrk:
To be more consistent with how we present this info elsewhere, can we call `info` `spec` and `dir` `resources_dir`?

emilienschultz:
I made the changes suggested

takluyver:
:+1: @minrk, I'll let you take another quick look before we merge.

Carreau:
+1

minrk:
Huzzah!

Carreau:
images

emilienschultz:
youhou !

SylvainCorlay:
bravo @emilienschultz

PR #49, opened 2015-07-11
report actual version in `kernelspec --version`

minrk:


damianavila:
LGTM

takluyver:
Presumably `jupyter nbextension` and `jupyter nbextension install` need the same change? Are there any others?

minrk:
Roughly all of the split-out applications need this. I'm making my way around.

PR #55, opened 2015-07-23
add `blocking_client` method to KernelManager/Class

minrk:
allows easy creation of blocking sibling, which can be useful when deferring to the eventloop is undesirable (e.g. qtconsole is_complete).

see jupyter/qtconsole#33


minrk:
Merging in the next day or so, unless I hear otherwise.

PR #57, opened 2015-08-03
Updates instructions for simple wrapper kernel

basheersubei:
migrated from https://github.com/ipython/ipython/pull/8684

### DIFF
--- a/docs/wrapperkernels.rst
+++ b/docs/wrapperkernels.rst
@@ -68,6 +68,8 @@ To launch your kernel, add this at the end of your module::
         from ipyernel.kernelapp import IPKernelApp
         IPKernelApp.launch_instance(kernel_class=MyKernel)
 
+Now create a `JSON kernel spec file <http://jupyter-client.readthedocs.org/en/latest/kernels.html#kernel-specs>`_ and install it using ``jupyter kernelspec install </path/to/kernel>``. Place your kernel module anywhere Python can import it (try current directory for testing). Finally, you can run your kernel using ``jupyter console --kernel <mykernelname>``. Note that ``<mykernelname>`` in the below example is ``echo``. 
+
 Example
 -------
 


docs/wrapperkernels.rst
```diff
@@ -68,6 +68,8 @@ To launch your kernel, add this at the end of your module::
         from ipyernel.kernelapp import IPKernelApp
         IPKernelApp.launch_instance(kernel_class=MyKernel)
 
+Place your kernel anywhere that you could import it (e.g. current directory). Now create a JSON kernel spec file (see below for example), and place it in ``~/.ipython/kernels/<mykernelname>/``. Now, you can run your kernel using ``ipython console --kernel <mykernelname>``
```
takluyver: The location to put it in now [depends on the system](http://jupyter.readthedocs.org/en/latest/system.html#data-files). People could use the `jupyter kernelspec install` command to put it in the right place.

Also, `ipython console` -> `jupyter console`

basheersubei:
Is this better now?

takluyver:
Thanks. The mention of `~/.ipython/kernels/<mykernelname>` should be gone, though.

basheersubei:
Oh, so the JSON file should be placed in the same directory as the kernel module file when the kernel is installed (using kernelspec install)?

basheersubei:
fixed again and squash previous commits

takluyver:
Not quite - the kernel module needs to be in a location where Python can import it - the cwd works, so there's no need to copy it anywhere for testing. The kernel spec (the directory containing kernel.json) needs to be in one of those data files locations. You can add a ```:ref:`kernelspecs` ``` to link to [this info](http://jupyter-client.readthedocs.org/en/latest/kernels.html#kernel-specs) about kernel spec.s

basheersubei:
oh, sorry I had no idea there was more documentation. I'll make the changes.

takluyver:
No worries, thanks for your patience on this.

I think it's still a bit confusing. The `jupyter kernelspec` commands are all about the kernel specs - the metadata directories that contain the `kernel.json` file. `jupyter kernelspec install` is to get that in the right place, not to install the module. The `kernelspec` machinery doesn't know about Python modules at all.

So can you move the bit about the `jupyter kernelspec install` command into the previous sentence about kernel specs?

basheersubei:
ah, I see. I misunderstood earlier (should've read the new docs). Is it good now?

takluyver:
Thanks, that's looking good. Can you just add a space before 'Finally', and then I'll merge.

basheersubei:
done

takluyver:
Thanks!

Issue #56, opened 2015-07-31
KernelSpecManager seems not to be fully Configurable


danielballan:
I am working on [a "conda kernels" project](https://github.com/danielballan/conda-kernels) that overrides the `KernelSpecManager`. When I run the notebook like so:

```
ipython notebook --NotebookApp.kernel_spec_manager_class=conda_kernels.CondaKernelSpecManager
```

a flurry of print statements has convinced me that my custom subclass (`CondaKernelSpecManager`) is being called most of the time but the built-in `KernelSpecManager` is still being called in at least one case (specifically, [here](https://github.com/ipython/ipython/blob/3.x/IPython/html/services/kernelspecs/handlers.py#L49)). I suspect that [this line](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L62) should have been changed as part of ipython/ipython#6969.

Does that seem plausible?

I'm developing this on IPython 3.2 while the dust settles from the big split. If you think things would go easier using jupyter_client, I'll switch over. Thanks.

takluyver:
The help message added in ipython/ipython#6969 says classes you provide as kernel spec managers need to inherit from our KSM class, so `Instance(kernelspec.KernelSpecManager)` is still fine.

The notebook app class creates `k_s_m` using `k_s_m_class`, and then instantiates the kernel manager with that. MultiKernelManager appears to take care of passing the value down to the individual kernel managers. `k_s_m` is also passed to the tornado webapp settings, which is where the handler property you highlighted gets it from.

It's certainly plausible that we missed something when making this configurable, but I haven't spotted what yet.

danielballan:
Thanks for taking a look. I agree that `Instance(kernelspec.KernelSpecManager)` is still fine. I think that the next bit, [here](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L65) is the trouble:

```
class KernelManager(ConnectionFileMixin):
   ...
    def _kernel_spec_manager_default(self):
        return kernelspec.KernelSpecManager(data_dir=self.data_dir)
```

If I *remove* that definition, the KM falls over, so I think the configured KSM is not being propagated. If I *replace* that definition to use my custom class, everything works the way I want.

Am I misunderstanding how to invoke the configuration? Or is there a problem I can help fix?

danielballan:
[...]

PR #58, opened 2015-08-13
Make ipykernel and kernel_spec properties instead of traitlets

takluyver:
There was a dictionary order dependent issue if kernel_name got set before kernel_spec_manager. Turning a couple of traitlets into properties avoids that.

Closes gh-56 (hopefully) - @danielballan, can you test this?

kernel_spec is cached, so it only gets looked up once. Changing kernel_name or kernel_spec_manager invalidates the cache.

This would break passing kernel_spec directly to KernelManager, but I don't think we do that anywhere. It's hard to know with traitlets whether the ability to pass a given attribute is by design or just
incidental.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -64,20 +64,23 @@ class KernelManager(ConnectionFileMixin):
     def _kernel_spec_manager_default(self):
         return kernelspec.KernelSpecManager(data_dir=self.data_dir)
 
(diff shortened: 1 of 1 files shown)


danielballan:
Yep, this fixes #56. Thanks a lot for digging into this.

takluyver:
No problem, thanks for testing it.

Issue #62, opened 2015-08-24
Session.debug prints messages sent but not messages received


jpallas:
In developing/debugging a kernel, it's helpful to see messages in both directions.  Setting Session.debug=True will only print messages sent to the kernel, making only half the conversation visible.

Carreau:
Do you want to make a PR ?

PR #64, opened 2015-08-25
Session debug prints incoming messages

jpallas:
My stab at #62 

I'm not sure if this is the best place to put this or if it covers everything that might be useful.


Carreau:
Thanks !

PR #68, opened 2015-09-04
Interrupt the kernel and its child processes

jdemeyer:
When interrupting the kernel, you usually want to interrupt the kernel process and its children also. To implement this, we run the kernel in a new process group.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -398,11 +398,14 @@ class KernelManager(ConnectionFileMixin):
         only useful on Unix systems.
         """
         if self.has_kernel:
-            try:
-                pgid = os.getpgid(self.kernel.pid)
-                os.killpg(pgid, signum)
-            except Exception:
-                self.kernel.send_signal(signum)
+            if hasattr(os, "getpgid") and hasattr(os, "killpg"):
+                try:
+                    pgid = os.getpgid(self.kernel.pid)
+                    os.killpg(pgid, signum)
+                    return
+                except OSError:
+                    pass
+            self.kernel.send_signal(signum)
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -385,18 +385,24 @@ class KernelManager(ConnectionFileMixin):
                 from .win_interrupt import send_interrupt
                 send_interrupt(self.kernel.win32_interrupt_event)
             else:
-                self.kernel.send_signal(signal.SIGINT)
+                self.signal_kernel(signal.SIGINT)
         else:
             raise RuntimeError("Cannot interrupt kernel. No kernel is running!")
 
     def signal_kernel(self, signum):
-        """Sends a signal to the kernel.
+        """Sends a signal to the process group of the kernel (this
+        usually includes the kernel and any subprocesses spawned by
+        the kernel).
 
         Note that since only SIGTERM is supported on Windows, this function is
         only useful on Unix systems.
         """
         if self.has_kernel:
-            self.kernel.send_signal(signum)
+            try:
+                pgid = os.getpgid(self.kernel.pid)
+                os.killpg(pgid, signal.SIGINT)
+            except OSError:
+                self.kernel.send_signal(signum)
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -400,8 +400,8 @@ class KernelManager(ConnectionFileMixin):
         if self.has_kernel:
             try:
                 pgid = os.getpgid(self.kernel.pid)
-                os.killpg(pgid, signal.SIGINT)
-            except OSError:
+                os.killpg(pgid, signum)
+            except Exception:
                 self.kernel.send_signal(signum)
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -114,6 +114,10 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
         if independent:
             kwargs['preexec_fn'] = lambda: os.setsid()
         else:
+            # Create a new process group. This makes it easier to
+            # interrupt the kernel, because we want to interrupt the
+            # children of the kernel process also.
+            kwargs['preexec_fn'] = lambda: os.setpgrp()
             env['JPY_PARENT_PID'] = str(os.getpid())
 
         proc = Popen(cmd, **kwargs)


jupyter_client/manager.py
```diff
@@ -385,18 +385,24 @@ def interrupt_kernel(self):
...
+        the kernel).
 
         Note that since only SIGTERM is supported on Windows, this function is
         only useful on Unix systems.
         """
         if self.has_kernel:
-            self.kernel.send_signal(signum)
+            try:
+                pgid = os.getpgid(self.kernel.pid)
+                os.killpg(pgid, signal.SIGINT)
```
Carreau: would that be signum ?

Carreau:
Seem reasonable to me.

jupyter_client/manager.py
```diff
@@ -385,18 +385,24 @@ def interrupt_kernel(self):
...
+        usually includes the kernel and any subprocesses spawned by
+        the kernel).
 
         Note that since only SIGTERM is supported on Windows, this function is
         only useful on Unix systems.
         """
         if self.has_kernel:
-            self.kernel.send_signal(signum)
+            try:
+                pgid = os.getpgid(self.kernel.pid)
```
minrk: This will fail with AttributeError on Windows. Perhaps an explicit check for Windows?

jdemeyer:
This should fix both comments.

jupyter_client/manager.py
```diff
@@ -385,18 +385,24 @@ def interrupt_kernel(self):
...
+                os.killpg(pgid, signum)
+            except Exception:
```
damianavila: This is maybe to general... I guess Min was asking for an explicit check for win like, `if sys.platform == 'win32':`

minrk: or check directly for hasattr(getpgid/killpg)

jdemeyer: Is the "too general" really a problem? I tried the patch on Linux and it works as it should. Since I have no Windows machine to test on, I decided to play it safe and just do `except Exception:`.

> or check directly for hasattr(getpgid/killpg)

Does the fact that those functions exist mean that they will succeed? The problem is that I don't really know the failure modes of these `os.foo()` functions.

minrk: > Is the "too general" really a problem?

Yes, it can mask bugs. e.g. if someone introduces a typo that raises a NameError, this will silently do the wrong thing.

jdemeyer: > Yes, it can mask bugs. e.g. if someone introduces a typo that raises a NameError, this will silently do the wrong thing.

Right, I now remember such a bug in Sage once, it was hard to debug that one.

So, what's your concrete proposal to handle these exceptions? Is `except (OSError, AttributeError)` acceptable?

minrk: I would make an explicit `if hasattr(os, 'getpgid')`

minrk: ...and then only catch OSError

minrk:
Setting pgrp seems to cause problems with subprocesses not being able to properly launch subprocesses of their own. We may have to revert this if we can't figure out a way to fix that.

To test:

```python
from subprocess import PIPE, check_output
check_output(['bash', '-i', '-c', 'exit'], stderr=PIPE, stdin=PIPE)
```

which hangs after this PR and doesn't before.

I'm looking into a way to fix it, but this is a bit outside my area. I don't understand why creating the subprocess is failing, but I have narrowed it down to the `fork_exec` call in `_posixsubprocess` that is never finishing.

takluyver:
I'm trying to investigate this by sprinkling print statements through subprocess code, but it seems to hang at different points in a non-deterministic fashion. It is regularly getting past `fork_exec` on my system, though.

minrk:
Sorry, that wasn't quite right. It's not that `fork_exec` itself hangs, it's that it doesn't produce the message on the pipe which is flushed afterward.

takluyver:
It varies when I run it - sometimes it gets stuck reading the pipe to look for an error from the child process, but other times it gets past that.

takluyver:
And even more bizarre: sometimes I run it and `Popen()` returns - the cell can even finish executing - but the next thing that I return hangs. I wonder if bash is somehow hijacking the session and suspending the kernel...

jdemeyer:
```
check_output(['bash', '-i', '-c', 'exit'], stderr=PIPE, stdin=PIPE)
```
`bash -i` means "interactive". You cannot realisticly expect that to work in Jupyter. If you really need an interactive shell, use `pexpect` for example.

takluyver:
Normally `bash -i -c exit` exits immediately, it doesn't go into interactive mode. This is for xonsh, and I think @scopatz uses it to process bash config files, because bash processes a different set of config files in interactive mode and non-interactive mode.

The aim isn't to actually interact with bash. But even if that was the idea, `Popen()` shouldn't fail.

jdemeyer:
For the record, this is the relevant code from bash, in `jobs.c`:
```
      while ((terminal_pgrp = tcgetpgrp (shell_tty)) != -1)
        {
          if (shell_pgrp != terminal_pgrp)
            {
              SigHandler *ottin;

              ottin = set_signal_handler(SIGTTIN, SIG_DFL);
              kill (0, SIGTTIN);
              set_signal_handler (SIGTTIN, ottin);
              continue;
            }
          break;
        }
```

jdemeyer:
You can solve this by running `bash -i` in its own session:
```
import os
from subprocess import PIPE, check_output
check_output(['bash', '-i', '-c', 'exit'], stderr=PIPE, stdin=PIPE, preexec_fn=os.setsid)
```

takluyver:
Yep, coming at it from the other direction, I had just found that the kernel was getting SIGTTIN.

I'm inclined to say that @scopatz should run bash in its own process group (or even session). It's not clear to me whether we should also be doing #109 to run the kernel in its own session. I'm glad we've tracked down the cause of the issue, anyway.

jdemeyer:
Sessions (created with `setsid`) are usually associated to interactive logins from a terminal. That doesn't seem to apply to Jupyter, so I would stick with `setpgrp`.

> I'm inclined to say that @scopatz should run bash in its own process group (or even session).

Given that he wants an *interactive* `bash`, it should be run in its own session.

takluyver:
I don't think he wants an actually interactive bash, just for bash to read its 'interactive mode' config files. But it's probably best to run it in a new session either way.

willingc:
It likely is better to avoid setpgrp since there is no guarantee on its underlying implementation depending on specific os or portability. Like setting up a daemon process, I believe setsid is a more flexible option followed perhaps by setpgid or setgid and possibly clearing umask depending on the use case.

scopatz:
I think that @takluyver has it spot on.  For xonsh, even though we are running Bash in interactive mode, we don't actually want an interactive shell. For Bash, lot's of different things get rolled up into whether you want interactive mode, login mode, both, or neither.  This includes what run control files, profiles, and other configuration files get loaded.  Running in interactive mode is (in my opinion) the only reliable way to get these configurations right, even if you don't care about the actual interactivity.  We have been down both forks of this road a lot in xonsh.  Running Bash in its own process group or session seems like the right idea to me.

PR #75, opened 2015-09-25
Use ENV_JUPYTER_PATH as default system path for installing kernels

jdemeyer:
When installing kernels using `install_kernel_spec()`, the default directory should be what distutils uses to install data files, which is not `/usr/local/share` but some directory based on `sys.prefix`. So, we should use `ENV_JUPYTER_PATH` instead of `SYSTEM_JUPYTER_PATH` for the system installation directory.

Apply this instead of jupyter/jupyter_core#61

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -15,7 +15,7 @@ from ipython_genutils.py3compat import PY3
 from traitlets import HasTraits, List, Unicode, Dict, Set
 from traitlets.config import LoggingConfigurable
 
-from jupyter_core.paths import jupyter_data_dir, jupyter_path, SYSTEM_JUPYTER_PATH
+from jupyter_core.paths import jupyter_data_dir, jupyter_path, ENV_JUPYTER_PATH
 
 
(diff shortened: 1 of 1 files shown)


minrk:
`sys.prefix` is not a reliable indicator of where things should go, and it has the wrong value is many common cases as discussed in jupyter/jupyter_core#61.

minrk:
That said, we could accept this change and push a patch to core that changes how ENV_JUPYTER_PATH is determined to actually get the information from distutils instead of the unreliable sys.prefix. I never like interacting with distutils, but it might be the right source of this information.

pkgw:
If this is applied, a similar change should be made to `_get_nbext_dir` in `notebook/nbextensions.py`.

FWIW, in the Anaconda distribution, `sys.prefix` and friends have eminently reasonable values.

minrk:
@pkgw yes, in envs sys.prefix should always be reliable. But that's not everywhere, and sys.prefix is definitely wrong in very common cases, so we would need to special-case those if we end up going this way.

jdemeyer:
I think this can be closed now.

PR #77, opened 2015-10-04
Renamed `jupyter-kernelspec` for Windows support

tmr232:
The `.py` suffix is required for the script to work on Windows machines.


minrk:
On Windows, installation should be done with setuptools, in which case entrypoints are used with `.exe` wrappers, and these script files are ignored.

tmr232:
I installed using pip. The installation generated some `.exe` files (e.g. `jupyter-qtconsole.exe`), but also several scripts, namely `jupyter-nbconvert` and `jupyter-kernelspec`.

Should I list that as an issue?

minrk:
@tmr232 sorry, I didn't realize that we hadn't pushed a release for these two packages since making the setuptools changes. There will be a release of jupyter-client and nbconvert shortly that will fix this.

PR #88, opened 2015-10-28
require netifaces

### LINKED ISSUE #87: Replace calls to `ipconfig.exe` and the likes with actual API calls
Currently, IP addresses of the host are checked using OS utilities (`ipconfig.exe`, `ifconfig` and the likes).
While this currently works, it is at the risk of failing should the utilities change in their output format (I believe that Windows localization might already cause failures on some machines).

I suggest we change the code to use proper APIs.

[this](https://gist.github.com/sphaero/f9da6ebb9a7a6f679157) is a good starting point for the Windows implementation.

minrk:
Has always been preferred when available, but I think it's safe to require it now.

This removes the need to fallback on shelling out to subprocesses.

closes #87

### DIFF
--- a/jupyter_client/localinterfaces.py
+++ b/jupyter_client/localinterfaces.py
@@ -3,11 +3,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-import os
-import re
-import socket
-from subprocess import Popen, PIPE
-
 from warnings import warn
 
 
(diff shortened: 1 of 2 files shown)


Carreau:
I'm a bit concern in removing all that as netifaces seem to not have been updated in more than a year (no 3.5 wheel either), and not available through conda. 
(yes you can pip in conda) 

So this would actually mean that 4.2 would need extra packaging downstream, so what do you think of pinging continuum on that ?

minrk:
It hasn't been updated in a while, but I'm not sure that's an indicator of problems for a pretty feature-complete wrapper of super stable APIs. Keeping the builds going on different systems seems to be the only occasional issue. I pinged the dev about a 3.5 wheel, and [made conda packages](https://github.com/conda/conda-recipes/pull/451), which was a trivial `conda skeleton pypi`. We'll see how that goes, and can hold this in the meantime. I'm not too worried about it, since people can still get the same behavior with a manual install of netifaces after the fact.

Carreau:
Thanks ! Seem like a good solution that suits me.

damianavila:
> Thanks ! Seem like a good solution that suits me.

I agree...

takluyver:
Still no release, and apparently installing the latest release from source [fails on Python 3](https://bitbucket.org/al45tair/netifaces/issues/13/0104-install-is-broken-on-python-3x). I think this has to stay on the shelf for the time being.

minrk:
Yup, I'll close here. Perhaps a `ctypes` approach will be preferable.

PR #92, opened 2015-11-13
Make kernel_spec_class configurable

mariusvniekerk:
Added configurable behavior to KernelSpec for subclassing.

This opens the door for other kernel spec managers to add additional
properties (such as remote hosts for remote kernel, docker image names
for a docker based kernel etc)

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -75,7 +75,7 @@ class NoSuchKernel(KeyError):
 class KernelSpecManager(LoggingConfigurable):
 
     kernel_spec_class = DottedObjectName(
-        "jupyter_client.kernel_spec.KernelSpec", config=True,
+        "jupyter_client.kernelspec.KernelSpec", config=True,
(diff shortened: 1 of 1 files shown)


jupyter_client/kernelspec.py
```diff
@@ -73,6 +73,21 @@ def __init__(self, name):
         self.name = name
 
 class KernelSpecManager(LoggingConfigurable):
+
+    kernel_spec_class = DottedObjectName(
```
minrk: This can be `Type`, and then the factory becomes unnecessary.

mariusvniekerk: Okay, let me change that.  Basically just looked at MultiKernelManager for how to set that up.  Is there a better example in the codebase?

minrk: Yeah, we realized that Type was simpler after we had started using DON + factory in various places. We haven't updated everywhere yet, but we are at least using Type in new places. Looks like you got it worked out.

PR #93, opened 2015-11-15
Add KernelSpecManager.get_all_specs

### LINKED ISSUE #66: simplify get/find_kernel_specs.
listing all kernelspec (`$ jupyter kernelspec list`) make N+1 call to find_kernel_specs (one for each found kernel) 

This could be highly simplified with a "find_all_specs".

Not important, but simple for beginner, so marking as sprint friendly.

captainsafia:
Closes #66.

cc/ @minrk @Carreau

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -167,7 +167,9 @@ class KernelSpecManager(LoggingConfigurable):
 
         return self._get_kernel_spec_by_name(kernel_name, resource_dir)
 
-    def find_all_specs(self):
+    def get_all_specs(self):
+        """Returns a dict mapping kernel names and resource directories.
+        """
         d = self.find_kernel_specs()
         return {kname: {
                 "resource_dir": d[kname],
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -138,7 +138,7 @@ class KernelSpecManager(LoggingConfigurable):
         return d
         # TODO: Caching?
 
-    def get_kernel_spec_helper(self, kernel_name, resource_dir):
+    def _get_kernel_spec_by_name(self, kernel_name, resource_dir):
         """ Returns a :class:`KernelSpec` instance for a given kernel_name
         and resource_dir.
         """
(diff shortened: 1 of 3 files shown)


jupyter_client/kernelspec.py
```diff
@@ -138,17 +138,10 @@ def find_kernel_specs(self):
         return d
         # TODO: Caching?
 
-    def get_kernel_spec(self, kernel_name):
-        """Returns a :class:`KernelSpec` instance for the given kernel_name.
-
-        Raises :exc:`NoSuchKernel` if the given kernel name is not found.
+    def get_kernel_spec_helper(self, kernel_name, resource_dir):
```
JanSchulz: This is IMO misnamed, at least I have trouble seeing this NOT as a function, which gets a helper object, but it returns a kernel spec as far as I see it. So `_get_kernel_spec_for_name(name, dir)`?

jupyter_client/kernelspec.py
```diff
@@ -138,17 +138,10 @@ def find_kernel_specs(self):
...
+        """ Returns a :class:`KernelSpec` instance for a given kernel_name
+        and resource_dir.
         """
-        d = self.find_kernel_specs()
-        try:
-            resource_dir = d[kernel_name.lower()]
```
JanSchulz: Doesn't this make it `get_kernel_spec_helper(self, kernel_name)` as it makes the passed in `ressource_dir` argument obsolete?

JanSchulz:
Does this also need a change in the notebook to take advantage of this new thing?

jupyter_client/kernelspec.py
captainsafia: I concur with what you say, @JanSchulz. Let me update the code.

captainsafia:
Noted, not sure if it'll be include in the 4.1 release but I don't see any reason as to why it shouldn't.

jupyter_client/kernelspec.py
```diff
@@ -161,6 +154,26 @@ def get_kernel_spec(self, kernel_name):
...
+        """
+        d = self.find_kernel_specs()
+        try:
+            resource_dir = d[kernel_name.lower()]
+        except KeyError:
+            raise NoSuchKernel(kernel_name)
+
+        return self._get_kernel_spec_by_name(kernel_name, resource_dir)
+
+    def find_all_specs(self):
```
minrk: let's call this `get_all_specs`. That way `get_` returns kernelspec objects, while `find_` returns locations on disk.

And don't forget to add a docstring.

minrk:
Thanks!

minrk:
@captainsafia feel free to ping with a comment when you've addressed notes, so we can review/merge; we don't get notified of new commits, only comments.

Thanks again, this is great!

captainsafia:
Yep! For some strange reason, I thought i had more work to do on this before it was merge ready but if it looks good, sweet!

captainsafia:
I remember it now. Had to rename the `test_find_all_specs` to `test_get_all_specs` but I don't think that's too big a deal.

PR #96, opened 2015-11-29
Add kernel spec name to connection file

### LINKED ISSUE #50: Include kernelspec name in connection file
For frontends like [sidecar](https://github.com/rgbkrk/jupyter-sidecar), it would be nice if connection files included the kernelspec name. This is solely to find out the logo and language name without having to make a `kernel_info_request`.

/cc @minrk @jdfreder from discussions at Medici Cafe in Austin

captainsafia:
Closes #50.

This doesn't feel like the right approach so opening up this PR for advisement.

cc/ @minrk

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -33,7 +33,7 @@ from jupyter_core.paths import jupyter_data_dir, jupyter_runtime_dir
 
 def write_connection_file(fname=None, shell_port=0, iopub_port=0, stdin_port=0, hb_port=0,
                          control_port=0, ip='', key=b'', transport='tcp',
-                         signature_scheme='hmac-sha256',
+                         signature_scheme='hmac-sha256', kernel_name=''
                          ):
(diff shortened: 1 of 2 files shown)


minrk:
What doesn't feel right about this?

captainsafia:
I'm not entirely sure that the `kernel_name` is propagated properly when the connection file is loaded. For example, if in the test case I sent the `kernel_name` key in the `sample_info` to a value other than "Python" then the test for `load_connection_file` still returns "python" instead of "test" which makes me think that perhaps that `load_connection_file` overwrites the `kernel_name` variable somewhere.

Let me know if some of that doesn't make sense. I'll also push the test set I've been using.

rgbkrk:
Looks like you've updated it. How are you feeling about it @minrk?

minrk:
AOK :+1:

rgbkrk:
Thanks @captainsafia!

PR #97, opened 2015-11-30
clarify that msg_id, session are strings

minrk:
UUID is a convenient way to generate unique strings,
but these are not fundamentally UUID objects.

closes jupyter/notebook#768


Carreau:
+1

PR #103, opened 2015-12-06
Write the new connection file after establishing SSH tunnels

### LINKED ISSUE #102: modified connection file not created with --existing --ssh
When I use `ipython console --existing kernel-9072.json --ssh server`, it works and informs me that
```
[ZMQTerminalIPythonApp] To connect another client via this tunnel, use:
[ZMQTerminalIPythonApp] --existing kernel-9072-ssh.json
```

However, that modified connection file is apparently never created. I think the reason is that [with --existing a kernel manager is not created](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/consoleapp.py#L255), so the [code that is responsible for writing the new connection file](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/consoleapp.py#L287) is never run.
Also, the `basename` usage in the [code responsible for creating the name of the new connection file](https://github.com/jupyter/jupyter_client/blame/master/jupyter_client/consoleapp.py#L235) seems strange, perhaps one of them was supposed to be `dirname`?

```
$ python -c "import IPython; print(IPython.sys_info())" 
{'commit_hash': u'f534027',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/home/ondrej/anaconda/lib/python2.7/site-packages/IPython',
 'ipython_version': '4.0.0',
 'os_name': 'posix',
 'platform': 'Linux-3.16.0-4-amd64-x86_64-with-debian-8.2',
 'sys_executable': '/home/ondrej/anaconda/bin/python',
 'sys_platform': 'linux2',
 'sys_version': '2.7.10 |Anaconda 2.3.0 (64-bit)| (default, Oct 19 2015, 18:04:42) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]'}
```

smartass101:
also make sure the connection file is in the same dir as the original file
Fixes #102

I have tested this like this:
```
PYTHONPATH=~/projects/jupyter_client: ipython console --ssh server --existing kernel-9072.json
```
Afterwards another connection can be made with
```
ipython console --existing kernel-9072-ssh.json
```

### DIFF
--- a/jupyter_client/consoleapp.py
+++ b/jupyter_client/consoleapp.py
@@ -231,11 +231,11 @@ class JupyterConsoleApp(ConnectionFileMixin):
         self.shell_port, self.iopub_port, self.stdin_port, self.hb_port = newports
         
         cf = self.connection_file
-        base,ext = os.path.splitext(cf)
-        base = os.path.basename(base)
-        self.connection_file = os.path.basename(base)+'-ssh'+ext
+        root, ext = os.path.splitext(cf)
+        self.connection_file = root + '-ssh' + ext
+        self.write_connection_file() # write the new connection file
         self.log.info("To connect another client via this tunnel, use:")
-        self.log.info("--existing %s" % self.connection_file)
+        self.log.info("--existing %s" % os.path.basename(self.connection_file))
     
     def _new_connection_file(self):
         cf = ''


minrk:
Thanks!

PR #106, opened 2016-01-13
warn if message signing is disabled

minrk:
closes jupyter/notebook#945


willingc:
@minrk This pr looks good to me.

I did notice that later in the file (line 660) there is a ``get_logger().warn`` statement. Since warn is deprecated in CPython and in not written into traitlets, I believe this should be updated to ``get_logger().warning``.

PR #107, opened 2016-01-13
make fork-safety check optional

minrk:
allows other code to take over fork-safety, as is planned in ipython/ipykernel#85

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -268,7 +268,13 @@ class Session(Configurable):
 
     """
 
-    debug=Bool(False, config=True, help="""Debug output in the Session""")
+    debug = Bool(False, config=True, help="""Debug output in the Session""")
+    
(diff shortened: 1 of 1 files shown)


Carreau:
unfortunately need rebase.

minrk:
rebased

minrk:
Perhaps what makes more sense is a per-socket fork-safe signaling mechanism (e.g. `getattr(socket, '_fork_safe', False)`). Or we could assume objects that aren't zmq.Socket are fork-safe. I'm not sure what the best solution is.

Carreau:
> Perhaps what makes more sense is a per-socket fork-safe signaling mechanism (e.g. getattr(socket, '_fork_safe', False)). Or we could assume objects that aren't zmq.Socket are fork-safe. I'm not sure what the best solution is.

Let's go with this one for now.

PR #108, opened 2016-01-13
Making the NoSuchKernel error produce a more useful error message

ngoldbaum:
Before if this error is raied, it just prints `NoSuchKernel`, and you need to use a debugger to see which kernel it's complaining about. Now the error message printed along with the exception will include the name of the kernel:

```
Exception occurred:
    raise NoSuchKernel(kernel_name)
NoSuchKernel: No such kernel named python3
```


takluyver:
:+1:

PR #109, opened 2016-01-14
start new session instead of setpgrp for detaching kernel

minrk:
setpgrp seems to cause problems with children spawning subprocesses. setsid doesn't cause the same problems, but has its own issues.

This is WIP because I need to at least add some tests to exercise various subprocesses-of-subprocesses cases.

cc @takluyver who knows more about terminals and process groups than I do.

### DIFF
--- a/jupyter_client/tests/signalkernel.py
+++ b/jupyter_client/tests/signalkernel.py
@@ -27,20 +27,23 @@ class SignalTestKernel(Kernel):
     def do_execute(self, code, silent, store_history=True, user_expressions=None,
                    allow_stdin=False):
         code = code.strip()
-        reply = {'status': 'ok'}
+        reply = {
+            'status': 'ok',
+            'user_expressions': {},
+        }
         if code == 'start':
             child = Popen(['bash', '-i', '-c', 'sleep 30'], stderr=PIPE)
             self.children.append(child)
-            reply['pid'] = self.children[-1].pid
+            reply['user_expressions']['pid'] = self.children[-1].pid
         elif code == 'check':
-            reply['poll'] = [ child.poll() for child in self.children ]
+            reply['user_expressions']['poll'] = [ child.poll() for child in self.children ]
         elif code == 'sleep':
             try:
                 time.sleep(10)
             except KeyboardInterrupt:
-                reply['interrupted'] = True
+                reply['user_expressions']['interrupted'] = True
             else:
-                reply['interrupted'] = False
+                reply['user_expressions']['interrupted'] = False
         else:
             reply['status'] = 'error'
             reply['ename'] = 'Error'
--- a/jupyter_client/tests/signalkernel.py
+++ b/jupyter_client/tests/signalkernel.py
@@ -1,0 +1,57 @@
+"""Test kernel for signalling subprocesses"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from __future__ import print_function
+
+from subprocess import Popen, PIPE
+import sys
+import time
+
+from ipykernel.kernelbase import Kernel
+from ipykernel.kernelapp import IPKernelApp
+
+
+class SignalTestKernel(Kernel):
+    """Kernel for testing subprocess signaling"""
+    implementation = 'signaltest'
+    implementation_version = '0.0'
+    banner = ''
+    
+    def __init__(self, **kwargs):
+        super(SignalTestKernel, self).__init__(**kwargs)
+        self.children = []
+        
+
+    def do_execute(self, code, silent, store_history=True, user_expressions=None,
+                   allow_stdin=False):
+        code = code.strip()
+        reply = {'status': 'ok'}
+        if code == 'start':
+            child = Popen(['bash', '-i', '-c', 'sleep 30'], stderr=PIPE)
+            self.children.append(child)
+            reply['pid'] = self.children[-1].pid
+        elif code == 'check':
+            reply['poll'] = [ child.poll() for child in self.children ]
+        elif code == 'sleep':
+            try:
+                time.sleep(10)
+            except KeyboardInterrupt:
+                reply['interrupted'] = True
+            else:
+                reply['interrupted'] = False
+        else:
+            reply['status'] = 'error'
+            reply['ename'] = 'Error'
+            reply['evalue'] = code
+            reply['traceback'] = ['no such command: %s' % code]
+        return reply
+
+class SignalTestApp(IPKernelApp):
+    kernel_class = SignalTestKernel
+    def init_io(self):
+        pass # disable stdout/stderr capture
+    
+if __name__ == '__main__':
+    SignalTestApp.launch_instance()
--- a/jupyter_client/tests/test_kernelmanager.py
+++ b/jupyter_client/tests/test_kernelmanager.py
@@ -1,15 +1,46 @@
-"""Tests for the notebook kernel and session manager"""
+"""Tests for the KernelManager"""
 
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+
+import json
+import os
+pjoin = os.path.join
+import signal
 from subprocess import PIPE
+import sys
 import time
 from unittest import TestCase
 
 from ipython_genutils.testing import decorators as dec
 
 from traitlets.config.loader import Config
+from jupyter_core import paths
 from jupyter_client import KernelManager
+from ..manager import start_new_kernel
+from .utils import test_env
+
+TIMEOUT = 30
 
 class TestKernelManager(TestCase):
+    def setUp(self):
+        self.env_patch = test_env()
+        self.env_patch.start()
+    
+    def tearDown(self):
+        self.env_patch.stop()
+    
+    def _install_test_kernel(self):
+        kernel_dir = pjoin(paths.jupyter_data_dir(), 'kernels', 'signaltest')
+        os.makedirs(kernel_dir)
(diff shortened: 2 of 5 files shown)


takluyver:
The problem doesn't happen with any subprocess - I think it's something that bash does, and more specifically something related to the `-i` flag: if I take that out, then `Popen()` returns.

My guess is that `bash -i` does something with the controlling TTY. setsid() detaches the process from the controlling tty, which probably prevents bash doing something.

I'm still puzzled as to how bash can cause `Popen()` to hang, though. I'd like to understand this properly before we merge this, even if it is the right fix.

minrk:
I'm *super* puzzled, but I've spent the entire day investigating this and don't have a lot more information than when I started. I'm running out of steam to pursue this one.

It's not strictly `-i`, since I've seen the problem with bash without `-i`, though trivial commands like `bash -c exit` or `sleep 5` work fine.

I really don't understand how the pgrp could affect whatever it is that's failing if **all** of stdout, stderr, and stdin are setup as pipes.

takluyver:
The 'controlling tty' of a process is a separate notion from std(in/out/err) - even if those are all connected to pipes, a process can still get its controlling tty by opening `/dev/tty`. Part of what `setsid()` does is to detach the process from its controlling tty, so opening `/dev/tty` no longer works.

minrk:
So it seems like starting a new session is what we want, then, not setpgrp. I'll update this to use `start_new_session` on py3 and fallback on setsid for Legacy Python.

willingc:
@minrk Nice tests :+1: The logic for all looks good.

minrk:
this now uses `start_new_session` when available, and `setsid` on old and busted Python. I added a test that verifies that the kernel can start `bash -i` subprocesses and the whole process group gets interrupted (the same test passes without the session patch as long as the subprocess is simpler, such as `sleep 30`).

jdemeyer:
The comment is wrong:
```
We don't use setpgrp, which can cause problems for kernels starting subprocesses
```

The problem you're seeing is very specific to `bash -i`, it has nothing to do with running subprocesses in general.

minrk:
@jdemeyer It has also been seen by users trying to start supposedly headless java processes, and `bash -c` *without* `-i`, so it's not confined to `bash -i`.

jupyter_client/tests/signalkernel.py
```diff
@@ -0,0 +1,57 @@
...
+        
+
+    def do_execute(self, code, silent, store_history=True, user_expressions=None,
+                   allow_stdin=False):
+        code = code.strip()
+        reply = {'status': 'ok'}
+        if code == 'start':
+            child = Popen(['bash', '-i', '-c', 'sleep 30'], stderr=PIPE)
+            self.children.append(child)
+            reply['pid'] = self.children[-1].pid
```
takluyver: We're sticking arbitrary extra fields in the reply? I'm a bit surprised none of our machinery complains about that. Do we want to assume that will always be the case? We could stick them in metadata or `user_expressions`.

jupyter_client/tests/signalkernel.py
```diff
@@ -0,0 +1,57 @@
...
+                time.sleep(10)
+            except KeyboardInterrupt:
+                reply['interrupted'] = True
+            else:
+                reply['interrupted'] = False
+        else:
+            reply['status'] = 'error'
+            reply['ename'] = 'Error'
+            reply['evalue'] = code
+            reply['tmraceback'] = ['no such command: %s' % code]
```
takluyver: typo in traceback

jupyter_client/tests/test_kernelmanager.py
```diff
@@ -51,3 +82,41 @@ def test_get_connect_info(self):
...
         ])
         self.assertEqual(keys, expected)
+    
+    @dec.skip_win32
+    def test_signal_kernel_subprocesses(self):
+        self._install_test_kernel()
+        km, kc = start_new_kernel(kernel_name='signaltest')
+        def execute(cmd):
+            kc.execute(cmd)
+            reply = kc.get_shell_msg(TIMEOUT)
```
takluyver: 30 seconds is going to feel like waiting forever if we ever break what you just fixed and the test kernel is suspended. But I guess any shorter and it risks sporadic failures on Travis.

takluyver:
:+1: overall, just a couple of comments.

jdemeyer:
> @jdemeyer It has also been seen by users trying to start supposedly headless java processes, and bash -c without -i, so it's not confined to bash -i.

Well, the issue that you reported at #68 is certainly confined to `bash -i`: I showed the actual code in the `bash` sources which is responsible for the breakage.

I guess that those other problems you mention are unrelated and perhaps those are not fixed by this PR. Do you have concrete examples?

So I would still suggest to mention in the comments that this PR is meant to fix just `bash -i` and nothing else.

willingc:
@jdemeyer While I understand that you are trying to be precise, the tone of your message is not constructive. "setpgrp" is problematic because its behavior is unpredictable based on OS implementation. The "setpgrp" command may be removed in future standards [reference](http://pubs.opengroup.org/onlinepubs/9699919799/).

jdemeyer:
> the tone of your message is not constructive

Why do you think so? I think you should justify why you said that.

I downloaded the `bash` source code and pointed out which part of `bash` is responsible for the issue mentioned at #68. It shows that the issue is really caused by very specific behaviour of `bash -i`. And then pointing this out is not constructive?

jdemeyer:
> "setpgrp" is problematic because its behavior is unpredictable based on OS implementation.

1) Why do you think that?

2) Why does `setsid()` not have this problem?

> The "setpgrp" command may be removed in future standards reference.

Because it will be replaced by [setpgid()](http://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html), not because process groups are fundamentally problematic.

jdemeyer:
To be clear: I have nothing against replacing `setpgrp()` by `setsid()` (although I'm not particularly in favour of it either).

I am only saying that the very vague comment
```
We don't use setpgrp, which can cause problems for kernels starting subprocesses
```
should be replaced by something like
```
We don't use setpgrp, which can cause problems for running 'bash -i'
```

takluyver:
We observed something that might be similar in jupyter/jupyterhub#382, where it looked like starting another process was causing problems. We didn't investigate that to find out the real cause, though.

@willingc the docs you linked to say:

> applications are encouraged to use setpgid() or setsid() as appropriate.

Any thoughts on which is more likely to be appropriate for us? My gut feeling is that we may as well do setsid - as this PR does - but I don't know this stuff very well.

Shall we make the comment say something like '...starting certain applications in a subprocess'? It's definitely not a problem with starting any subprocess, and it's likely that bash isn't the only application ever written that fires signals around. :-)

jdemeyer:
> applications are encouraged to use setpgid() or setsid() as appropriate.

Right, I missed this part. It's a bit confusing because there is nothing really wrong with process group, only with the `setpgrp()` system call (which should be replaced by `setpgid`).

jdemeyer:
> My gut feeling is that we may as well do setsid - as this PR does - but I don't know this stuff very well.

> Any thoughts on which is more likely to be appropriate for us? My gut feeling is that we may as well do setsid - as this PR does - but I don't know this stuff very well.

Hard to tell, I don't have a strong opinion either way. I cannot really find a major semantic difference between process groups and sessions. Sessions are typically associated to a single interactive login and one might as well consider a Jupyter notebook an "interactive login".

One thing to test is whether this affects shutting down the Jupyter server.

takluyver:
The functional differences I'm aware of:

- `setsid()` detaches the process (and any children) from the controlling terminal. They can still write to the terminal as long as they have an fd for it. This is probably beneficial for us, because it provides a layer of protection against applications trying to grab control of the terminal.
- A process that has called `setsid()` doesn't get sent SIGHUP when its parent process exits (IIUC - this is part of starting a daemon process). When the server shuts down normally, we go through an explicit shutdown process for all the kernels it has started, so this shouldn't make a difference. However, if the notebook server were to crash, presumably its kernels would hang around. That could be good or bad, depending on your point of view.

Is there anything else?

jdemeyer:
> and it's likely that bash isn't the only application ever written that fires signals around. :-)

Sure, but sending signals depending on whether or not the application was started as a separate session is already much more specific.

> Shall we make the comment say something like '...starting certain applications in a subprocess'?

How about "...certain interactive applications (in particular bash -i) in a subprocess"?

takluyver:
That sounds fine to me. :-)

willingc:
@takluyver @jdemeyer I've been traveling all day. 
A few resources:
[Subprocess article (last section on process groups and sessions) by Doug Hellman](https://pymotw.com/2/subprocess/)

I think ``setsid`` fits here. I'm unsure if this should follow the fork/setsid/umask/fork pattern used in a daemon or if setsid would be sufficient. The "Advanced Programming in Unix 3rd ed" by W. Richard Stevens has  excellent chapters on process groups and sessions.

I think we are in agreement that setpgrp is obsolete.
screen shot 2016-01-17 at 6 49 02 pm
screen shot 2016-01-17 at 6 54 27 pm

jupyter_client/tests/signalkernel.py
minrk: That's me being lazy. With the kernelbase API there's no way to add metadata, so I used user_expressions.

minrk:
Added a note to the effect of `bash -i` and other subprocesses being the reason we don't use setpgrp.

willingc:
@minrk +1

takluyver:
Thanks everyone, I think this is ready to go.

PR #112, opened 2016-01-21
don't store signature if just peeking at headers

minrk:
avoids duplicate signature errors if one bit of code peeks at headers, but another bit actually handles the message later.


willingc:
LGTM. :ship:

PR #113, opened 2016-01-26
Update shutdown_request docs

blink1073:
Clarify the semantics of the `shutdown_request` and `shutdown_reply` messages.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -362,7 +362,7 @@ Message type: ``execute_reply``::
       # One of: 'ok' OR 'error' OR 'abort'
       'status' : str,
 
-      # The global kernel counter that increases by one with each request that 
+      # The global kernel counter that increases by one with each request that
       # stores history.  This will typically be used by clients to display
       # prompt numbers to the user.  If the request did not store history, this will
       # be the current value of the counter in the kernel.
@@ -859,7 +859,8 @@ multiple cases:
 
 The client sends a shutdown request to the kernel, and once it receives the
 reply message (which is otherwise empty), it can assume that the kernel has
-completed shutdown safely.
+completed shutdown safely.  The request can be sent on either the `control` or
+`shell` channels.
 
 Upon their own shutdown, client applications will typically execute a last
 minute sanity check and forcefully terminate any kernel that is still alive, to
@@ -868,13 +869,13 @@ avoid leaving stray processes in the user's machine.
 Message type: ``shutdown_request``::
 
     content = {
-        'restart' : bool # whether the shutdown is final, or precedes a restart
+        'restart' : bool # False if final shutdown, or True if shutdown precedes a restart
     }
 
 Message type: ``shutdown_reply``::
 
     content = {
-        'restart' : bool # whether the shutdown is final, or precedes a restart
+        'restart' : bool # False if final shutdown, or True if shutdown precedes a restart
     }
 
 .. Note::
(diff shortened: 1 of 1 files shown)


docs/messaging.rst
```diff
@@ -859,7 +859,9 @@ multiple cases:
 
 The client sends a shutdown request to the kernel, and once it receives the
 reply message (which is otherwise empty), it can assume that the kernel has
-completed shutdown safely.
+completed shutdown safely.  The request can be sent on either the `control` or
+`shell` channels.  The reply is made both to the requesting channel and to
+the `iopub` channel.
```
takluyver: Is it? @minrk is it intentional that a `*_reply` message is broadcast over iopub? I can understand that it makes sense for all clients to get a notification that shutdown is underway, but I'm a bit surprised we do it by broadcasting the `shutdown_reply`.

blink1073: https://github.com/ipython/ipykernel/blob/60f58112a0068d28c452669d2fd20a5244aab687/ipykernel/kernelbase.py#L503

takluyver: But I'm pretty sure the R kernel, for instance, doesn't broadcast shutdown_reply on IOPub. And so far, it hasn't been a problem. So I want to work out whether we document what IPython currently does as the expectation, or leave it undocumented and possibly introduce a separate official shutdown broadcast message.

willingc:
While working on the shutdown documentation, it would be helpful to users to clarify the comments in these lines:
```
Message type: shutdown_request:

content = {
    'restart' : bool # whether the shutdown is final, or precedes a restart
}
Message type: shutdown_reply:

content = {
    'restart' : bool # whether the shutdown is final, or precedes a restart
}
```

Perhaps changing  comments from ``whether the shutdown is final, or precedes a restart`` to ``False if the final shutdown, or True if shutdown precedes a restart``.

blink1073:
I think the crux of the problem is that "control" responses never get to the notebook. Maybe the right solution is to add that capability.

blink1073:
@willingc, thanks, I'll make that change.

docs/messaging.rst
minrk: It is what the IPython kernel does, and I think it was added during some of the multi-frontend stuff with the QtConsole (shutdown logic is a tangled mess over there). Sending the reply instead of a notification or status message was definitely a mistake.

I'll look at the QtConsole code and see if I can figure out what it's used for. If it behaves okay with kernels that don't implement this undocumented behavior, I would prefer updating the IPython kernel to match the spec rather than the spec to match this icky behavior from the kernel.

minrk:
> I think the crux of the problem is that "control" responses never get to the notebook

No, that's not a problem at all. Responses go to the client that requested them. In this case, the web frontend did not make a shutdown request to the kernel, it made a REST request to the notebook server. That part is behaving exactly as it should be.

minrk:
The QtConsole uses a shutdown_reply on iopub to show a dialog connected to existing kernels. This dialog currently makes very little sense:

screen shot 2016-01-27 at 10 10 31

So I'm not 100% sure what to do. The fact that the dialog makes no sense for the circumstance tells me that it's probably not coming up much, if at all. While it would technically be a regression for the IPython kernel, I'm somewhat inclined to revert that behavior and propose an official mechanism to notify that a shutdown has been requested, be it a `status:shutting_down` message or a separate `shutdown_notification` message.

takluyver:
That makes sense to me. In that case, let's leave what IPython currently does as an undocumented thing.

blink1073:
@jasongrout, @minrk, then we have the question of what to do about receiving undocumented iopub messages in `jupyter-js-services`.  We had been adding validation for the known iopub message types, which resulted in an error that triggered this discussion.

minrk:
They should probably just be logged. A bug like this in one kernel shouldn't cause problems in the client.

PR #116, opened 2016-01-28
Strip trailing path separators off kernelspec source directory

takluyver:
I just discovered that if you do `jupyter kernelspec install blah/`, with the trailing slash, it replaces your entire .../kernels/ directory rather than creating a subdirectory. This is because when it finds the kernel name, it gets the empty string rather than the directory name.

This fixes that.


Carreau:
+1

damianavila:
+1

PR #118, opened 2016-02-02
Timeout in start_new_kernel if kernel fails to respond

takluyver:
Closes jupyter/nbconvert#217

`start_new_kernel()` has a `startup_timeout` parameter, but it wasn't being used anywhere. I pass it through to `KernelClient.wait_for_ready()`, and raise RuntimeError if it times out. (Maybe should raise something else? TimeoutError? But this is for OS level operations with errno ETIMEDOUT - not sure if we should use it for arbitrary timeouts).

It also checks whether the kernel is still alive, so if the kernel has already died, we don't need to wait 60 seconds before giving up. It will still wait one second trying to get the message from the kernel.

Ping @minrk

### DIFF
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -21,7 +21,7 @@ class BlockingKernelClient(KernelClient):
         if timeout is None:
             abs_timeout = float('inf')
         else:
-            abs_timeout = time.time()
+            abs_timeout = time.time() + timeout
         # Wait for kernel info reply on shell channel
         while True:
             try:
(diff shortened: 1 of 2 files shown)


jupyter_client/blocking/client.py
```diff
@@ -16,13 +17,27 @@
 from .channels import ZMQSocketChannel
 
 class BlockingKernelClient(KernelClient):
-    def wait_for_ready(self):
+    def wait_for_ready(self, timeout=None):
+        if timeout is None:
+            abs_timeout = float('inf')
+        else:
+            abs_timeout = time.time()
```
minrk: Presumably missing `+ timeout`

minrk:
Looks like one small omission, but :+1: .

jupyter_client/blocking/client.py
takluyver: Oops, yes. Done!

PR #119, opened 2016-02-03
language_info name doc fixes

sn6uv:
This PR solves two issues with the docs related to the `language_info` field of `kernel_info_message`:

The messaging docs suggest that `name` should be the language that implements the kernel (e.g. a python kernel written in C should use `'C'`). Is this correct?

Wrapper kernels also need to provide this field, otherwise I see an error message when trying to save files. In principle this could be provided from the `Kernel.language` field but it's not being done currently.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -778,7 +778,7 @@ Message type: ``kernel_info_reply``::
 
         # Information about the language of code for the kernel
         'language_info': {
-            # Name of the programming language in which kernel is implemented.
+            # Name of the programming language that the kernel implements.
             # Kernel included in IPython returns 'python'.
             'name': str,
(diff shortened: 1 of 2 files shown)


Carreau:
> The messaging docs suggest that name should be the language that implements the kernel (e.g. a python kernel written in C should use 'C'). Is this correct?

Hum no, a Python kernel written in C should say `Python`. 

> Wrapper kernels also need to provide this field, otherwise I see an error message when trying to save files. In principle this could be provided from the Kernel.language field but it's not being done currently.

This is likely a bug. 

I'm going to merge this. Would you like to  send PR or open an issue for each of the above issues you mentioned ?

Thanks !

PR #121, opened 2016-02-04
remove source from display_data docs

### LINKED ISSUE #120: 'source' field of 'display_data' is unclear
I'm writing a kernel and the docs:
```
Message type: display_data:
content = {
    # Who create the data
    'source' : str,
    ...
}```

are not at all clear what source should be.

Experimenting with the jupyter notebook, it seems as if it doesn't even matter what value is used?

sn6uv:
fixes #120

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -928,7 +928,8 @@ Message type: ``display_data``::
     content = {
 
         # Who create the data
-        'source' : str,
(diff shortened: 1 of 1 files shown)


takluyver:
@minrk am I right that this field was removed?

minrk:
Yes, it was.

takluyver:
Thanks @sn6uv

PR #128, opened 2016-02-17
Avoid passing down unexpected kwargs

takluyver:
Squashes some warnings from recent traitlets changes.

There's still another issue with KernelClient - it's passed parameters from the connection info which it doesn't use (key and signature_scheme), because it's also passed a reference to a Session object which already has those things. I'm not sure of the neatest way to resolve that.

### DIFF
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -104,7 +104,6 @@ class MultiKernelManager(LoggingConfigurable):
         km = self.kernel_manager_factory(connection_file=os.path.join(
                     self.connection_dir, "kernel-%s.json" % kernel_id),
                     parent=self, log=self.log, kernel_name=kernel_name,
-                    **kwargs
         )
         km.start_kernel(**kwargs)
         self._kernels[kernel_id] = km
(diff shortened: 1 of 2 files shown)


jupyter_client/multikernelmanager.py
```diff
@@ -104,7 +104,6 @@ def start_kernel(self, kernel_name=None, **kwargs):
         km = self.kernel_manager_factory(connection_file=os.path.join(
                     self.connection_dir, "kernel-%s.json" % kernel_id),
                     parent=self, log=self.log, kernel_name=kernel_name,
-                    **kwargs
```
takluyver: I haven't hunted for anything that might be relying on arbitrary kwargs passed through here, but the tests pass. Does anyone know of anything?

We're passing them again to `start_kernel` just below, so if we do rely on supplying a mixture of arguments, we should probably more clearly define which ones are for which call.

minrk: This line should be left alone, since `start_kernel` is meant as a passthrough for the single kernel-manager constructor.

If nothing else, one line above kernel_spec_manager is added to kwargs.

takluyver: But it's also passing through to `start_kernel()` just below. Is there any way to define the arguments that either call should take, so we can filter them out?

minrk: Oh, now that I read it I see the problem. `kernel_spec_manager` should be passed to the constructor and kwargs should be passed on to start. You are right to remove kwargs from the constructor.

takluyver: Done

minrk:
I took a stab at the connection_info one in #130.

PR #130, opened 2016-02-17
add session flag to KernelManager.get_connection_info

minrk:
- `session=True`: include session object itself
- `session=False`: include session info only (key, signature scheme)

We were passing ignored arguments to the constructor, revealed by a recent warning added to traitlets.

related to #128


takluyver:
:+1: This approach looks OK to me. I'll merge in a day or so if no-one objects.

PR #133, opened 2016-02-23
Add ability to specify the env dict for launching the kernel

### LINKED ISSUE #117: Allow start_new_kernel to set env
Keyword arguments to `start_new_kernel` are generally forwarded to `Popen`, except env is set manually to `os.environ.copy()` (https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L232). This denies `start_new_kernel` the ability to restrict environment variables being passed into `Popen`.

By default, setting `env=None` in `Popen` will copy all of the environment variables over anyway.

pankajp:
This adds ability to use different dict other than os.environ
as the environment for launching the kernel

Fixes #117


minrk:
Thanks!

PR #134, opened 2016-02-24
Prepare docs for 4.2

minrk:
- changelog highlights
- document get_all_specs

We should be ready to ship 4.2 after this, unless there's something anyone else would like to make sure to get in.


Carreau:
Thanks !

willingc:
Thanks @minrk. Looks great.

PR #139, opened 2016-03-06
[WIP] Travis errors when a kernel client blocks and prematurely times out

willingc:
It looks like from the Travis logs of nbgrader and kernel gateway (when tests are run at verbosity=3) that the client was timing out prematurely. The logs seemed to indicate wait_for_ready().

I believe that there's a typo in the inequality.

Probably should add a test to cover the line with inequality.

Debug prints still need to be removed.
cc/@minrk, @takluyver, @jhamrick, @parente

PR #140, opened 2016-03-06
Fixes for starting a new kernel

minrk:
- timeout was backwards in Client.wait_for_ready
- don't start heartbeat in paused state, which causes is_beating() to return False

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -72,8 +72,10 @@ class HBChannel(Thread):
         self.address = address
         atexit.register(self._notice_exit)
 
+        # running is False until `.start()` is called
         self._running = False
-        self._pause = True
+        # don't start paused
+        self._pause = False
         self.poller = zmq.Poller()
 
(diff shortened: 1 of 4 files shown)


jupyter_client/channels.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, context=None, session=None, address=None):
         atexit.register(self._notice_exit)
 
         self._running = False
-        self._pause = True
+        self._pause = False
```
willingc: @minrk While I believe the fix is correct here, it's a little counterintuitive to have ``_running`` and ``_pause`` to both be False. Perhaps comment?

minrk: Clarifying comments added

minrk:
@willingc I rebased this on your #139, to incorporate the docstring/comment improvements.

willingc:
Happy to merge this. Thanks @minrk :sunny:

minrk:
Thanks, @willingc!

PR #141, opened 2016-03-09
Fix wait_for_ready() dying early for slow-starting kernels

minrk:
KernelClient.is_alive() checks heartbeat responsiveness, so it can be False early on if the kernel process is slow to start. This can cause `wait_for_ready` to falsely claim the Kernel died before it has finished starting.

The fix is two-fold:

1. Use KernelManager.is_alive when Manager is available (most of the time).
   This isn't sensitive to slow startup.
2. When Manager isn't available,
   wait for `is_alive()` to become True at least once before entering the wait
   for the ready-reply.
 
cc @jhamrick, @willingc

### DIFF
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -36,11 +36,12 @@ class BlockingKernelClient(KernelClient):
 
         from ..manager import KernelManager
         if not isinstance(self.parent, KernelManager):
-            # We aren't connected to a manager,
-            # so first wait for kernel to become responsive to heartbeats
+            # This Client was not created by a KernelManager,
+            # so wait for kernel to become responsive to heartbeats
+            # before checking for kernel_info reply
             while not self.is_alive():
                 if time.time() > abs_timeout:
-                    raise RuntimeError("Kernel didn't respond to heartbeats in %d seconds" % timeout)
+                    raise RuntimeError("Kernel didn't respond to heartbeats in %d seconds and timed out" % timeout)
                 time.sleep(0.2)
 
         # Wait for kernel info reply on shell channel
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -34,6 +34,15 @@ class BlockingKernelClient(KernelClient):
         else:
             abs_timeout = time.time() + timeout
 
+        from ..manager import KernelManager
+        if not isinstance(self.parent, KernelManager):
+            # We aren't connected to a manager,
+            # so first wait for kernel to become responsive to heartbeats
+            while not self.is_alive():
+                if time.time() > abs_timeout:
+                    raise RuntimeError("Kernel didn't respond to heartbeats in %d seconds" % timeout)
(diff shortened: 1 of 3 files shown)


minrk:
Marking as WIP until I verify that this actually passes nbgrader tests

minrk:
It works!

jupyter_client/blocking/client.py
```diff
@@ -34,6 +34,15 @@ def wait_for_ready(self, timeout=None):
         else:
             abs_timeout = time.time() + timeout
 
+        from ..manager import KernelManager
+        if not isinstance(self.parent, KernelManager):
+            # We aren't connected to a manager,
```
willingc: # Kernel is not yet connected to a kernel manager,

jupyter_client/blocking/client.py
```diff
@@ -34,6 +34,15 @@ def wait_for_ready(self, timeout=None):
...
+            # so first wait for kernel to become responsive to heartbeats
```
willingc: ``# so wait for kernel to respond to heartbeats

jupyter_client/blocking/client.py
```diff
@@ -34,6 +34,15 @@ def wait_for_ready(self, timeout=None):
...
+            while not self.is_alive():
+                if time.time() > abs_timeout:
+                    raise RuntimeError("Kernel didn't respond to heartbeats in %d seconds" % timeout)
```
willingc: ``...seconds and timed out"``

jupyter_client/client.py
```diff
@@ -181,9 +181,13 @@ def hb_channel(self):
 
     def is_alive(self):
         """Is the kernel process still running?"""
+        from .manager import KernelManager
+        if isinstance(self.parent, KernelManager):
+            # We were created by a KernelManager, we can ask them:
```
willingc: ``# Kernel client was created by a KernelManager, we can ask the parent KernelManager: ``

jupyter_client/tests/signalkernel.py
```diff
@@ -56,13 +56,13 @@ def kernel_info_request(self, *args, **kwargs):
...
         return super(SignalTestKernel, self).kernel_info_request(*args, **kwargs)
 
 class SignalTestApp(IPKernelApp):
     kernel_class = SignalTestKernel
     def init_io(self):
         pass # disable stdout/stderr capture
-    
+
 if __name__ == '__main__':
+    time.sleep(2)
```
willingc: Just clarifying that this is just adding a delay so the object constructor finishes before the test is launched.

willingc:
@minrk Excellent! I've left a few clarifications on the comments while reviewing. Code is +1.

minrk:
@willingc comments updated, thanks!

willingc:
Small typo. Feel free to merge @minrk. Thanks for making the  comment changes...and finding a solution!

minrk:
@willingc thanks for your diligence. Merging and releasing 4.2.2.

PR #145, opened 2016-03-23
Update wrapper kernel docs

takluyver:
`language_info` is now used instead of the separate `language` and `language_version` fields.


willingc:
@takluyver LGTM. Thanks!

PR #146, opened 2016-03-24
warn about deprecation of jupyter kernelspec install-self

minrk:
It's been deprecated since 4.0, but we haven't been showing a message when people use it.


willingc:
@minrk Thanks!

PR #151, opened 2016-03-29
Add landscape.io badge

Carreau:


minrk:
Is there a good way to say "No landscape, you are wrong"? Since most of its criticisms seem to be wrong. I wouldn't want a badge like this to seem to be encouraging pull requests to make changes that shouldn't be made.

rgbkrk:
I'm not sure what's up with the rating on `__init__.py`, you tend not to use imports there. Do we have to have a config file to make certain rules go away?

Carreau:
> Is there a good way to say "No landscape, you are wrong"? Since most of its criticisms seem to be wrong. I wouldn't want a badge like this to seem to be encouraging pull requests to make changes that shouldn't be made.

I think the `# noqa` thing should work, we should be able to turn on/off specific types of errors.

Carreau:
https://docs.landscape.io/configuration.html , and likely https://github.com/landscapeio/prospector might be useful.

PR #157, opened 2016-04-25
Make kernel shutdown timeout configuration

remram44:
Adds a `shutdown_wait_time` parameter to kernelspecs, which allows setting the timeout to a greater value (the default is one second, after which the kernel is killed).

This is useful for special kernels that do things on exit, such as closing a remote connection, saving history/state to disk, cleaning up files, etc.

Alternatively users might be able to implement this using a custom Manager, though it is more work.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -22,7 +22,7 @@ import zmq
 from ipython_genutils.importstring import import_item
 from .localinterfaces import is_local_ip, local_ips
 from traitlets import (
-    Any, Instance, Integer, Unicode, List, Bool, Type, DottedObjectName
+    Any, Float, Instance, Unicode, List, Bool, Type, DottedObjectName
 )
 from jupyter_client import (
     launch_kernel,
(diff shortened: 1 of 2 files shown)


takluyver:
I think this should probably be configurable, but I don't think it should be retrieved from the kernelspec. I'd like kernelspecs to stay as small as possible, not grow all sorts of extra fields.

minrk:
Agreed that it should be configurable, but not part of the kernelspec. A KernelManager trait should do it.

remram44:
My reasoning for making this per-kernel is that most kernels can be killed
safely. But I can look into that.

minrk:
Even IPython, which will generally have open database connections, can be problematic if you kill it, and 1 second is awfully short before giving up on a slower system. A single configurable with a default closer to 5s is probably a good compromise.

remram44:
This is now done through a KernelManager trait as suggested, `--KernelManager.shutdown_wait_time` (which defaults to 5).

jupyter_client/manager.py
```diff
@@ -67,6 +67,11 @@ def _kernel_spec_manager_default(self):
     def _kernel_spec_manager_changed(self):
         self._kernel_spec = None
 
+    shutdown_wait_time = Integer(
```
takluyver: I think a time in seconds should probably be `Float`, not `Integer` - someone may want to set it to e.g. 0.1 seconds.

remram44: Quite right -- changed with 456979d

minrk:
Thanks, @remram44!

PR #160, opened 2016-05-11
Expose prefix argument on install_kernel_spec function

takluyver:
The underlying method gained an argument, but it wasn't exposed by the function.


minrk:
Whoops!

PR #164, opened 2016-06-29
Clarifications on some of the specification

minrk:
- indicate that status=error doesn't require any fields from when status=ok to be defined
- mention that status messages can be used to fence output (with async output caveat)

closes jupyter/notebook#1576
closes jupyter/notebook#1575

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -141,6 +141,7 @@ A message is defined by the following four-dictionary structure::
     so implementers are strongly encouraged to include it.
     It will be mandatory in 5.1.
 
+
 .. _wire_protocol:
 
 The Wire Protocol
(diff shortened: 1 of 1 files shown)


minrk:
cc @willingc for docs review.

docs/messaging.rst
```diff
@@ -1052,9 +1064,26 @@ Message type: ``status``::
         execution_state : ('busy', 'idle', 'starting')
     }
 
+When a kernel receives a request and begins processing it,
+it shall immediately publish a status message with ``execution_state: 'busy'``.
+When that kernel has completed processing the request
+and has finished publishing associated IOPub messages, if any,
+it shall publish a status message with ``execution_state: 'idle'``.
+Thus, the outputs associated with a given execution shall generally arive
+between the idle and busy status messages associated with a given request.
```
takluyver: Flip round 'idle and busy' to 'busy and idle' for clarity, IMO.

docs/messaging.rst
```diff
@@ -256,8 +257,37 @@ All messages sent to or received by any IPython process should have this
...
-===========================================
+Messages on the shell (ROUTER/DEALER) channel
+=============================================
+
+.. _request_reply:
+
+Request-Reply
+-------------
+
+In general, the ROUTER/DEALER sockets follow a request-reply pattern.
```
willingc: Suggest adding a colon at the end of this line and then bulleting the individual process steps.

docs/messaging.rst
```diff
@@ -1052,9 +1064,26 @@ Message type: ``status``::
         execution_state : ('busy', 'idle', 'starting')
     }
 
+When a kernel receives a request and begins processing it,
+it shall immediately publish a status message with ``execution_state: 'busy'``.
```
willingc: it -> the kernel

docs/messaging.rst
```diff
@@ -1052,9 +1064,26 @@ Message type: ``status``::
...
+When that kernel has completed processing the request
+and has finished publishing associated IOPub messages, if any,
+it shall publish a status message with ``execution_state: 'idle'``.
+Thus, the outputs associated with a given execution shall generally arive
+between the idle and busy status messages associated with a given request.
+
+.. note::
+
+    There is a caveat for asynchronous output,
```
willingc: **A caveat for asynchronous output**

Asynchronous output may be produced after the kernel has sent the idle status message. The kernel sends an idle message when the kernel has completed processing the request. The handling of asynchronous output messages is currently undefined in the specification (reference link to spec??). The currently implemented behavior is the Jupyter Notebook ...

willingc:
Looks good to me. A few suggestions that may help the reader.

docs/messaging.rst
minrk: This file is the spec, so I don't think there's a link to add, since it would be to here.

minrk: I've updated the text to "this specification" to hopefully clarify that.

minrk:
Updated per review comments.

PR #172, opened 2016-07-22
adapt parent_header as well as header

minrk:
Might address issues validating parent_header on adapted messages

cf https://github.com/jupyter/jupyterlab/issues/470


takluyver:
Looks fine to me, I'll give others a chance to look at it before merging.

Carreau:
Thanks !

PR #175, opened 2016-07-27
MNT: Make kernel_dirs configurable.

danielballan:
Any reason not to make this configurable?

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -103,7 +103,7 @@ class KernelSpecManager(LoggingConfigurable):
         By default, all installed kernels are allowed.
         """
     )
-    kernel_dirs = List(
+    kernel_dirs = List(config=True,
         help="List of kernel directories to search. Later ones take priority over earlier."
     )
     def _kernel_dirs_default(self):


takluyver:
The search path is already customisable by other means, and more configurable options means more ways for things to go wrong. What do you want to do that would benefit from it being configurable?

danielballan:
Fair point. What is the recommended way to configure the path? My use case is to set the `kernel_dirs` in a `FormSpawner`. To do this, I've been using my own `KernelSpecManger` subclass, but configuring the `kernel_dirs` seems like a lighter way to do accomplish the same thing.

takluyver:
You can use the `JUPYTER_PATH` environment variable to add locations to the search path (it will look for a `kernels` subdirectory in each entry). The limitation is that you can only prepend entries to the default search path (akin to using PYTHONPATH to customise imports).

http://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html#data-files

minrk:
Yup, by adding to JUPYTER_PATH, you get access to adding locations for other sources of nbextensions, kernelspecs, and any other Jupyter things that might be added in the future. `kernel_dirs` is indeed not configurable on purpose. The main case this doesn't support is removing entries, as @takluyver alluded to.

danielballan:
OK, that works for me. No need for removing entries. I agree that more configurable knobs leads to more potential sources of confusion, so I'll close this PR.

PR #177, opened 2016-07-28
Make displayhook in signal test kernel app

takluyver:
Recent changes in ipykernel have made the tests fail because the IPKernelApp subclass used in tests was missing a displayhook attribute.


minrk:
Thanks for tracking this one down!

PR #180, opened 2016-08-01
Deprecate connect_request/reply

### LINKED ISSUE #179: Messaging Spec does not match IPython behaviour
The messaging spec does not match the IPython behaviour in a fairly serious way. According to the message spec...

image

But IPython sends `shell`, `iopub`, `stdin`, and `hb`, *and* includes `control`. 

Here is why:

1. Dictionary is created [here](https://github.com/ipython/ipykernel/blob/6035f9eaf561383ae7b78b9125dec235ac8c7810/ipykernel/kernelapp.py#L296) with those keys.
2.  Those ports are stored using `record_ports` [here](https://github.com/ipython/ipykernel/blob/6035f9eaf561383ae7b78b9125dec235ac8c7810/ipykernel/kernelapp.py#L366)
3. `record_ports` is a method of the kernel that just stores the ports in a variable `_recorded_ports` [here](https://github.com/ipython/ipykernel/blob/3ead7c361f7a93ee51d0e09c5128d09374641be3/ipykernel/kernelbase.py#L294)
4. `connect_request` uses `_recorded_ports` and just copies it into the message output [here](https://github.com/ipython/ipykernel/blob/3ead7c361f7a93ee51d0e09c5128d09374641be3/ipykernel/kernelbase.py#L472)

I see no evidence that `jupyter_client` even issues `connect_request` messages (searched for "connect_request" in `jupyter_client` repo and found nothing but messaging RST).

I would change the messaging spec with the new keys, but that *would* technically be a backwards incompatible change. Regardless, please let me know what my client library should do and what keys it should expect (or whether it should simply accept both).

minrk:
and update the spec to clarify that it should contain ports for all running channels.

closes #179

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -707,6 +707,11 @@ Message type: ``is_complete_reply``::
 Connect
 -------
 
+.. deprecated:: 5.1
+
+    connect_request/reply have not proved useful, and are considered deprecated.
+    Kernels are not expected to implement handlers for this message.
+
 When a client connects to the request/reply socket of the kernel, it can issue
 a connect request to get basic information about the kernel, such as the ports
 the other ZeroMQ sockets are listening on. This allows clients to only have
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -703,6 +703,7 @@ Message type: ``is_complete_reply``::
         'indent': str,
     }
 
+
 Connect
 -------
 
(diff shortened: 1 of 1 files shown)


takluyver:
+1

Carreau:
@minrk can you give me access to rtd for `jupyter_client` , build say it passed but is not deployed on RTD.

takluyver:
I've just added you to it

Issue #183, opened 2016-08-07
prompt_toolkit breaks redirect 


dsblank:
I believe that prompt_toolkit has broken this use of jupyter-client:

```bash
$ jupyter console --kernel metakernel_python < generate_help.py
Traceback (most recent call last):
  File "/usr/local/bin/jupyter-console", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/traitlets/config/application.py", line 595, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-113>", line 2, in initialize
  File "/usr/local/lib/python3.4/dist-packages/traitlets/config/application.py", line 74, in catch_config_error
    return method(app, *args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/jupyter_console/app.py", line 137, in initialize
    self.init_shell()
  File "/usr/local/lib/python3.4/dist-packages/jupyter_console/app.py", line 110, in init_shell
    client=self.kernel_client,
  File "/usr/local/lib/python3.4/dist-packages/traitlets/config/configurable.py", line 412, in instance
    inst = cls(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/jupyter_console/ptshell.py", line 252, in __init__
    self.init_prompt_toolkit_cli()
  File "/usr/local/lib/python3.4/dist-packages/jupyter_console/ptshell.py", line 404, in init_prompt_toolkit_cli
    self.pt_cli = CommandLineInterface(app, eventloop=self._eventloop)
  File "/usr/local/lib/python3.4/dist-packages/prompt_toolkit/interface.py", line 81, in __init__
    self.input = input or StdinInput(sys.stdin)
  File "/usr/local/lib/python3.4/dist-packages/prompt_toolkit/input.py", line 67, in __init__
    assert self.stdin.isatty()
AssertionError
```

takluyver:
You mean `jupyter console`, right? That's a separate package from jupyter_client. I don't think piping text into `jupyter console` was ever something we deliberately supported, but I believe it used to work and no longer does.

It should be straightforward to write a little program using `jupyter_client` that takes a filename and a kernel name, starts the desired kernel, and runs the given file in it.

dsblank:
[...]

PR #184, opened 2016-08-09
Adds jupyter-run to jupyter_client

dsblank:


jupyter_client/runapp.py
```diff
@@ -0,0 +1,146 @@
...
+            content = msg['content']
+            if msg_type == 'status':
+                if content['execution_state'] == 'idle':
+                    # idle means output is done
+                    break
+            elif msg_type == 'stream':
+                stream = getattr(sys, content['name'])
+                stream.write(content['text'])
+            elif msg_type in ('display_data', 'execute_result', 'error'):
+                if msg_type == 'error':
```
takluyver: This nesting is a bit odd - why not have `elif msg_type == 'error'` broken out separately before display_data and execute_result.

jupyter_client/runapp.py
```diff
@@ -0,0 +1,146 @@
...
+                if reply['parent_header'].get('msg_id') == msg_id:
+                    self.kernel_info = reply['content']
+                    return
+
+    def start(self):
+        self.log.debug("jupyter run: starting...")
+        super(RunApp, self).start()
+        for filename in self.filenames_to_run:
+            self.log.debug("jupyter run: executing `%s`" % filename)
+            cell = open(filename).read()
```
takluyver: As is good practice in general, we should ensure the file is closed deterministically:

```python
with open(filename) as f:
    cell = f.read()
```

jupyter_client/runapp.py
```diff
@@ -0,0 +1,146 @@
...
+            self.run_cell(cell)
+
+    def run_cell(self, cell):
```
takluyver: We already have several `run_cell` methods, so I'd rather call this something more precise. Maybe `execute_printing_output`?

jupyter_client/runapp.py
dsblank: I think this is cut and paste from Min's link:

https://github.com/dask/distributed/pull/370/files#diff-b7f744240c50427e50e9c844a13a00fdR43

so @minrk can comment on that.

jupyter_client/runapp.py
dsblank: Good catch; thanks. Pushed fix.

takluyver:
If I do `jupyter run foo.py`, and it errors, I certainly think the process should exit with code 1, so that e.g. CI knows something has gone wrong.

If I do `jupyter run foo.py bar.py` and there's an error running `foo.py`, IMO it should stop without running `bar.py`.

dsblank:
Agreed regarding stopping-on-error. I added two checks for error conditions, as I found some kernels had not yet used the error message type, so printing on stderr is also detected as an error code here.

dsblank:
Mentioning https://github.com/jupyter/jupyter_client/issues/183 as I forgot before. This should take care of that issue.

takluyver:
Sorry, I don't think printing to stderr should count as an error - it's common to write logging and things to stderr even when there's no exception condition.

We should, however, get the `execute_reply` message from the shell channel and check for `'status': 'error'` or `'status': 'aborted'` on that.

dsblank:
@takluyver I agree about counting as an error. However, there some kernels that need to be update to be able to handle this. I'll add "aborted" too.

jupyter_client/runapp.py
```diff
@@ -0,0 +1,157 @@
...
+                # TODO: remove this when all kernels use error msg_type:
+                if content['name'] == "stderr":
+                    return_code = 1
+            elif msg_type in ('display_data', 'execute_result', 'error'):
+                if msg_type == 'error':
+                    print('\n'.join(content['traceback']), file=sys.stderr)
+                    return_code = 1
+                else:
+                    sys.stdout.write(content['data'].get('text/plain', ''))
+            elif msg_type == 'aborted':
```
takluyver: This is not a message type. We need to get a message from the shell channel with `self.kernel_client.get_shell_msg`, check that it's an `execute_reply` message, and check its 'status' field.

http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-results

jupyter_client/runapp.py
```diff
@@ -0,0 +1,157 @@
...
+            content = msg['content']
+            if msg_type == 'status':
+                if content['execution_state'] == 'idle':
+                    # idle means output is done
+                    break
+            elif msg_type == 'stream':
+                stream = getattr(sys, content['name'])
+                stream.write(content['text'])
+                # TODO: remove this when all kernels use error msg_type:
+                if content['name'] == "stderr":
```
takluyver: No, really, we shouldn't do this. It breaks kernels that do signal errors correctly to make a vague affordance to kernels that don't.

I suspect that when we correctly check the `execute_reply` message in addition to this, we'll find that most kernels do work correctly.

minrk:
Exit code should only consider the `status` field of the `execute_reply`. Outputs shouldn't be relevant.

dsblank:
Removed the break condition when printing to stderr, and removed the "aborted" msg_type handler. Somehow, this needs to poll handle iopub and shell, and print output from both. But I don't think @minrk 's dask link was doing this, right?

dsblank:
Oh, I think I see my issue: not all kernels publish the execute_reply also on the iopub channel. If they do, I think that this will work. What would probably also make this more clear is:

* documenting what is done with the error message parts (evalue, and ename; that traceback is a list of lines ending with a newline, etc)
* simple_kernel should show how to handle errors

takluyver:
No kernels should publish `execute_reply` on the iopub channel.

When the kernel gets an `execute_request` message on the shell channel, it sends 0 to N output messages on iopub, which may include messages with msg_type `error` (in addition to stream, display_data, execute_result). Then it sends precisely one `execute_reply` message on the shell channel, which indicates that the requested execution is done, and includes a status field (ok/error/aborted).

So the error information can be duplicated: in an `error` message on iopub, and the `execute_reply` message on shell. This is certainly somewhat confusing, and we hope to simplify it in a future version of the message spec, but that's what we have for now.

There should be no need to poll shell or print output from it. We can poll iopub until we get the status:idle message telling us that we've seen all outputs. And then we can fetch the execute_reply message from shell.

dsblank:
@takluyver Ok, thanks. That is confusing, and isn't reflected in the code that @minrk linked to, that I can see. I don't see how the code would keep the error message from printing out twice, if it can come in two forms (one on iopub and one on shell).

I think I'll wait for someone to write the ultimate function for handling this. Then I'll call it from this code.

minrk:
@dsblank thanks! I opened #185 adding BlockingKernelClient.run, which should be the implementation of the messaging. After that gets in, you should be able to simplify this one with:

```python
reply = self.kernel_client.run(code)
exit_code = 0 if reply['content']['status'] == 'ok' else 1
```

dsblank:
Updated PR to use reply=True. Very nice!

dsblank:
I think that this can be tagged 4.5 like https://github.com/jupyter/jupyter_client/pull/185

rgbkrk:
Neat.

minrk:
@dsblank sorry for the delay on this one. Thanks!

Marked as 5.0, since that's what the next release will be.

PR #185, opened 2016-08-11
add BlockingKernelClient.execute_interactive

### LINKED ISSUE #176: add run_cell, etc. utilities
We've reimplemented running a cell against a Jupyter kernel in Python in a few places (`nbconvert --execute`, a few others).

We should probably have a few utility functions (or methods) in this repo that can be re-used.

Cases:

1. run cell and capture all output
1. record that output in nbconvert
1. redisplay that output in IPython

minrk:
Redisplays output, returns reply.

Can be used in #184

Closes #176

cc @dsblank

### DIFF
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -7,6 +7,7 @@ Useful for test suites and blocking terminal interfaces.
 
 from __future__ import print_function
 
+from getpass import getpass
 try:
     from queue import Empty  # Python 3
 except ImportError:
@@ -14,10 +15,7 @@ except ImportError:
 import sys
 import time
 
-try:
-    monotonic = time.monotonic
-except AttributeError: # py2
-    monotonic = time.clock # close enough
+import zmq
 
 from traitlets import Type
 from jupyter_client.channels import HBChannel
@@ -25,10 +23,24 @@ from jupyter_client.client import KernelClient
 from .channels import ZMQSocketChannel
 
 try:
+    monotonic = time.monotonic
+except AttributeError:
+    # py2
+    monotonic = time.time # close enough
+
+try:
     TimeoutError
 except NameError:
     # py2
     TimeoutError = RuntimeError
+
+try:
+    # py2
+    input = raw_input
+except NameError:
+    # py3
+    pass
+
 
 def reqrep(meth):
     def wrapped(self, *args, **kwargs):
@@ -144,10 +156,28 @@ class BlockingKernelClient(KernelClient):
     kernel_info = reqrep(KernelClient.kernel_info)
     comm_info = reqrep(KernelClient.comm_info)
     shutdown = reqrep(KernelClient.shutdown)
+    
+    def _handle_input(self, req):
+        """Handle an input request"""
+        content = req['content']
+        prompt = getpass if content.get('password', False) else input
+        try:
+            raw_data = prompt(content["prompt"])
+        except EOFError:
+            # turn EOFError into EOF character
+            raw_data = '\x04'
+        except KeyboardInterrupt:
+            sys.stdout.write('\n')
+            return
+
+        # only send stdin reply if there *was not* another request
+        # or execution finished while we were reading.
+        if not (self.stdin_channel.msg_ready() or self.shell_channel.msg_ready()):
+            self.input(raw_data)
 
 
     def execute(self, code, silent=False, store_history=True,
-                 user_expressions=None, stop_on_error=True,
+                 user_expressions=None, allow_stdin=None, stop_on_error=True,
                  reply=False, timeout=None,
                 ):
         """Execute code in the kernel.
@@ -178,6 +208,7 @@ class BlockingKernelClient(KernelClient):
             Some frontends (e.g. the Notebook) do not support stdin requests.
             If raw_input is called from code executed from such a frontend, a
             StdinNotImplementedError will be raised.
+            If reply=True, stdin requests from the kernel will call input / getpass.
 
         stop_on_error: bool, optional (default True)
             Flag whether to abort the execution queue, if an exception is encountered.
@@ -197,12 +228,19 @@ class BlockingKernelClient(KernelClient):
         """
         if reply and not self.iopub_channel.is_alive():
             raise RuntimeError("IOPub channel must be running to receive output")
+        if allow_stdin is None:
+            allow_stdin = self.allow_stdin
+        if allow_stdin and not self.stdin_channel.is_alive():
+            raise RuntimeError("stdin channel must be running to allow input")
         msg_id = super(BlockingKernelClient, self).execute(code,
                               silent=silent,
                               store_history=store_history,
                               user_expressions=user_expressions,
+                              allow_stdin=allow_stdin,
                               stop_on_error=stop_on_error,
         )
+        if not reply:
+            return msg_id
 
         if 'IPython' in sys.modules:
             from IPython import get_ipython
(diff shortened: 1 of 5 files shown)


minrk:
I didn't really like calling this `run`, since I think this is how `execute` should behave on the blocking client, but that would be backward-incompatible.

@takluyver what do you think about BlockingClient adding a `reply=True` kwarg to trigger waiting for and returning a reply on every req/rep message? That way, the regular methods could be used without needing to come up with synonyms.

minrk:
I like the `reply=True` better, since it's not creating new methods, just completing the loop started by the existing methods, so I went ahead and updated this PR.

dsblank:
Thanks, @minrk ... this works pretty well! One thing I am having trouble with is `input()` should that be able to work?

minrk:
@dsblank not the way I've implemented it, I didn't include support for stdin requests, but I can.

minrk:
@dsblank it handles input now.

jupyter_client/blocking/client.py
```diff
@@ -74,3 +132,179 @@ def wait_for_ready(self, timeout=None):
...
+            A dict mapping names to expressions to be evaluated in the user's
+            dict. The expression values are returned as strings formatted using
+            :func:`repr`.
+
+        allow_stdin : bool, optional (default self.allow_stdin)
+            Flag for whether the kernel can send stdin requests to frontends.
+
+            Some frontends (e.g. the Notebook) do not support stdin requests.
+            If raw_input is called from code executed from such a frontend, a
+            StdinNotImplementedError will be raised.
```
takluyver: This should also mention that with `reply=True`, input requests will trigger terminal interaction using input/getpass.

dsblank:
Works great!

jupyter_client/blocking/client.py
```diff
@@ -74,3 +132,179 @@ def wait_for_ready(self, timeout=None):
...
+            if msg_type == 'status':
+                if content['execution_state'] == 'idle':
+                    # idle means output is done
+                    break
+            elif msg_type == 'stream':
+                stream = getattr(sys, content['name'])
+                stream.write(content['text'])
+            elif msg_type in ('display_data', 'execute_result', 'error'):
+                if in_kernel:
+                    session.send(socket, msg_type, content, parent=parent_header)
```
takluyver: I'm a bit uncomfortable about this, because it's making a general method aware of a certain specific context in which it may be running.

jupyter_client/blocking/client.py
```diff
@@ -5,17 +5,75 @@
...
+        return self._recv_reply(msg_id, timeout=timeout)
+    
+    basedoc, _ = meth.__doc__.split('Returns\n', 1)
+    parts = [basedoc.strip()]
+    if 'Parameters' not in basedoc:
+        parts.append("""
+        Parameters
+        ----------
+        """)
+    parts.append("""
```
minrk: Yeah, me, too. I couldn't think of a better way, though. If you have a less gross idea, I'm all for it.

The main reason I did this is that I can't see a good way to get a nice *signature* for the wrapped functions. If I had a good signature, I'd be happier with a simpler "See KernelClient.method for more details...". It's easy to inherit the wrapped method's signature exactly, but I didn't see a good way to do that and add arguments.

jupyter_client/blocking/client.py
minrk: Fair enough. What would you prefer? What I really want is for rich Jupyter display methods to be fully redisplayed when run in a an IPython kernel, and for plaintext output to be shown in a simple terminal way when that's the case.

Do you think these should be different methods, or explicit user-provided flags?

dsblank:
@minrk This issue is also a problem for non-ipython kernels. That is, we have to monkeypatch IPython.display.display in order to get rich display. If a kernel had a special method for this, then in_kernel and non-ipython kernels could both call it.

minrk:
@dsblank good point. What do you folks think? Special method or argument for kernel-aware behavior?

minrk:
Honestly, I do think that the *default behavior* in IPython should be to redisplay rich output, so detecting IPython kernels by default still seems like the right choice to me. But I take @dsblank's point that to make it more generally useful, there should be some ways to pass explicit arguments in addition to that.

jupyter_client/blocking/client.py
takluyver: If we're willing to depend on (or bundle) a backported version of the Py3 inspect.signature machinery for Py2, it should be relatively easy to add a couple of parameters to a signature and set the `f.__signature__` attribute. IIRC, IPython's inspection will use that information even on Python 2, which is good enough for me.

dsblank:
@minrk That makes sense (there has to be a default). Just remember that some languages will make this *the* interface to their language. As long as there is some method for dealing with the issue that is not too painful, that should work. Thanks for this!

jupyter_client/blocking/client.py
takluyver: I think I would actually make this back to being a separate method. I think the kw arg works for the general `reply=True` to wait for the response, but this feels like another level on top of what the kernel client methods usually do. Maybe `execute_interactive()`?

minrk: Sounds good to me. I'll leave `execute(reply=True)` matching the other methods, without touching output.

minrk:
It is now `execute_interactive`. It now takes an `output_hook` callable that will be passed all output messages. The default is still to redisplay, including detection of IPython, but passing the hook lets the caller decide what to do with all IOPub messages. I think that should be enough that we can use this in `nbconvert --execute` now, as well as allow uses in any other kernel context.

takluyver:
I guess it should have a stdin_hook as well, then.

Pfff, this is growing complexity rapidly. I think I'd be inclined to say it's intended as a terminal method, stick to displaying text output, and maybe offer a hook just for rich output (display_data/execute_result).

minrk:
>  I think I'd be inclined to say it's intended as a terminal method, 

I would disagree with this, since the main reason I want this is to use it for interacting with kernels from notebooks. `jupyter-run` is just one entry point, but not the primary one, to me.

takluyver:
Fair enough. I'm just disappointed that something that originally seemed quite simple has now ballooned into a fair amount of extra complexity.

minrk:
Fair enough. I don't feel bad, though, as the extra complexity should also mean that it ought to be able to replace a bunch of similar code in existing projects (nbconvert, jupyter_console, dask) once this is out. So to me it's more a consolidation of code to where it belongs.

jupyter_client/blocking/client.py
```diff
@@ -74,3 +126,209 @@ def wait_for_ready(self, timeout=None):
...
+        reply: dict
+            The reply message for this request
+        """
+        if not self.iopub_channel.is_alive():
+            raise RuntimeError("IOPub channel must be running to receive output")
+        if allow_stdin is None:
+            allow_stdin = self.allow_stdin
+        if allow_stdin and not self.stdin_channel.is_alive():
+            raise RuntimeError("stdin channel must be running to allow input")
+        msg_id = super(BlockingKernelClient, self).execute(code,
```
takluyver: This doesn't need to be a `super()` call now that it's a separate method, right?

takluyver:
Other than that comment, this looks OK. I've milestoned it for 5.0 as it's clearly a new feature, but I won't get upset if we want to call it 4.4.

minrk:
Since 4.4 is about ready to ship, I'm happy calling it 4.5. I don't know that we have any breakages planned that would cause 5.0 in the forseeable future, though.

minrk:
(If there is one, 5.0 is fine, too)

takluyver:
Fair enough. I'm never entirely sure whether 4.x+1 releases are feature releases or bugfix releases.

minrk:
> I'm never entirely sure whether 4.x+1 releases are feature releases or bugfix releases.

We are on minor releases until there's a big and/or breaking change that prompt the major version bump. This doesn't happen often on the smaller repos, so I don't expect to have things like backport branches of ipykernel or jupyter_client or nbformat. It happens almost immediately on the big, fast repos like IPython and notebook, but there aren't many of those.

jupyter_client/blocking/client.py
minrk: Good catch, fixed.

minrk:
I made jupyter/nbconvert#360 to see if this works over there. It's almost enough, but there are two features over there that I haven't implemented here:

1. flushing output on its own, e.g. calling again after timeout (perhaps this should be split into a dedicated method)
2. a separate timeout that just checks the time between outputs, which better detects 'stuck' kernels vs long-running code that keeps producing output

minrk:
I can save those for another PR, though

ivan-krukov:
Is there a way that this can be used to echo input and traceback in a remote client?

minrk:
@ivan-krukov the code you gave has no output, so I wouldn't expect to see any. When I run something that has an exception, I do see it:

```python
from jupyter_client import find_connection_file, BlockingKernelClient
kc = BlockingKernelClient(connection_file = find_connection_file())
kc.load_connection_file()
kc.execute_interactive('1/0')
```

produces:

```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-5-969594e39729> in <module>()
----> 1 a/0

ZeroDivisionError: division by zero
```

ivan-krukov:
@minrk Thanks.
I was expecting to see the output in the console of the target kernel. I assume that can be done by posting to `iopub` channel (with having `ZMQTerminalInteractiveShell.include_other_output=True`)?

minrk:
Yes, the outputs are posted to IOPub, so they will appear in the console if that flag is specified, but not immediately. Since this was originally written against readline, we could only check for outputs after each prompt, and with prompt-toolkit it looks like we are now only checking on each *execution*. However, because prompt-toolkit has a proper eventloop, we should be able to handle it better than before. I opened jupyter/jupyter_console#105 to track this.

minrk:
Since 4.4 is out, I'll merge this soon unless people want more time to review.

PR #189, opened 2016-08-24
Remove docs for target_module

blink1073:
Reverts #83.

This behavior is not used currently by ipywidgets itself, and would force front end clients (e.g. JuptyerLab) to support dynamic loads of individual modules.  

Removing this allows us to only support explicitly registered comm targets.

We do have the problem that this documentation has been available since the 4.2 release of this library.

cc @jasongrout

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1226,9 +1226,7 @@ Opening a Comm produces a ``comm_open`` message, to be sent to the other side::
     {
       'comm_id' : 'u-u-i-d',
       'target_name' : 'my_comm',
-      'data' : {},
-      # Optional, the target module
-      'target_module': 'my_module',
+      'data' : {}
     }
 
 Every Comm has an ID and a target name.
@@ -1241,8 +1239,6 @@ The ``data`` key is always a dict and can be any extra JSON information used in 
 If the ``target_name`` key is not found on the receiving side,
 then it should immediately reply with a ``comm_close`` message to avoid an inconsistent state.
 
-The optional ``target_module`` is used to select a module that is responsible
-for handling the ``target_name``.
 
 Comm Messages
 -------------


jasongrout:
CC @minrk, @ellisonbg, @fperez, @SylvainCorlay (at least).

Notebook 4.2 did implement this part of the spec (https://github.com/jupyter/notebook/blob/4.2.x/notebook/static/services/kernels/comm.js#L84), even if ipywidgets didn't use it.

The problem we are running into is that this seems to be the only place in the spec that requires dynamic importing of javascript modules. This is hard to accomplish if you are trying to bundle your javascript (as we do in JupyterLab). We are discussing taking a step back and restricting dynamic javascript imports in JupyterLab for now.

Another approach is leaving this in the spec, but not releasing JupyterLab to the general public until a solution is in place to do dynamic module importing, or explicitly saying that we don't support this part of the spec.

jasongrout:
FYI, I don't know of any use of the `target_module` attribute. I checked a few repos, like bqplot, altair, matplotlib, etc., but they seem to either be using the widget infrastructure or the display infrastructure.

A valid question is: is it a good idea to offload registration of comm targets to a module system?

Another thought: `target_module` is more about providing a namespace to avoid collisions. Perhaps we can have a registration system still, but the registration system takes a module name as well as a target name.

jasongrout:
CC @sccolbert, since he was also in the conversation about JupyterLab extensions.

SylvainCorlay:
This is dependent on a decision on the jupyterlab extensions.

SylvainCorlay:
Let's wait for #728

rgbkrk:
I'm certainly in favor of removing this field. I recall (from looking at it last year) that it was coupled to the notebook implementation at the time (for ipywidgets) which made it hard for frontends like Thebe + others to handle. I really like the bootstrapping approach taken by ipywidgets in JupyterLab.

minrk:
I don't quite follow jupyterlab/jupyterlab#728's relationship to this. Is `target_module` unused? I can't find a reference to it in current widget code. I'm happy to merge or close here, depending on what's appropriate.

blink1073:
ipywidgets is not using it, and we are explicitly not including `requirejs` in JupyterLab.  We are still supporting it in [services](https://github.com/jupyterlab/services/blob/ed4176e31bf9600c5070a7410610ad03eb052e30/src/utils.ts#L398) at the moment.

minrk:
That means we should merge this, right?

blink1073:
Yes, please.

minrk:
Thanks. Sorry for being slow.

blink1073:
Thanks!

PR #190, opened 2016-08-26
Changelog for 4.4

minrk:
Includes the new .load_connection_info method in docs, as well as a few other missing methods.

can release after this.


willingc:
@minrk will do a quick pr using the changelog format that we have started using for JupyterHub repos.

PR #191, opened 2016-08-26
don't log full env when starting kernels fails

minrk:
since it could contain sensitive info like API keys.

PATH is included explicitly, but no other env is logged.

This does hide useful info like PYTHONPATH, locale, etc. but there is no way to maintain a safe language-agnostic env whitelist.

cc @jasongrout

closes jupyter/notebook#1714


minrk:
Alternative: include env *keys*, but not values. However, one could argue that even the presence of certain keys should not be disclosed.

jasongrout:
Thanks!

PR #200, opened 2016-09-13
deprecate status='abort'

minrk:
it's basically `status=error` without info and ~no kernels implement it, so deprecate in favor of `status='error'`.

cf ipython/ipykernel#190

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -288,10 +288,14 @@ All reply messages have a ``'status'`` field, which will have one of the followi
           'traceback' : list(str), # traceback frames as strings
        }
 
-- ``status='abort'``: The request has been aborted in a manner that does not result in an error
-  (e.g. the request has been canceled somehow).
-  Some kernels may set this for execution halted by signals,
-  but should only do so if there is no corresponding error raised in the evaluating code.
+- ``status='abort'``: This is the same as ``status='error'``
+  but with no information about the error.
+  No fields should be present other that `status`.
+
+.. versionchanged:: 5.1
+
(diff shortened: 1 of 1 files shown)


minrk:
In the future, I'd like to get rid of status=abort, and just have ok, error. Prior to IPython 4.4, it's possible that zero kernels implemented it, and most clients don't support it, either. If we do our revisions to make exceptions more generic, status=abort should be able to be encapsulated in a simple status=error message, even when no Exception is raised.

minrk:
We could also mark `status='abort'` as deprecated, as it has proved not useful given that clients and kernels have chosen not to implement it for years.

docs/messaging.rst
```diff
@@ -288,7 +288,11 @@ All reply messages have a ``'status'`` field, which will have one of the followi
           'traceback' : list(str), # traceback frames as strings
        }
 
-- ``status='abort'``: the task has been aborted. In this case, no additional fields should be present.
+- ``status='abort'``: The request has been aborted in a manner that does not result in an error
+  (e.g. the request has been canceled somehow).
+  Some kernels may set this for execution halted by signals,
+  but should only do so if there is no corresponding error raised in the evaluating code.
```
takluyver: I'd rather not make it depend on how the language deals with SIGINT - if we don't want kernels to do this, let's note that it's used that way but say that it's deprecated and kernels should use `'error'` for that instead.

minrk: I was trying to get at "This is like status='error' when no error info is available", which might be due to signals not corresponding to exceptions. Updated to reflect this more directly, including deprecating the value entirely.

Carreau:
+1

PR #203, opened 2016-09-21
Make launch_kernel accept additional arguments to Popen

ccordoba12:
According to its docstring, `start_kernel` accepts additional arguments for `Popen`:

https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L217

However, that's failing right now because `launch_kernel` is not doing anything with those arguments.

This PR fixes that :-)

----

By the way, I need to pass additional args to `Popen` in Spyder. so that's why I need this :-)

### DIFF
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -13,10 +13,7 @@ from traitlets.log import get_logger
 
 
 def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
-                        independent=False,
-                        cwd=None,
-                        **kw
-                        ):
+                  independent=False, cwd=None, **kw):
     """ Launches a localhost kernel, binding to the specified ports.
 
     Parameters
@@ -66,13 +63,15 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
     env = env if (env is not None) else os.environ.copy()
 
     encoding = getdefaultencoding(prefer_stream=False)
-    kwargs = dict(
+    kwargs = kw.copy()
+    main_args = dict(
         stdin=_stdin,
         stdout=_stdout,
         stderr=_stderr,
         cwd=cwd,
         env=env,
     )
+    kwargs.update(main_args)
 
     # Spawn a kernel.
     if sys.platform == 'win32':
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -24,6 +24,9 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
     stdin, stdout, stderr : optional (default None)
         Standards streams, as defined in subprocess.Popen.
 
+    env: dict, optional
+        Environment variables passed to the kernel
+
     independent : bool, optional (default False)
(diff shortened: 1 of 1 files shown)


ccordoba12:
@minrk, we need this little change for Anaconda 4.2 (to be released in a week) so we'd really appreciate if you could review it and release `4.5.1` with it. Thanks :-)

jupyter_client/launcher.py
```diff
@@ -13,10 +13,7 @@
 
 
 def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
-                        independent=False,
-                        cwd=None,
-                        **kw
-                        ):
+                  independent=False, cwd=None, **kw):
     """ Launches a localhost kernel, binding to the specified ports.
```
takluyver: The docstring should mention what happens to extra kwargs.

takluyver:
This looks OK to me, but I'd be interested to know what extra keyword args you're passing - it might point to something we need to think about for the future.

ccordoba12:
Hey Thomas, thanks a lot for the quick review :-)

We're having this problem: https://github.com/spyder-ide/spyder/issues/3427, which is fixed by passing `shell=True` to the `start_kernel` method of our kernel manager: https://github.com/spyder-ide/spyder/pull/3442/files.

This only happens when Spyder is started as a Python script on Windows (not when it's started from `spyder.exe`). I don't understand the subtleties of that difference, but this PR fixes the issue for us :-)

jupyter_client/launcher.py
ccordoba12: I extended the docstring with a mention to `**kw`.

takluyver:
I'm not entirely sure what is causing that issue, but I suspect that using `shell=True` is not really a good solution - it's the sort of band aid that fixes some problems but causes others. See jupyter/jupyter#203 for a recent example in a slightly different context.

If that's specifically what this is needed for, I'd rather work out a way to reproduce it and debug the underlying cause. I suspect that your initial suggestion of using `pythonw` is a better place to start from.

ccordoba12:
> I suspect that your initial suggestion of using pythonw is a better place to start from.

@takluyver, mmmhh, replacing `python.exe` with `pythonw.exe` in the kernel command we use in our kernel spec has the same effect as `start_kernel(shell=True)`, so I'll use that instead :-) Thanks for the suggestion.

Before I was using `pythonw.exe` to launch the Spyder application and I thought that was enough to make the kernels use it too.

----

In any case, this PR fixes what `start_kernel` promises, and so it fixes a bug that took me a couple of hours to unearth and fix ;-)

takluyver:
When ipykernel creates its own kernelspec, it will use `sys.executable` in the process that creates it. Maybe we should have a special case to switch between `python` and `pythonw` on Windows, because which one you want depends on whether your application already has a command prompt allocated (using pythonw for the kernel will prevent it from logging any output to the command prompt). That's a bit ugly and Python specific, though.

Yep, I think the PR is the right thing to do (at least in the immediate sense - I have some reservations because passing arbitrary kwargs to Popen limits flexibility for launching kernels in other ways, but that's a longer term question).

PR #206, opened 2016-09-26
reimplement uuid4 in session

minrk:
avoids call-time imports in stdlib uuid, which can cause threading issues with Python import locks.

We don't ever need actual UUIDs, we just want random strings, so we could skip UUID's `-` formatting.

closes ipython/ipykernel#193

cc @rkern

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -110,24 +110,24 @@ DONE = zmq.MessageTracker()
 # Mixin tools for apps that use Sessions
 #-----------------------------------------------------------------------------
 
-def uuid4():
-    """Generate a new UUID.
+def new_id():
+    """Generate a new random id.
 
     Avoids problematic runtime import in stdlib uuid on Python 2.
 
     Returns
     -------
 
-    uuid string (16 random bytes as hex-encoded unicode str)
+    id string (16 random bytes as hex-encoded text, chunks separated by '-')
     """
     buf = os.urandom(16)
     return u'-'.join(b2a_hex(x).decode('ascii') for x in (
-        buf[:4], buf[4:6], buf[6:8], buf[8:10], buf[10:]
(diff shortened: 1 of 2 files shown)


rkern:
Other than the failing type-assertion in the 2.7 tests, LGTM!

jupyter_client/session.py
```diff
@@ -110,14 +110,33 @@ def squash_unicode(obj):
...
+    Avoids problematic runtime import in stdlib uuid on Python 2.
+
+    Returns
+    -------
+
+    uuid string (16 random bytes as hex-encoded unicode str)
+    """
+    buf = os.urandom(16)
+    return u'-'.join(b2a_hex(x).decode('ascii') for x in (
+        buf[:4], buf[4:6], buf[6:8], buf[8:10], buf[10:]
```
takluyver: Even a random UUID is not purely random bytes; there are a couple of type fields in there that mark what kind of UUID it is:

https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29

As you say, we don't really need UUIDs, just a string that's unique in context. So if we're just generating some random bytes, that's OK, but let's not call it a UUID or format it like a UUID.

takluyver: Even a random UUID is not just random bytes; there are a couple of type fields in there to indicate what kind of UUID it is.

https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29

As you say, we don't really need UUID, just a string that's unique in context. So if we're just going to generate random bytes, let's not call it a uuid or format it like a uuid.

minrk: Good call. Called it `new_id`, and skipped the uuid-specific layout.

takluyver:
:+1:

PR #210, opened 2016-10-07
declare the 'buffers' name in the messaging doc

### LINKED ISSUE #208: What should the name be for the blob/buffers on messages?
I noticed that the [notebook websockets](https://github.com/jupyter/notebook/blob/23ca9602c2fdd0cd705ec97c2feee909e7865e67/notebook/static/services/kernels/kernel.js#L83) use `msg.buffers` as the property name for extra buffers while ijavascript's `jmp` uses `msg.blob`. This matters for frontends that are able to straddle both websockets and direct zmq - Hydrogen and nteract.

At least for me, I just want something consistent and documented. What's the right name we should use?

Related issue in jmp: https://github.com/n-riesco/jmp/issues/14

/cc @n-riesco

minrk:
closes #208

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -128,6 +128,10 @@ A message is defined by the following four-dictionary structure::
       # The actual content of the message must be a dict, whose structure
       # depends on the message type.
       'content' : dict,
+
+      # optional: buffers is a list of binary data buffers for implementations
+      # that support binary extensions to the protocol.
+      'buffers': list,
     }
 
 .. versionchanged:: 5.0
@@ -170,7 +174,7 @@ Every message is serialized to a sequence of at least six blobs of bytes:
       b'{parent_header}', # serialized parent header dict
       b'{metadata}',      # serialized metadata dict
       b'{content}',       # serialized content dict
-      b'blob',            # extra raw data buffer(s)
+      b'0x1234',          # extra raw data buffer(s)
       ...
     ]
 


rgbkrk:
Great. I've also made a PR to `jmp` for this. https://github.com/n-riesco/jmp/pull/16

docs/messaging.rst
```diff
@@ -170,7 +174,7 @@ Every message is serialized to a sequence of at least six blobs of bytes:
       b'{parent_header}', # serialized parent header dict
       b'{metadata}',      # serialized metadata dict
       b'{content}',       # serialized content dict
-      b'blob',            # extra raw data buffer(s)
+      b'0x1234',          # extra raw data buffer(s)
```
takluyver: The `0x` here suggests that it is always hex-encoded, which as far as I know is not the case.

docs/messaging.rst
```diff
@@ -128,6 +128,10 @@ A message is defined by the following four-dictionary structure::
       # The actual content of the message must be a dict, whose structure
       # depends on the message type.
       'content' : dict,
+
+      # optional: buffers is a list of binary data buffers for implementations
+      # that support binary extensions to the protocol.
+      'buffers': list,
```
takluyver: Do we want to be more precise about what this is either in Python (`bytes` objects) or in JSON (presumably base64 encoded strings)?

docs/messaging.rst
rgbkrk: Ah yeah we could add some gobbeldy gook here. It's intended to be raw binary data.

docs/messaging.rst
rgbkrk: `buffers` is raw data - on the websockets (and in node zmq), these end up being `Uint8Array`.

minrk: These aren't JSON, and are specifically *raw* binary data, not JSON-encoded at any point.

docs/messaging.rst
minrk: Added #212, making it a non-ascii binary string, to clarify that it's raw bytes.

PR #217, opened 2016-10-24
ensure timestamps are timezone-aware

minrk:
- use UTC timestamps everywhere
- use common Z format instead of +00:00 for UTC
- naïve timestamps are given local timezone on serialize/deserialize
- use python-dateutil for parsing, local timezone

This should trigger a major version bump because it affects the Python API (existing code that compares timestamps with naïve timestamps will break). It does not affect the message _protocol_ version, because the protocol's definition of timestamp behavior is unchanged, other than resolving the ambiguity that timezone-free timestamps are interpreted as local time.

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 """Utilities to manipulate JSON objects."""
 
 # Copyright (c) Jupyter Development Team.
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -38,7 +38,7 @@ def _ensure_tzinfo(dt):
     """
     if not dt.tzinfo:
         # No more naïve datetime objects!
-        warnings.warn(u"Interpreting naïve datetime as local %s. Please add timezone info to timestamps." % dt,
+        warnings.warn(u"Interpreting naive datetime as local %s. Please add timezone info to timestamps." % dt,
             DeprecationWarning,
             stacklevel=4)
         dt = dt.replace(tzinfo=tzlocal())
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -4,8 +4,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-import re
 from datetime import datetime
+import re
+import warnings
 
 from dateutil.parser import parse as _dateutil_parse
 from dateutil.tz import tzlocal
@@ -37,6 +38,9 @@ def _ensure_tzinfo(dt):
     """
     if not dt.tzinfo:
         # No more naïve datetime objects!
+        warnings.warn(u"Interpreting naïve datetime as local %s. Please add timezone info to timestamps." % dt,
+            DeprecationWarning,
+            stacklevel=4)
         dt = dt.replace(tzinfo=tzlocal())
     return dt
 
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -6,6 +6,8 @@
 import re
 from datetime import datetime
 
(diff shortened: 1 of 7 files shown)


jupyter_client/session.py
```diff
@@ -43,6 +43,13 @@
     # limiting the surface of attack
     def compare_digest(a,b): return a == b
 
+try:
+    from datetime import timezone
+    utc = timezone.utc
+except ImportError:
+    from dateutil.tz import tzutc
```
takluyver: Let's have a comment here noting that it's a Python 3/2 difference, for clarity.

jupyter_client/jsonutil.py
```diff
@@ -27,6 +29,16 @@
 # Classes and functions
 #-----------------------------------------------------------------------------
 
+def _ensure_tzinfo(dt):
+    """Ensure a datetime object has tzinfo
+    
+    If no tzinfo is present, add tzlocal
+    """
+    if not dt.tzinfo:
+        # No more naïve datetime objects!
```
takluyver: Python 2 needs the encoding header to have non-ascii characters in the code.

jupyter_client/tests/test_jsonutil.py
```diff
@@ -47,27 +52,15 @@ def test_parse_ms_precision():
...
-    now = datetime.datetime.now()
-    utcnow = now.replace(tzinfo=UTC)
-    data = dict(now=now, utcnow=utcnow)
+    naive = datetime.datetime.now()
+    data = dict(naive=naive, utc=utcnow(), withtz=naive.replace(tzinfo=gettz('CEST')))
     jsondata = json.dumps(data, default=jsonutil.date_default)
-    nt.assert_in("+00", jsondata)
-    nt.assert_equal(jsondata.count("+00"), 1)
+    nt.assert_in("Z", jsondata)
+    nt.assert_equal(jsondata.count("Z"), 1)
```
takluyver:
```
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/jupyter/jupyter_client/jupyter_client/tests/test_jsonutil.py", line 61, in test_date_default
    nt.assert_equal(jsondata.count("Z"), 1)
AssertionError: 3 != 1
```

minrk:
py2 fixes added, and updated the test to not be sensitive to the locale of the machine running them.

jupyter_client/jsonutil.py
```diff
@@ -75,7 +83,8 @@ def squash_dates(obj):
 def date_default(obj):
     """default function for packing datetime objects in JSON."""
     if isinstance(obj, datetime):
-        return obj.isoformat()
+        obj = _ensure_tzinfo(obj)
```
takluyver: AIUI, this assumes that any naive datetime is in the local timezone. Would it be reasonable to instead refuse to serialise naive datetime objects, i.e. require the calling code to make them datetime aware? There's a risk that half-smart code could use UTC timestamps without explicitly attaching the timezone info and mess things up - especially if it's written by people whose timezone aligns with UTC, and might not notice a problem.

minrk: That is the intention. I'm ever so slightly concerned about breaking things with a hard change. How about a warning that this is what's happening? I added a DeprecationWarning, but it could also be one that's on by default.

takluyver: Fair enough. Then in some future version, we could potentially drop support for passing naive datetime objects to the serialisation machinery.

takluyver:
This looks good to me. Anything we want to do before master becomes 5.0? If not, go ahead and merge.

minrk:
> This looks good to me. Anything we want to do before master becomes 5.0?

I don't think so. While this is technically a backward-incompat change, only ipyparallel should actually be affected, so I'm not too concerned.

PR #218, opened 2016-10-24
Test on Python 3,6 and nightly.

Carreau:


takluyver:
Any sign of Travis support for `3.6.0b2`?

Carreau:
Not that I am aware of.

I honestly would like to have either:
   - a 3.6, 3.7, 3.8 tags that are no-op but valid for now.
   - a 3.x+ notation that expand as soon as a new stable is out.

PR #221, opened 2016-10-31
validate kernelspec names

minrk:
allow only: ascii letters, numbers, period, hyphen, underscore.

cf https://github.com/jupyterlab/jupyterlab/issues/1018

### DIFF
--- a/docs/kernels.rst
+++ b/docs/kernels.rst
@@ -108,6 +108,9 @@ locations:
 The user location takes priority over the system locations, and the case of the
 names is ignored, so selecting kernels works the same way whether or not the
 filesystem is case sensitive.
+Since kernelspecs show up in URLs and other places,
+they are required to be simple names, only containing ASCII letters and numbers,
+and the simple separators: ``-``, ``.``, ``_``.
 
 Other locations may also be searched if the :envvar:`JUPYTER_PATH` environment
 variable is set.
--- a/docs/kernels.rst
+++ b/docs/kernels.rst
@@ -109,8 +109,7 @@ The user location takes priority over the system locations, and the case of the
 names is ignored, so selecting kernels works the same way whether or not the
 filesystem is case sensitive.
 Since kernelspecs show up in URLs and other places,
-they are required to be simple names, only containing ASCII letters and numbers,
-and the simple separators: ``-``, ``.``, ``_``.
+a kernelspec is required to have a simple name, only containing ASCII letters, ASCII numbers, and the simple separators: ``-`` hyphen, ``.`` period, ``_`` underscore.
 
 Other locations may also be searched if the :envvar:`JUPYTER_PATH` environment
 variable is set.
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -65,7 +65,8 @@ def _is_valid_kernel_name(name):
     return _kernel_name_pat.match(name)
 
 
-_kernel_name_description = "Kernel names can only contain ASCII letters and numbers and these separators: -._"
+_kernel_name_description = "Kernel names can only contain ASCII letters and numbers and these separators:" \
+ " - . _ (hyphen, period, and underscore)."
 
 
 def _is_kernel_dir(path):
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -6,6 +6,7 @@
 import io
 import json
 import os
+import re
 import shutil
 import warnings
 
@@ -19,6 +20,7 @@ from jupyter_core.paths import jupyter_data_dir, jupyter_path, SYSTEM_JUPYTER_PA
 
 
 NATIVE_KERNEL_NAME = 'python3' if PY3 else 'python2'
+
 
 class KernelSpec(HasTraits):
     argv = List()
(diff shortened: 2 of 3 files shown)


takluyver:
+1

If we'd specified this sooner, I would have disallowed `.` too, as it's confusing in file names. But I guess Julia's already using it.

jupyter_client/tests/test_kernelspec.py
```diff
@@ -141,3 +141,23 @@ def test_remove_kernel_spec_app(self):
...
+            'ipython',
+            'R',
+            'python_3',
+            'Haskell-1-2-3',
+        ]:
+            assert kernelspec._valid_kernel_name(good)
+        
+        for bad in [
+            'has space',
+            u'ünicode',
```
takluyver: Python 2 doesn't like this without the `#encoding: utf-8` magic comment

blink1073:
cc @damianavila.  It looks like the names used by the `nb_conda_kernels` exactly match the name of the environment.  All of the ones on my machine would pass the above criteria.

jupyter_client/kernelspec.py
```diff
@@ -242,10 +273,12 @@ def install_kernel_spec(self, source_dir, kernel_name=None, user=False,
         if not kernel_name:
             kernel_name = os.path.basename(source_dir)
         kernel_name = kernel_name.lower()
-        
+        if not _valid_kernel_name(kernel_name):
+            raise ValueError("Invalid kernel name %r.  %s" % (kernel_name, _kernel_name_description))
```
Carreau: Should we raise already or have a grace period ? I have kernel spec on my machine that are invalid.

minrk: There are two places right now:

1. installing new specs (I'd like to raise here)
2. listing existing specs (we can reduce this to a warning, to avoid disabling already-installed specs)

jupyter_client/kernelspec.py
```diff
@@ -54,19 +56,47 @@ def to_json(self):
...
+    for f in os.listdir(dir):
+        path = pjoin(dir, f)
+        if not _is_kernel_dir(path):
+            continue
+        key = f.lower()
+        if not _valid_kernel_name(key):
+            warnings.warn("Invalid kernelspec directory name (%s): %s",
+                _kernel_name_description, path, stacklevel=3,
+            )
+            continue
```
ivanov: If we want to have a grace period for invalid names, we need to eliminate this `continue` since it will just skip over this `key`

jupyter_client/kernelspec.py
```diff
@@ -6,12 +6,13 @@
 import io
 import json
 import os
+import re
```
ivanov: "Now you have two problems." 😹

jupyter_client/kernelspec.py
```diff
@@ -6,12 +6,13 @@
...
 import shutil
 import warnings
 
 pjoin = os.path.join
 
-from ipython_genutils.py3compat import PY3
+from ipython_genutils.py3compat import PY3, cast_bytes_py2
```
ivanov: unused import

jupyter_client/kernelspec.py
```diff
@@ -54,19 +56,47 @@ def to_json(self):
         """
         return json.dumps(self.to_dict())
 
+
+_kernel_name_pat = re.compile(r'^[a-z0-9._\-]+$', re.IGNORECASE)
+
+def _valid_kernel_name(name):
+    """Check that a kernel name is valid.
+    
+    Returns True if valid, False if not.
```
ivanov: If you rename this function to `_is_valid_kernel_name` you could get away with not having to explain the return value, and be stylistically consistent with `_is_kernel_dir` below.  A small suggestion, feel free to ignore if you like it better this way.

Carreau:
> installing new specs (I'd like to raise here)
> listing existing specs (we can reduce this to a warning, to avoid disabling already-installed specs)

Seem like a good compromise.

docs/kernels.rst
```diff
@@ -108,6 +108,9 @@ locations:
 The user location takes priority over the system locations, and the case of the
 names is ignored, so selecting kernels works the same way whether or not the
 filesystem is case sensitive.
+Since kernelspecs show up in URLs and other places,
+they are required to be simple names, only containing ASCII letters and numbers,
```
willingc: a kernelspec is required to have a simple name, only containing ASCII letters, ASCII numbers, and the simple separators: ``-`` hyphen, ``.`` period, ``_`` underscore.

jupyter_client/kernelspec.py
```diff
@@ -54,19 +56,43 @@ def to_json(self):
...
+
+_kernel_name_pat = re.compile(r'^[a-z0-9._\-]+$', re.IGNORECASE)
+
+def _is_valid_kernel_name(name):
+    """Check that a kernel name is valid."""
+    # quote is not unicode-safe on Python 2
+    return _kernel_name_pat.match(name)
+
+
+_kernel_name_description = "Kernel names can only contain ASCII letters and numbers and these separators: -._"
```
willingc: Perhaps to make visually more obvious for users, add a space between separators. `- . _` and refer to the separators by name `...these separators (hyphen, period, underscore): - . _`

minrk: thanks!

jupyter_client/kernelspec.py
minrk: Indeed! My first implementation was testing `url_quote(name) == name`, but I figured a simple character whitelist would be easier to understand. I think my regex-fu is strong enough for that, at least.

minrk:
Thanks, everyone! It now raises on new kernelspecs, but just warns when it finds existing ones that should not have been allowed.

Carreau:
Thanks all !

PR #222, opened 2016-11-02
Remove mention of switching the message serialisation format

takluyver:
We have no plans to use anything other than JSON at present.


rgbkrk:
Agreed, and we can always adapt to generic language agnostic formats later without leaving it in the commentary here. There's no way we'd switch to pickles given that we're polyglot now.

willingc:
Thanks @takluyver :smile:

PR #233, opened 2017-02-01
[WIP] add kernelspec directory to PATH when starting kernels

minrk:
- [ ] <del>verify that pip rewrites shebang lines in data_files, otherwise this is pretty pointless</del> it doesn't
- [ ] test

allows kernel command to be a script in the spec dir, which should in turn allow shebang in the script to set executable, enabling kernelspecs in Python wheels, so that pip installs can behave the same as conda installs.

This should allow us to install the IPython kernelspec on pip install, making it much harder to have ipykernel installed without  the IPython kernelspec present (largely eliminating any use of the NATIVE_KERNEL).

relevant to https://github.com/jupyter/echo_kernel/issues/3

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -242,6 +242,14 @@ class KernelManager(ConnectionFileMixin):
             # Environment variables from kernel spec are added to os.environ
             env.update(self.kernel_spec.env or {})
         
+        if os.path.isdir(self.kernel_spec.resources_dir):
+            # add kernelspec directory to PATH,
+            # so that any executables are in there
+            if 'PATH' in env:
+                path = env['PATH']
+            else:
+                path = os.getenv('PATH') or os.defpath
+            env['PATH'] = os.pathsep.join(self.kernel_spec.resources_dir, path)
         # launch the kernel subprocess
         self.log.debug("Starting kernel: %s", kernel_cmd)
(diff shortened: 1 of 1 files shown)


blink1073:
Nice!  🤞

minrk:
Turns out pip *doesn't* touch shebang in data_files, so I'd need another solution. Ideally one that's not "conda's better at everything", which is where I am right now.

blink1073:
As in, it doesn't modifying the existing shebang line to point to the correct path?

minrk:
Right. Pip does one tiny thing at install time: set the shebang line in scripts to point to the installing executable, so `#!/usr/bin/env python` becomes `#!/path/to/env/bin/python`. I was hoping to piggy-back on that by putting a script in data_files. Apparently it only does this for items identified in the metadata as scripts specifically, not any data file that happens to be a script (what I was hoping for). That's what we need for wheels to include kernelspecs: something that will be hardcoded to sys.executable at install time.

I suppose for the in-wheel case, we can rely on the env being active and use `python` as the executable.

minrk:
https://github.com/ipython/ipykernel/pull/223 installs the IPython kernelspec with ipykernel, using `python` for the executable, instead of absolute path.

minrk:
Closing because this doesn't solve the problem I thought it did.

PR #234, opened 2017-02-01
Check that buffers in a session send support the buffer protocol.

jasongrout:
Thanks to @minrk for suggesting code.

Should fix https://github.com/ipython/ipywidgets/issues/1059


minrk:
Thanks, @jasongrout!

PR #236, opened 2017-02-02
Add note on compatibility to messaging spec

takluyver:
Closes jupyter/notebook#1574

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -133,6 +133,27 @@ A message is defined by the following four-dictionary structure::
     so implementers are strongly encouraged to include it.
     It will be mandatory in 5.1.
 
+.. _msging_compatibility:
+
+Compatibility
+=============
+
+Kernels must implement the :ref:`execute <execute>` and :ref:`kernel info
+<msging_kernel_info>` messages in order to be usable. All other message types
+are optional, although we recommend implementing :ref:`completion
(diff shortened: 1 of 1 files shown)


docs/messaging.rst
```diff
@@ -133,6 +133,27 @@ A message is defined by the following four-dictionary structure::
...
+Kernels must implement the :ref:`execute <execute>` and :ref:`kernel info
+<msging_kernel_info>` messages in order to be usable. All other message types
+are optional, although we recommend implementing :ref:`completion
+<msging_completion>` if possible. Kernels do not need to send any reply for
+messages they don't handle, and frontends should provide sensible behaviour if
+no reply arrives (except for the required execution and kernel info messages).
+
+:ref:`stdin messages <stdin_messages>` are unique in that the request comes from
+the kernel, and the reply from the frontend. The frontend is not required to
+support this, but if it does not, it must set ``'allow_stdin' : False`` in
```
jasongrout: What about if a comm message from the frontend triggers an input message, but the frontend can't handle the input? I recently ran across this with the widgets. Do you think we should have an allow_stdin flag on comm messages? @minrk?

jasongrout: (this question is more about an extension of this idea, but definitely not blocking this PR)

rgbkrk:
Thanks!

PR #247, opened 2017-03-15
Fixed broken highlighting in messaging.rst

joeltg:
There was an extra space before Introduction > Heartbeat that caused the whole line to be highlighted.


Carreau:
Oops ! Thanks !

PR #250, opened 2017-03-31
protocol version is 5.1

minrk:
We forgot to update this field in the 5.0 release. Should cut a 5.0.1 with this promptly.


takluyver:
The docs also need an update - they say the current version is 5.0.

minrk:
@takluyver done, thanks

rgbkrk:
😄

minrk:
5.0.1 released

PR #251, opened 2017-04-03
Machinery for translating pipe stream into messages

takluyver:
This is the initial piece for capturing output from the kernel at the OS level, by hooking stdout and stderr up to pipes.

Kernels wanting to make use of this machinery will need to delimit the stream output resulting from executing a cell with markers containing the execute message ID:

```python
b"""
\0JUPYTER STREAM BEGIN d853e19e-012f-438b-ac71-6970f49a53b7
bar
baz
\0JUPYTER STREAM END d853e19e-012f-438b-ac71-6970f49a53b7
"""
```

These are used to set the message id in the parent header of the messages generated, so that a frontend can associate them with the relevant cell. The end marker is translated into an `EndOfOutput` object - once the frontend has seen one of these from each captured stream, plus the kernel-idle message on iopub, it knows it has all output produced while that cell was executing. The cell may have started background threads, and output from those may be lost or misdirected, but that's already the case, and I think it's unavoidable for piped data. I believe this machinery should recover if e.g. one of the markers it looks for isn't there.

Question: at present, output outside the marked blocks (after one END, before the next BEGIN) is silently discarded. Do we want to expose it to the frontend in some way, e.g. as `stream` output messages without a parent header?

### DIFF
--- a/jupyter_client/piper.py
+++ b/jupyter_client/piper.py
@@ -1,0 +1,113 @@
+from codecs import getincrementaldecoder
+from itertools import groupby
+from operator import attrgetter
+
+from .session import new_id, msg_header
+
+BEGIN_LINE = b'\0JUPYTER STREAM BEGIN '
+END_LINE = b'\0JUPYTER STREAM END '
+BUFFER_LIMIT = 512
+
+class EndOfOutput(object):
+    """Marker for the end of output triggered by a parent message."""
+    def __init__(self, parent_id):
+        self.parent_id = parent_id
+
+class Output(object):
+    def __init__(self, text, parent_id):
+        self.text = text
+        self.parent_id = parent_id
+
+def concat_outputs(outputs):
+    for (cls, parent), group in groupby(outputs, key=lambda x: (type(x), x.parent_id)):
+        if cls is Output:
+            concat_text = ''.join(o.text for o in group)
+            yield Output(concat_text, parent)
+        else:
+            # There should never be more than one consecutive end marker
+            yield list(group)[0]
+
+class PipeCapturer(object):
+    """Translate output from a pipe into Jupyter stream messages."""
+    def __init__(self, stream_name='stdout', username='', session_id=''):
+        self.stream_name = stream_name
+        self.username = username
+        self.session_id = session_id
+        self.buffer = b''
+        self.decoder = getincrementaldecoder('utf-8')(errors='replace')
+        self.current_parent_id = None
+        self._queued_output = []
+
+    def new_stream_msg(self, text, parent_id):
+        msg = {}
+        header = msg_header(new_id(), 'stream', self.username, self.session_id)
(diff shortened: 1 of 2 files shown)


minrk:
> Do we want to expose it to the frontend in some way, e.g. as stream output messages without a parent header?

I think so, or preserve the current behavior of associating it with the most recently seen parent.

jupyter_client/piper.py
```diff
@@ -0,0 +1,113 @@
...
+        self.session_id = session_id
+        self.buffer = b''
+        self.decoder = getincrementaldecoder('utf-8')(errors='replace')
+        self.current_parent_id = None
+        self._queued_output = []
+
+    def new_stream_msg(self, text, parent_id):
+        msg = {}
+        header = msg_header(new_id(), 'stream', self.username, self.session_id)
+        parent_header = msg_header(parent_id, 'execute_request',
```
minrk: I don't think we should be constructing parent_headers here. The actual request header should be arriving here somehow.

jupyter_client/piper.py
```diff
@@ -0,0 +1,113 @@
...
+                                   self.username, self.session_id)
+        msg['header'] = header
+        msg['msg_id'] = header['msg_id']
+        msg['msg_type'] = header['msg_type']
+        msg['parent_header'] = parent_header
+        msg['content'] = {u'name': self.stream_name, u'text': text}
+        msg['metadata'] = {}
```
minrk: I think this can be simplified with `msg = session.msg(msg_type, content=content, parent_header=parent_header)`

jupyter_client/piper.py
takluyver: I deliberately did it this way so that this machinery doesn't need to see the messages going to the kernel - I think it's going to be simpler to integrate into applications that way, and as far as I can think, frontends only rely on the parent message ID to know what a message is a response to.

If the application does need the full parent header, a separate piece further up the stack could keep a cache of message headers sent to the kernel, and apply them to the messages coming from this machinery by matching the IDs.

minrk: 'execute_request' is often the wrong value for msg_type, though. The parent_header needs to match the msg_type and session in the original requests (it really should match all fields, really, but those are especially important).

Maybe when a new request is seen, it gets forwarded on this pipe?

I thought with kernel nanny the Nanny was going to proxy all requests, so that it would necessarily see every request as it passes by, so the msg_id could be looked up in the recent requests that passed through.

takluyver: I've lost track of what we actually wanted from the kernel nanny, so I thought I'd try to work on this bit separately, because at least I understand it. I thought we were heading in the direction of allowing a Python based process such as the notebook server integrate the nanny functionality without running a separate process, so I'd start by doing output capturing without the nanny inside one or more of our frontends. At the moment, I can't even think of an easy way to integrate it, though, because it requires changes in multiple repositories.

PR #252, opened 2017-04-11
add Session.clone

minrk:
and use it when instantiating new Clients with Manager.client()

If multiple Clients are instantiated against the same kernel, they mustn’t share a digest_history because they could all be receiving the same iopub messages which would cause hash collisions.

relevant to https://github.com/jupyter/kernel_gateway/pull/230


Carreau:
+1, any thing delaying merge ?

PR #254, opened 2017-04-28
handle when python -OO removes docstrings

minrk:
`__doc__` is None when run with `-OO`, so don't bother trying to build the expanded docstrings for wrapped methods.

closes https://github.com/ipython/ipyparallel/issues/261


takluyver:
The test failure is pip trying to install IPython 6 on Python 2.7, and should be fixed by #255 .

PR #258, opened 2017-05-08
check that buffers are contiguous

minrk:
zmq.send will fail on non-contiguous buffers

ensures errors due to bad arguments are raise in `Session.send` instead of potentially delayed to the async callback, which is not handled well.

cf https://github.com/ipython/ipykernel/issues/245


takluyver:
I manually re-ran the Python 2 test build, but the failure appears to be consistent.

minrk:
Yup, forgot to also skip the test when I skipped the check on Python 2.

Issue #104, opened 2015-12-10
subprocesses of kernels are not killed by restart


jankatins:
Restarting a kernel only kills the direct child, but not any opened subprocesses of that child.

This is a problem if the kernel spawns multiple childprocess like the R kernel (on windows: `R - cmd - R - cmd - Rterm`) or if the python kernel opens a subprocess which sleeps.

Issue on the irkernel repo: https://github.com/IRkernel/IRkernel/issues/226

> IMO the [kernel manager](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L357) should [kill the complete process group](http://stackoverflow.com/questions/4789837/how-to-terminate-a-python-subprocess-launched-with-shell-true) and not only the process which is directly spawned by the kernel manager. 
> 
> I use restart kernel when I want resources back which I lost in the kernel (due to the stupid commands I executed) and currently I have a 4 processes running (cmd - R - cmd - Rterm) where I killed the kernel 15min ago (and it's still 25% CPU and 1GB Ram). So if I have to wait for the actual process which does the work to finish, the restart command is kind of useless :-(

reproducible example (at least on windows): put it into a cell, execute it and then restart the kernel via the notebook UI.

```
import platform
import subprocess
if platform.system() == 'Windows':
    CREATE_NO_WINDOW = 0x08000000
    subprocess_creation_flags = CREATE_NO_WINDOW
else:
    # Apparently None won't work here
    subprocess_creation_flags = 0
exec_line = u"""python -c "import time;time.sleep( 50000 )" """
p = subprocess.Popen(exec_line, shell=False,
                     stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                     creationflags=subprocess_creation_flags)
p.communicate()
```

takluyver:
I suspect that makes sense on Unix, but `killpg` is only available on Unix, so that's not going to help you on Windows.

The question you linked has a couple of answers that would apply on Windows: one with psutil, and one that involves launching `TASKKILL` in another subprocess.

janschulz:
ipyparallel has some better way to handle signals on windows: https://github.com/ipython/ipyparallel/blob/d35b4901adde6e651bb6ffc5c8aa9df0d20f1a73/ipyparallel/apps/launcher.py#L282-L288

[...]

PR #261, opened 2017-05-19
Prototype new kernel discovery machinery

takluyver:
Following jupyter/nbformat#81, this PR introduces extensible kernel finder machinery, to pave the way for different kernel pickers.

The kernel finder machinery would work like this:

* A MetaKernelFinder object holds a collection of different kernel finder objects.
* Each kernel finder has a short ID, and two methods:
  * `find_kernels()` yields pairs of id (str) and kernel attributes (dict). The attributes contain information about the kernel. At present this has no required fields, but we'll probably want to require things similar to the language info in `kernel_info_reply`.
  * `make_manager(kernel_id)` returns an instance of `KernelManager`, or a subclass, ready to start a kernel of the specified kind.
* The implementation here provides two kernel finders: one using kernel specs (the same way we already look for kernels), and one trying to import `ipykernel` in the current Python.
* Other kernel finders could find kernels in conda environments, virtualenvs, docker containers, VMs, or remote hosts. These would be distributed separately and found using entry points (not yet implemented).
* The MetaKernelManager namespaces kernels IDs according to the finder responsible for them, so an ID such as `spec/python3` means the `python3` kernel from the `spec` finder. Other qualified kernel IDs might look like: `conda/myenv` or `docker/ipython/ipykernel-nightly`.
* Specifying a kernel at e.g. a command line will use the qualified ID, e.g. `jupyter console --kernel spec/python3`. I propose that unqualified names (e.g. `python3`) are treated as if they were prefixed with `spec/`, ensuring backwards compatibility.
* Thus, specifying `pyimport/kernel` (or whatever we change that name to) will always get a kernel running on the same Python executable as the process starting it.
* When we start a kernel for a notebook, a KernelPicker object (not yet implemented) will get the IDs and attributes from `find_kernels()` and use it to select the most appropriate one.

Implementation notes:

* This involves going back on changes we had previously made to KernelManager to make it aware of kernelspecs. In this design, the kernel discovery machinery sits a level above KernelManager, rather than a level below it. In the long run, I think that we'll be glad of reduced complexity in KernelManager, but in the shorter term, it still needs to be aware of kernelspecs for backwards compatibility.
* We will likely want some finders to cache kernel info, especially if finding kernels involves launching processes or communicating with other systems. I haven't yet tried to design any mechanisms to control gathering or invalidating cache information, but I expect that we'll need these.

### DIFF
--- a/docs/kernel_providers.rst
+++ b/docs/kernel_providers.rst
@@ -1,0 +1,146 @@
+================
+Kernel providers
+================
+
+.. note::
+   This is a new interface under development. Not all Jupyter applications
+   use this yet. See :ref:`kernelspecs` for the established way of discovering
+   kernel types.
+
+By writing a kernel provider, you can extend how Jupyter applications discover
+and start kernels. To do so, subclass
+:class:`jupyter_client.discovery.KernelProviderBase`, giving your provider an ID
+and overriding two methods.
+
+.. class:: MyKernelProvider
+
+   .. attribute:: id
+
+      A short string identifying this provider. Cannot contain forward slash
+      (``/``).
+
+   .. method:: find_kernels()
+
+      Get the available kernel types this provider knows about.
+      Return an iterable of 2-tuples: (name, attributes).
+      *name* is a short string identifying the kernel type.
+      *attributes* is a dictionary with information to allow selecting a kernel.
+
+   .. method:: make_manager(name)
+
+      Prepare and return a :class:`~jupyter_client.KernelManager` instance
+      ready to start a new kernel instance of the type identified by *name*.
+      The input will be one of the names given by :meth:`find_kernels`.
+
+For example, imagine we want to tell Jupyter about kernels for a new language
+called *oblong*::
+
+    # oblong_provider.py
+    from jupyter_client.discover import KernelProviderBase
+    from jupyter_client import KernelManager
+    from shutil import which
+
+    class OblongKernelProvider(KernelProviderBase):
+        id = 'oblong'
+
+        def find_kernels(self):
+            if not which('oblong-kernel'):
+                return  # Check it's available
+
+            # Two variants - for a real kernel, these could be different
+            # environments
+            yield 'standard', {
+                'display_name': 'Oblong (standard)',
+                'language': {'name': 'oblong'},
+                'argv': ['oblong-kernel'],
+            }
+            yield 'rounded', {
+                'display_name': 'Oblong (rounded)',
+                'language': {'name': 'oblong'},
+                'argv': ['oblong-kernel'],
+            }
+
+        def make_manager(self, name):
+            if name == 'standard':
+                return KernelManager(kernel_cmd=['oblong-kernel'],
+                                     extra_env={'ROUNDED': '0'})
+            elif name == 'rounded':
+                return KernelManager(kernel_cmd=['oblong-kernel'],
+                                     extra_env={'ROUNDED': '1'})
+            else:
+                raise ValueError("Unknown kernel %s" % name)
+
+You would then register this with an *entry point*. In your ``setup.py``, put
+something like this::
+
+    setup(...
+        entry_points = {
+        'jupyter_client.kernel_providers' : [
+            # The name before the '=' should match the id attribute
+            'oblong = oblong_provider:OblongKernelProvider',
+        ]
+    })
+
+To find and start kernels in client code, use
+:class:`jupyter_client.discovery.KernelFinder`. This has a similar API to kernel
+providers, but it wraps a set of kernel providers. The kernel names it works
+with have the provider ID as a prefix, e.g. ``oblong/rounded`` (from the example
+above).
+
+::
+
+    from jupyter_client.discovery import KernelFinder
+    kf = KernelFinder.from_entrypoints()
+
+    ## Find available kernel types
+    for name, attributes in kf.find_kernels():
+        print(name, ':', attributes['display_name'])
+    # oblong/standard : Oblong (standard)
+    # oblong/rounded : Oblong(rounded)
+    # ...
+
+    ## Start a kernel by name
+    manager = kf.make_manager('oblong/standard')
+    manager.start_kernel()
+
+.. module:: jupyter_client.discovery
+
+.. autoclass:: KernelFinder
+
+   .. automethod:: from_entrypoints
+
+   .. automethod:: find_kernels
+
+   .. automethod:: make_manager
+
+Included kernel providers
+=========================
+
+``jupyter_client`` includes two kernel providers:
+
+.. autoclass:: KernelSpecProvider
+
+   .. seealso:: :ref:`kernelspecs`
+
+.. autoclass:: IPykernelProvider
+
+Glossary
+========
+
+Kernel instance
+  A running kernel, a process which can accept ZMQ connections from frontends.
+  Its state includes a namespace and an execution counter.
+
+Kernel type
+  Allows starting multiple, initially similar kernel instances. The kernel type
+  entails the combination of software to run the kernel, and the context in
+  which it starts. For instance, one kernel type may be associated with one
+  conda environment containing ``ipykernel``. The same kernel software in
+  another environment would be a different kernel type. Another software package
+  for a kernel, such as ``IRkernel``, would also be a different kernel type.
+
+Kernel provider
+  A Python class to discover kernel types and allow a client to start instances
+  of those kernel types. For instance, one kernel provider might find conda
+  environments containing ``ipykernel`` and allow starting kernel instances in
+  these environments.
--- a/docs/kernel_providers.rst
+++ b/docs/kernel_providers.rst
@@ -81,6 +81,9 @@ something like this::
         ]
     })
 
+Finding kernel types
+====================
+
 To find and start kernels in client code, use
 :class:`jupyter_client.discovery.KernelFinder`. This has a similar API to kernel
 providers, but it wraps a set of kernel providers. The kernel names it works
--- a/docs/kernel_providers.rst
+++ b/docs/kernel_providers.rst
@@ -3,12 +3,18 @@ Kernel providers
 ================
 
 .. note::
-   This is a new interface under development. Not all Jupyter applications
-   use this yet. See :ref:`kernelspecs` for the established way of discovering
-   kernel types.
+   This is a new interface under development, and may still change.
+   Not all Jupyter applications use this yet.
+   See :ref:`kernelspecs` for the established way of discovering kernel types.
+
+Creating a kernel provider
+==========================
 
 By writing a kernel provider, you can extend how Jupyter applications discover
-and start kernels. To do so, subclass
+and start kernels. For example, you could find kernels in an environment system
+like conda, or kernels on remote systems which you can access.
+
+To write a kernel provider, subclass
 :class:`jupyter_client.discovery.KernelProviderBase`, giving your provider an ID
 and overriding two methods.
 
@@ -47,8 +53,8 @@ called *oblong*::
             if not which('oblong-kernel'):
                 return  # Check it's available
 
-            # Two variants - for a real kernel, these could be different
-            # environments
+            # Two variants - for a real kernel, these could be something like
+            # different conda environments.
             yield 'standard', {
                 'display_name': 'Oblong (standard)',
                 'language': {'name': 'oblong'},
@@ -85,8 +91,9 @@ Finding kernel types
 ====================
 
 To find and start kernels in client code, use
-:class:`jupyter_client.discovery.KernelFinder`. This has a similar API to kernel
-providers, but it wraps a set of kernel providers. The kernel names it works
+:class:`jupyter_client.discovery.KernelFinder`. This uses multiple kernel
+providers to find available kernels. Like a kernel provider, it has methods
+``find_kernels`` and ``make_manager``. The kernel names it works
 with have the provider ID as a prefix, e.g. ``oblong/rounded`` (from the example
 above).
 
@@ -116,8 +123,8 @@ above).
 
    .. automethod:: make_manager
 
-Included kernel providers
-=========================
+Kernel providers included in ``jupyter_client``
+===============================================
 
 ``jupyter_client`` includes two kernel providers:
 
@@ -135,9 +142,9 @@ Kernel instance
   Its state includes a namespace and an execution counter.
 
 Kernel type
-  Allows starting multiple, initially similar kernel instances. The kernel type
-  entails the combination of software to run the kernel, and the context in
-  which it starts. For instance, one kernel type may be associated with one
+  The software to run a kernel instance, along with the context in which a
+  kernel starts. One kernel type allows starting multiple, initially similar
+  kernel instances. For instance, one kernel type may be associated with one
   conda environment containing ``ipykernel``. The same kernel software in
   another environment would be a different kernel type. Another software package
   for a kernel, such as ``IRkernel``, would also be a different kernel type.
--- a/jupyter_client/discovery.py
+++ b/jupyter_client/discovery.py
@@ -25,7 +25,7 @@ class KernelProviderBase(six.with_metaclass(ABCMeta, object)):
         pass
 
 class KernelSpecProvider(KernelProviderBase):
-    """Find kernels from installed kernelspec directories.
+    """Offers kernel types from installed kernelspec directories.
     """
     id = 'spec'
 
@@ -48,7 +48,9 @@ class KernelSpecProvider(KernelProviderBase):
 
 
 class IPykernelProvider(KernelProviderBase):
-    """Find ipykernel on this Python version by trying to import it.
+    """Offers a kernel type using the Python interpreter it's running in.
+
+    This checks if ipykernel is importable first.
     """
     id = 'pyimport'
 
@@ -83,7 +85,9 @@ class IPykernelProvider(KernelProviderBase):
 
 
 class KernelFinder(object):
-    """Manages a collection of kernel providers to find available kernels
+    """Manages a collection of kernel providers to find available kernel types
+
+    *providers* should be a list of kernel provider instances.
     """
     def __init__(self, providers):
         self.providers = providers
@@ -109,17 +113,17 @@ class KernelFinder(object):
         return cls(providers)
 
     def find_kernels(self):
-        """Iterate over available kernels.
+        """Iterate over available kernel types.
 
-        Yields 2-tuples of (id_str, attributes)
+        Yields 2-tuples of (prefixed_name, attributes)
         """
         for provider in self.providers:
             for kid, attributes in provider.find_kernels():
                 id = provider.id + '/' + kid
                 yield id, attributes
 
-    def make_manager(self, id):
-        """Make a KernelManager instance for a given kernel ID.
+    def make_manager(self, name):
+        """Make a KernelManager instance for a given kernel type.
         """
         provider_id, kernel_id = id.split('/', 1)
         for provider in self.providers:
--- a/jupyter_client/discovery.py
+++ b/jupyter_client/discovery.py
@@ -1,13 +1,14 @@
-from abc import ABC, abstractmethod
+from abc import ABCMeta, abstractmethod
 import entrypoints
 import logging
+import six
 
 from .kernelspec import KernelSpecManager
 from .manager import KernelManager
 
 log = logging.getLogger(__name__)
 
-class KernelFinderBase(ABC):
+class KernelFinderBase(six.with_metaclass(ABCMeta, object)):
     id = None  # Should be a short string identifying the finder class.
 
     @abstractmethod
--- a/jupyter_client/discovery.py
+++ b/jupyter_client/discovery.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 import entrypoints
 import logging
 
@@ -6,7 +7,23 @@ from .manager import KernelManager
 
 log = logging.getLogger(__name__)
 
-class KernelSpecFinder(object):
+class KernelFinderBase(ABC):
+    id = None  # Should be a short string identifying the finder class.
+
+    @abstractmethod
+    def find_kernels(self):
+        """Return an iterator of (kernel_name, kernel_info_dict) tuples."""
+        pass
+
+    @abstractmethod
+    def make_manager(self, name):
+        """Make and return a KernelManager instance to start a specified kernel
+
+        name will be one of the kernel names produced by find_kernels()
+        """
+        pass
+
+class KernelSpecFinder(KernelFinderBase):
     """Find kernels from installed kernelspec directories.
     """
     id = 'spec'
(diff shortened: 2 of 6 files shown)


rgbkrk:
Cool, I definitely want to be tracking this.

damianavila:
>Cool, I definitely want to be tracking this.

Me too :wink: 
Pinging people who is probably interested on this as well: @janschulz, @bollwyvl and @Cadair

Cadair:
👍

willingc:
Would be very helpful for education folks if things are simplified.

bollwyvl:
Would love to make sure that `nbconvert --execute` can take part in this!

bollwyvl:
> These would be distributed separately and found using entry points (not yet implemented).

hooray `entry_points`!

takluyver:
> Would be very helpful for education folks if things are simplified.

I can't really claim that this will simplify things, unfortunately - it's probably more complex, at least in the default case. My aim is to make it easier to combine different ways of finding kernels.

willingc:
@takluyver Oddly, I think that will simplify things on a higher level ;-)

jupyter_client/discovery.py
```diff
@@ -0,0 +1,80 @@
...
+        self.ksm = KernelSpecManager()
+
+    def find_kernels(self):
+        for name, resdir in self.ksm.find_kernel_specs().items():
+            spec = self.ksm._get_kernel_spec_by_name(name, resdir)
+            yield name, {
+                # TODO: get full language info
+                'language': {'name': spec.language},
+                'display_name': spec.display_name,
+                'argv': spec.argv,
```
janschulz: AFAIK `env` is also possible in the json file

jupyter_client/manager.py
```diff
@@ -250,6 +240,8 @@ def start_kernel(self, **kw):
             # If kernel_cmd has been set manually, don't refer to a kernel spec
             # Environment variables from kernel spec are added to os.environ
             env.update(self.kernel_spec.env or {})
+        elif self.extra_env:
+            env.update(self.extra_env)
```
janschulz: Why this change?

Or better, why not:

```
if self.extra_env:
   ...
```

janschulz:
For https://github.com/Cadair/jupyter_environment_kernels/ the most important feature(after an entry point :-) ) would be that it is possible to add env variables which are lazily loaded on kernel creation (in this case by running `source activate` in the env and outputting all env variables).

In our case, we currently accomplish that by supplying a loader function to a subclassed KernelSpec class:
https://github.com/Cadair/jupyter_environment_kernels/blob/master/environment_kernels/envs_common.py#L51 

As far as I understand the new API, the `find_kernels` return a dict (we had a implementation of a lazy loading dict, but that was basically a big hack and broke on py3.6). If that's so, I would really appreciate if the dict could be changed to an object which can lazy load the env.

Or did I misunderstod the API and that could be accomplished in the make_manager method?

janschulz:
> We will likely want some finders to cache kernel info, especially if finding kernels involves launching processes or communicating with other systems. I haven't yet tried to design any mechanisms to control gathering or invalidating cache information, but I expect that we'll need these.

The jupyter environment kernel manager has some kind of caching nowadays: https://github.com/Cadair/jupyter_environment_kernels/blob/master/environment_kernels/core.py#L110-L124 It currently caches information where kernels are, but not the env variables, which are lazily loaded. Cached values are updated every 3 minutes.

rolweber:
Is it really necessary to load all environment variables into the kernel spec? If you're going to start a kernel in a conda env, why not start it through a wrapper script that calls `source activate` before starting the kernel? Seems simpler to me than executing `source activate`, reading all environment variables, putting them into the kernel spec, and then creating a new environment with the exact same variables to start the kernel with.

jupyter_client/discovery.py
takluyver: It certainly is, but the aim of this is not necessarily to expose all the information from the kernelspec file; it's to provide a kernel picker with the information it may need to select a kernel. Maybe env is part of that, though. One of the big parts I haven't yet addressed is what information is required in this dictionary, and what standardised but optional fields we'll provide.

jupyter_client/manager.py
takluyver: KernelManager is rather complex because of the evolution of how we use it.

1. Originally it had a configurable kernel command line, which you could use to make it start a different kernel, before Jupyter (then IPython) was really aware of other kernels. But this changed it for every KernelManager in a process, so you couldn't mix kernels in one notebook server, or one Qt console.
2. Then with kernelspecs, it was meant to be given the name of a kernelspec, and it would retrieve argv and env from that. But we left the machinery for 1. in place so that people configuring `kernel_cmd` wouldn't find it broken as soon as they upgraded.
3. In this PR, I've reused the leftover machinery for 1. to allow kernel_cmd as an input, but not as a config option, and added `extra_env` to fit with this way of doing it. But I don't want to immediately break things for code using 2. This `elif` branch is for 3., and the `if` branch above is for 2. They shouldn't be mixed.

takluyver:
> As far as I understand the new API, the find_kernels return a dict (we had a implementation of a lazy loading dict, but that was basically a big hack and broke on py3.6). If that's so, I would really appreciate if the dict could be changed to an object which can lazy load the env.

> Or did I misunderstod the API and that could be accomplished in the make_manager method?

Yes, I think that this should be done in `make_manager`, or in a KernelManager subclass which `make_manager` returns.

The idea is that the dictionary contains the information needed to select which kernel to use; once you have made a choice, it calls `make_manager` to get the KernelManager object which will be used to launch the new kernel. This can use extra info which isn't necessary to select the kernel.

> It currently caches information where kernels are, but not the env variables, which are lazily loaded.

For kernel specs, I think we will need to cache the language info in a persistent manner; getting the language info requires starting a kernel, making a kernel_info request, waiting for the reply, and shutting the kernel down again. We could save this as e.g. `language.json` in each kernelspec directory.

I think the kernel finders should have some kind of way to control cached information. Questions I have:

- Should the available control be 'updated cached information now', or clear/invalidate cached information so that it will be updated next time it's needed? Or both?
- Should updating the caches happen asynchronously? What do we do if the data we want isn't there yet?
- Should we distinguish between persistent on-disk caches and ephemeral caches?

janschulz:
@rolweber
> If you're going to start a kernel in a conda env, why not start it through a wrapper script that calls source activate before starting the kernel?

Unfortunately, that has the problem that (at least on windows, which has no exec) you get a process tree like this:

Kernel starter (python) -> batch file (cmd) -> Kernel (python) 
 
This has it's own class of problems, one of them is that the current notebook can't interrupt/kill such kernels when they are busy (https://github.com/jupyter/jupyter_client/issues/104).

Getting the env variables wasn't actually the problem, the code is battle tested in xonsh. Getting the lazy loading was more work...

janschulz:
> For kernel specs, I think we will need to cache the language info in a persistent manner; getting the language info requires starting a kernel, making a kernel_info request, waiting for the reply, and shutting the kernel down again. We could save this as e.g. language.json in each kernelspec directory.

At least the environment manager will put the language tag into the json: https://github.com/Cadair/jupyter_environment_kernels/blob/master/environment_kernels/envs_common.py#L46

Why should that not work anymore?

jupyter_client/manager.py
janschulz: So If I would want to supply a env I would use 

```python
def make_manager(self, name):
    [...]
    env = get_env_for_name(name)
    return KernelManager(.., extra_env=env)
```

Sounds good to me

takluyver:
> At least the environment manager will put the language tag into the json:

That's just the language name. We'll probably want to include some more of the details from the kernel_info_reply, like the language version and the kernel implementation.

janschulz:
> That's just the language name. We'll probably want to include some more of the details from the kernel_info_reply, like the language version and the kernel implementation.

As long as that is computed when the kernel is started and not when the notebook/lab server starts to know about the kernel, as otherwise the lazy loading is again in vain.

Anyway: we currently do a full import of the kernel classes/library, when we probe for a kernel in the env. So these two informations could already be gotten at that stage.

takluyver:
That's why I was talking about persistent caching - that information probably requires starting kernel processes, but we'll probably want to know it when picking the kernel (e.g. distinguishing Python 3 vs 2).

jupyter_client/discovery.py
minrk: Maybe it should be the whole spec? Just so that we don't have to be responsible for what info the decider uses to select which kernel to run?

takluyver:
I've added a couple of tests, plus support for entry points (using the group name `jupyter_client.kernel_finders`).

These are, I think, the big remaining questions:

1. What information about kernels should be available when picking a kernel?
2. Can we rely on this information being rapidly available when we discover kernels (e.g. reading a small number of static files), or do we need mechanisms to gather the information asynchronously (e.g. starting kernels and communicating with them) and cache it?
3. If we do need asynchronous indexing, what happens if you try to use the machinery before all the information is ready? And when do we invalidate caches?

Information:

* Programming language - a Python kernel and a Julia kernel differ in an important way. I think it's reasonable to expect that this information is available from 'outside' the kernel, i.e. without starting it.
* Language version - how precisely to specify this is up to the implementation. Python 3 vs 2 is likely to be significant, but 3.5.1 vs 3.5.2 is probably not. This may be tricky to get without starting the kernel.
* Reference to environment/container? I'm thinking not - you can do equality checks with the kernel ID you get (which might look like `conda/myenv`), and I think that's the only operation that easily generalises across the different technologies.
* Space for kernel finders to stuff arbitrary information - if you want to use your custom kernel finder together with your custom kernel picker, they should be able to pass any information they care about.

What else might be relevant?

Also, using entry points ties this to Python, which would make it hard for implementations in other languages (e.g. nteract) to replicate this framework. Do we:

1. Provide interfaces (e.g. stdout, REST) to the discovery machinery which other languages can access, and expect them to rely on having Python & jupyter_client installed?
2. Discover the kernel finders using non-Python specific methods, e.g. scripts installed on PATH, which other languages can also use? This would mean significantly more complexity and worse performance for our own implementation, unfortunately.
3. Leave it up to kernel finder authors and frontend authors to write each kernel finder in each frontend language? This could be a nasty `n*m` problem, though there's a chance that m stays small (so far I only know of Python & Javascript frontends).

takluyver:
I think the best option for other languages is some variant of 1 (interface with Python & jupyter_client to discover kernels). If frontends can still start their native kernel directly (e.g. a JS kernel for nteract), but require Python to find other kernels, perhaps this is not too onerous.

The variants of this:

**1a.** As the proposal currently stands, kernels discovered through this mechanism would have to be started through Python as well.
**1b.** We could serialise the information needed to start a kernel (e.g. argv, env, cwd) and pass it back to the frontend for that to start the kernel. I expect this is more attractive for frontend authors, but it offers significantly less flexibility to kernel finders: they lose the ability to return a kernel manager which talks to another machine, or takes a pre-started kernel from a pool, or 101 other clever things we haven't thought of yet. I guess they could do still do these things by quickly creating a script and asking the frontend to run it, but that feels messy and complex.

@rgbkrk I'd welcome your input from the nteract perspective (& that goes for any other nteract devs watching this). How would you like to interface with an extensible kernel discovery mechanism? Can you see some option I'm missing?

rgbkrk:
/cc @lgeiger and @BenRussert for Hydrogen

I'll get up to date on the state of this and add new comments.

minrk:
> * Programming language

👍 to including language name

> * Language version

In practice, I've found this information *not* helpful. I don't generally want to distinguish Python 2 from 3 in my notebooks. Separating py2 and py3 may have made sense when we made that decision, but I don't think it does anymore. A Python notebook is a Python notebook, and I think language version belongs more at the same level as env/dependency-specification, and not at the same level as language name. I think we should even consider dropping the python2/python3 default kernel names for IPython, and just use `python`.

* Reference to environment/container? I'm thinking not - you can do equality checks with the kernel ID you get (which might look like `conda/myenv`), and I think that's the only operation that easily generalises across the different technologies.

👍 to pushing env info one level down into the Finder spec for the given env type

* Space for kernel finders to stuff arbitrary information - if you want to use your custom kernel finder together with your custom kernel picker, they should be able to pass any information they care about.

👍 to a place for Finders to put their own extra info.

As for where these things reside on Finders being part of the spec, my inclination would be option 3: rely on native implementations. Kernel Finders can choose to define their own specs for implementation in any language. This should be quite doable for the common cases: (base kernelspecs, envs), and I suspect of less interest for the cases where it would be more difficult. I think requiring Finders to emit argv/env misses much of the point of custom kernel finders, which can do things like launch kernels via remote APIs. These shouldn't involve subprocesses at any point. We could have a middle ground, where Finders *may* emit an argv/env which is accessible via CLI (e.g. `jupyter kernels list --json`), from which non-argv-supporting kernelspecs would be excluded. But so far, if I recall, nteract has mostly chosen to reimplement things even when Jupyter provides these CLI APIsbecause it avoids pulling in cross-language dependencies. I'll leave it to nteract folks to comment on that, though. 

Yet another CLI option is something like `jupyter kernel start --json`, which would actually handle invoking the kernel finder, manager, etc., and reply with connection info. That would probably be easier for alternative frontends to use, and further remove the need for kernel finders to emit argv or other serializable launch mechanisms.

takluyver:
I have renamed a couple of pieces:

- 'Kernel finders' are now *Kernel providers*
- `MetaKernelFinder` becomes `KernelFinder` - this is the piece that provides a single interface to a collection of kernel providers.

If we're happy with the rough shape of this, I'd like to get it in and refine it further as we try to use it. This PR is mostly adding new machinery, so hopefully it doesn't break anything we're currently doing.

takluyver:
Terminology question. Do we have wording to clearly distinguish between:

1. a kernel as a thing that can be started ("I have the bash kernel installed")
2. a kernel as a process which is processing your code ("The kernel now has the variable `a` stored")

I guess we could call 2 a "kernel instance" when we need to make the distinction, but I'm not sure about 1. Kernel class? Kernel type? Kernel program? A single kernel (1) can be running as several kernels (2) on the same system.

blink1073:
"kernel launcher", "kernel provider"?

takluyver:
Well, I hope we don't call *that* a "kernel provider", because I've just claimed that name for a higher level concept!

E.g. there could be a kernel *provider* for conda envs. That would provide multiple kernel *types* (kernel sense 1 in my previous comment) for different environments. Each kernel type (if that's what we call it) could then start multiple kernel *instances* (sense 2 in previous comment).

blink1073:
"kernel definition", as in, here is how to launch it and its associated metadata?

dsblank:
Thanks for not using MetaKernel :smile_cat: ... the naming is already a sensitive breakdown of the functionality without confusing it with [that other project](https://github.com/Calysto/metakernel).

blink1073:
Wait, haven't we already called this the "kernelspec"?

takluyver:
Good point re Metakernel, I hadn't even thought of that.

"kernelspec" is also used for the kernel.json file (and the directory containing it) which is just an advertising mechanism for a kernel (sense 1) installed elsewhere. And the kernelspec machinery is somewhat tied to launching a process locally, as it relies on argv. The concept of a type of kernel can extend to kernels that start on a remote machine.

I think so far "kernel type" is the term I can most imagine using. "Kernel class" incorrectly implies a Python class. "Kernel launcher" sounds like the parent process machinery to start a kernel. "Kernel definition" is possible, but suggests static metadata rather than a program you can launch.

E.g. "There are 7 kernel types available on your system" or "You need to install the C# kernel type to run this notebook"

willingc:
Perhaps we should do a draft doc page as part of this PR that summarizes this PR discussion and complements [`kernels.rst`](https://github.com/jupyter/jupyter_client/blob/master/docs/kernels.rst). It might help clarify functionality of the PR and naming as well. We could mark the doc file as subject to change or under active development.

takluyver:
That's a good idea. I'll work on it this evening or tomorrow.

willingc:
I wasn't thinking of anything fancy, @takluyver. More of a recap and nomenclature summary. I'll give the code a more thorough review later today or tomorrow. Thanks.

rgbkrk:
I tend to think of the running kernel as a _kernel instance_

takluyver:
Have [some docs](https://github.com/takluyver/jupyter_client/blob/aad40cb6ebb785f8ce51ee70b9872d62aef09e32/docs/kernel_providers.rst). It looks much nicer when run through Sphinx.

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
+================
+Kernel providers
+================
+
+.. note::
+   This is a new interface under development. Not all Jupyter applications
```
willingc: Perhaps mention that the interface is subject to change.

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+   use this yet. See :ref:`kernelspecs` for the established way of discovering
+   kernel types.
+
+By writing a kernel provider, you can extend how Jupyter applications discover
+and start kernels. To do so, subclass
```
willingc: Perhaps add one short sentence of why one might wish to create a kernel provider.

Add a subheading: Create a kernel provider

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+
+    class OblongKernelProvider(KernelProviderBase):
+        id = 'oblong'
+
+        def find_kernels(self):
+            if not which('oblong-kernel'):
+                return  # Check it's available
+
+            # Two variants - for a real kernel, these could be different
+            # environments
```
willingc: environment variables? pip environments? conda environments? Or implementations?

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+        def make_manager(self, name):
+            if name == 'standard':
+                return KernelManager(kernel_cmd=['oblong-kernel'],
+                                     extra_env={'ROUNDED': '0'})
+            elif name == 'rounded':
+                return KernelManager(kernel_cmd=['oblong-kernel'],
+                                     extra_env={'ROUNDED': '1'})
+            else:
+                raise ValueError("Unknown kernel %s" % name)
+
```
willingc: Subheading: Registering a Kernel Provider

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+Glossary
+========
+
+Kernel instance
+  A running kernel, a process which can accept ZMQ connections from frontends.
+  Its state includes a namespace and an execution counter.
+
+Kernel type
+  Allows starting multiple, initially similar kernel instances. The kernel type
```
willingc: Perhaps move "Allows ... instances." to end of the paragraph so definition is first.

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+  entails the combination of software to run the kernel, and the context in
+  which it starts. For instance, one kernel type may be associated with one
+  conda environment containing ``ipykernel``. The same kernel software in
+  another environment would be a different kernel type. Another software package
+  for a kernel, such as ``IRkernel``, would also be a different kernel type.
+
+Kernel provider
+  A Python class to discover kernel types and allow a client to start instances
```
willingc: NIce!

docs/kernel_providers.rst
willingc: Probably good to specify since environment is a pretty overloaded term.

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+            # The name before the '=' should match the id attribute
+            'oblong = oblong_provider:OblongKernelProvider',
+        ]
+    })
+
+Finding kernel types
+====================
+
+To find and start kernels in client code, use
+:class:`jupyter_client.discovery.KernelFinder`. This has a similar API to kernel
```
willingc: "This...providers." Let's be a bit more detailed and concrete here before comparing to similar API. Perhaps recap at start of section what is a kernel type.

docs/kernel_providers.rst
```diff
@@ -0,0 +1,149 @@
...
+
+.. autoclass:: KernelFinder
+
+   .. automethod:: from_entrypoints
+
+   .. automethod:: find_kernels
+
+   .. automethod:: make_manager
+
+Included kernel providers
```
willingc: "Basic Kernel Providers included in Jupyter Client"

I'm not totally happy with my suggestion above but I think we need a bigger distinction between prior example section and this reference section.

setup.py
```diff
@@ -93,6 +94,10 @@ def run(self):
         'console_scripts': [
             'jupyter-kernelspec = jupyter_client.kernelspecapp:KernelSpecApp.launch_instance',
             'jupyter-run = jupyter_client.runapp:RunApp.launch_instance',
+        ],
+        'jupyter_client.kernel_providers' : [
+            'spec = jupyter_client.discovery:KernelSpecProvider',
+            'pyimport = jupyter_client.discovery:IPykernelProvider',
```
willingc: Are we being consistent with the id before = as described in entrypoints doc section.

willingc:
> Have some docs. It looks much nicer when run through Sphinx.

[Rendered doc page](http://test-jupclient.readthedocs.io/en/takluyver-discovery/kernel_providers.html) on my test build. For the convenience of others.

setup.py
takluyver: Yep:

https://github.com/takluyver/jupyter_client/blob/c09b8aced5912901b4e96e11d9d4eeb3346b4fd7/jupyter_client/discovery.py#L30

https://github.com/takluyver/jupyter_client/blob/c09b8aced5912901b4e96e11d9d4eeb3346b4fd7/jupyter_client/discovery.py#L55

I can't work out when Github does inline code previews and when it doesn't!

docs/kernel_providers.rst
takluyver: I've opted not to add a subheading here because it feels like part of the previous section to me. I've implemented all your other suggestions.

takluyver:
Thanks @willingc !

docs/kernel_providers.rst
willingc: Sure no problem.

takluyver:
Does anyone else want to review this? If not, I'll push the button in another day or two.

takluyver:
Button being pressed.

jankatins:
Is the discovery mechnism actually used anywhere? E.g. the notebook server? As far as I see the notebook still uses `jupyter_client.kernelspec.KernelSpecManager` and [that one](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/kernelspec.py#L115) does not yet use the new discovery mechnism :-( Is that correct?

takluyver:
No, you're right, it's not used anywhere yet. We're still hashing out the details of how it works in different situations. #308 is a follow up to this PR.

takluyver:
If you've got time to experiment with making that work on top of PR #308, that could be useful - we're currently trying to figure out if we've got a useful API design by making prototype things on top of it (I've done [an ssh provider](https://github.com/takluyver/jupyter_ssh_kernels) and [a docker provider](https://github.com/takluyver/jupyter_docker_kernels)).

Kernel specs become one default provider for the new discovery system, so no, we wouldn't convert KernelSpecManager to use the new system.

Issue #259, opened 2017-05-16
incorrect cursor_pos for completion requests with non-BMP Unicode characters


stevengj:
As discussed in JuliaLang/IJulia.jl#541, it appears that Jupyter is sending an incorrect `cursor_pos` for completion requests when the string contains non-BMP Unicode characters.

For example, if I try to tab-complete the a cell containing `𝐚𝐚𝐚𝐚𝐚`, it sends `cursor_pos=10` even though the string has just 5 characters.

For these these non-BMP characters, the UTF-16 encoding used by JavaScript uses two 16-bit code units per character, rather than one as for most characters.  So, maybe JavaScript's Unicode encoding is leaking into the messages (contrary to the [spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html), which explicitly says that `cursor_pos` is in "unicode characters").

Carreau:
Hum, thanks for catching this.
You would have asked me, I would have said it was bytes, not characters. 

(BTW heard good things about the talk you gave today, seem like you never stop to code , even during conferences. )

minrk:
This is indeed a bug in CodeMirror itself, where `getCursor()` counts `𝐚` as having length 2. Most other multi-codepoint characters I have tried correctly have length 1 (e.g. `é` or あ). I think CodeMirror just needs to learn about some more combining characters. From [this issue](https://github.com/codemirror/CodeMirror/issues/1813), CodeMirror has to handle all combining characters, etc. itself since the underlying javascript is no help. My guess is this is just a unicode combining-range that CM doesn't know about yet.

stevengj:
@minrk, I think you are confusing [combining characters](https://en.wikipedia.org/wiki/Combining_character) with [surrogate pairs](https://en.wikipedia.org/wiki/UTF-16).

JavaScript (hence CodeMirror) uses the UTF-16 encoding to store Unicode strings: each string is stored as an array of 16-bit "code units", so *every* character is "multi-byte".   Originally, people thought 16 bits would be enough to encode all of Unicode, and hence each code unit would equal one character.  Unfortunately, people eventually realized that 16 bits was not enough, and newer characters (non-[BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)) characters) are encoded as two 16-bit code units (a "surrogate pair").

[...]

PR #262, opened 2017-05-23
describe cursor_pos ambiguity and bump protocol to 5.2

minrk:
cursor_pos with characters after 0x10000 is officially ambiguous with respect to surrogate pairs in versions 5.1 and earlier due to a widespread bug in javascript frontends.

Thus bumps the protocol to 5.2 with the sole change that cursor_pos is no longer ambiguous (i.e. clients that correctly implement cursor_pos in 5.1 can bump to 5.2 with no other changes).

cf #259

cross-refs for frontend implementations:

- https://github.com/sagemathinc/cocalc/issues/1994
- https://github.com/jupyterlab/jupyterlab/issues/2255
- https://github.com/nteract/nteract/issues/1706
- https://github.com/jupyter/notebook/issues/2509

cc @stevengj @rgbkrk @Carreau @blink1073 @williamstein

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -21,7 +21,7 @@ Versioning
 
 The Jupyter message specification is versioned independently of the packages
 that use it.
-The current version of the specification is 5.1.
+The current version of the specification is 5.2.
 
 .. note::
    *New in* and *Changed in* messages in this document refer to versions of the
@@ -547,6 +547,14 @@ Message type: ``inspect_request``::
     ``name`` key replaced with ``code`` and ``cursor_pos``,
     moving the lexing responsibility to the kernel.
 
+.. versionchanged:: 5.2
+
+    Due to a widespread bug in many frontends, ``cursor_pos``
+    in versions prior to 5.2 is ambiguous in the presence of "astral-plane" characters.
+    In 5.2, cursor_pos **must be** the actual encoding-independent offset in unicode codepoints.
+    See :ref:`cursor_pos_unicode_note` for more.
+
+
 The reply is a mime-bundle, like a `display_data`_ message,
 which should be a formatted representation of information about the context.
 In the notebook, this is used to show tooltips over function calls, etc.
@@ -594,6 +602,13 @@ Message type: ``complete_request``::
 
     ``line``, ``block``, and ``text`` keys are removed in favor of a single ``code`` for context.
     Lexing is up to the kernel.
+
+.. versionchanged:: 5.2
+
+    Due to a widespread bug in many frontends, ``cursor_pos``
+    in versions prior to 5.2 is ambiguous in the presence of "astral-plane" characters.
+    In 5.2, cursor_pos **must be** the actual encoding-independent offset in unicode codepoints.
+    See :ref:`cursor_pos_unicode_note` for more.
 
 
 Message type: ``complete_reply``::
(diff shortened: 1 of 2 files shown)


docs/messaging.rst
```diff
@@ -1370,12 +1385,44 @@ handlers should set the parent header and publish status busy / idle,
...
-Missing things include:
+.. _cursor_pos_unicode_note:
+
+``cursor_pos`` and unicode offsets
+----------------------------------
+
+Many frontends, especially those implemented in javascript,
+reported cursor_pos as the interpreter's string index,
+which is not the same as the unicode character offset if the interpreter uses UTF-16 (e.g. javascript or Python 2 on macOS),
+which stores astral-plane characters such as ``𝐚`` as surrogate pairs,
```
Carreau: precise  `U+1D41A` ? Because I'm going to assume somewhere it's not going to render correctly ?

docs/messaging.rst
```diff
@@ -1370,12 +1385,44 @@ handlers should set the parent header and publish status busy / idle,
...
+Kernels may choose to expect the UTF-16 offset from requests implementing protocol 5.1 and earlier, in order to behave correctly with the most popular frontends.
+But they should know that doing so *introduces* the inverse bug for the frontends that do not have this bug.
+
+Known affected frontends:
+
+- Jupyter Notebook
+- JupyterLab
+- nteract
+- CoCalc
+- Jupyter Console and QtConsole with Python 2 on macOS
```
Carreau: Should we precise versions ?

docs/messaging.rst
```diff
@@ -547,6 +547,14 @@ Message type: ``inspect_request``::
     ``name`` key replaced with ``code`` and ``cursor_pos``,
     moving the lexing responsibility to the kernel.
 
+.. versionchanged:: 5.2
+
+    Due to a widespread bug in many frontends, ``cursor_pos``
+    in versions prior to 5.2 is ambiguous in the presence of astral-plane.
```
stevengj: missing a noun:
```
in the presence of "astral-plane" characters
```
(added scare quotes around "astral-plane" since this is a slang term)

stevengj: similar below

stevengj: maybe also link the github issue for the technical details?

docs/messaging.rst
stevengj: probably on Windows, too

stevengj:
In `kernel_info_reply` the kernel sends what version of the protocol it implements.  Is the highest supported protocol version sent in the reverse direction (from frontend to kernel) anywhere?

docs/messaging.rst
minrk: Confirmed on Windows Python 2 as well.

docs/messaging.rst
minrk: Good idea

minrk:
> Is the highest supported protocol version sent in the reverse direction (from frontend to kernel) anywhere?

This ought to be the 'version' field in the request header, I believe.

blink1073:
Currently `5.0` in [JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/aa825677987983d798920b42879d7657d7f7ccd3/packages/services/src/kernel/messages.ts#L30) and [Notebook 5.0](https://github.com/jupyter/notebook/blob/cd47dac0ff5ddae4261eea4540ec5d0194984439/notebook/static/services/kernels/kernel.js#L83).

stevengj:
@minrk, currently `kernel_info_request` is documented as `content = {}`.  Adding a version field would be a good addition.  Can that be added to protocol version 5.2 as well?

minrk:
Came back to this. I believe I've addressed comments now, and issued PRs against jupyter/notebook and jupyterlab to handle the transform and bump the version to 5.2.

docs/messaging.rst
```diff
@@ -1370,12 +1385,48 @@ handlers should set the parent header and publish status busy / idle,
...
+----------------------------------
+
+Many frontends, especially those implemented in javascript,
+reported cursor_pos as the interpreter's string index,
+which is not the same as the unicode character offset if the interpreter uses UTF-16 (e.g. javascript or Python 2 on macOS),
+which stores "astral-plane" characters such as ``𝐚 (U+1D41A)`` as surrogate pairs,
+taking up two indices instead of one, causing a unicode offset
+drift of one per astral-plane character.
+Not all frontends have this behavior, however,
+and after JSON serialization information about encoding was used
```
stevengj: `about which encoding was used`

stevengj:
cross-ref: nteract/hydrogen#807

docs/messaging.rst
minrk: Good catch, thanks!

PR #267, opened 2017-06-23
remove incorrect number of sockets in messaging spec

rolweber:
The text still said "four sockets" to introduce a list of five. I've rephrased without a number, so that this error will not reappear when additional sockets are added in the future :-)


takluyver:
Thanks!

Issue #268, opened 2017-06-23
Contributor Documentation?


rolweber:
I'd like to develop and contribute some code changes, but I'm not super familiar with Python yet. I cloned the git repo, I see there is code, and there are tests. But I didn't see a Makefile, or instructions on how to run the test suite. Basically some "First Steps for Contributors", like this page for the kernel gateway:
https://jupyter-kernel-gateway.readthedocs.io/en/latest/devinstall.html

Is the jupyter_client project so similar to other Jupyter projects that I should look elsewhere for instructions? The Python part of the [notebook instructions](https://github.com/jupyter/notebook/blob/master/CONTRIBUTING.rst), maybe? But notebook is a server, whereas jupyter_client seems more like a library, which wouldn't be "launched" in the same way.

If you can point me in the right direction, I'll write up and contribute my initial steps.

takluyver:
You can run the tests with `py.test` - we should put that in the README.

It follows the standard patterns for open source Python packages, so you can do a development install with `pip install -e .`.

Carreau:
> If you can point me in the right direction, I'll write up and contribute my initial steps.

> You can run the tests with py.test - we should put that in the README.

That indeed would be a good thing to add to the readme, maybe also with instruction on what else need to be installed for testing. Mainly: ipykernel, ipython, nose_warnings_filters.

I guess `pytest` is also required (should it be in setup extra-requires[test] section ?). So I guess for now we should likely write in the README to that to run test one need to

`$ pip install -e .[test] pytest` adn then run `py.test`.

@rolweber do you want to add that to the readme ? If you are not familiar with Git/GitHub either feel free to ask questions ! Thanks you already for opening this issue and looking forward to your contribution !

rolweber:
Thanks, that was exactly the information I needed to get started. See PR #271 for my notes.

`pytest` is already pulled in by the `.[test]`dependencies.

Carreau:
> rolweber closed this an hour ago

Thanks, TIP: one thing you can do is say `Closes #268` in your commit(s) or any other relevant issues and GitHub will autoclose on merge.

PR #271, opened 2017-06-26
#268: README on setting up the development environment

rolweber:
see #268

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,5 @@
 # Contributing
 
 We follow the [IPython Contributing Guide](https://github.com/ipython/ipython/blob/master/CONTRIBUTING.md).
+
+See the [README](https://github.com/jupyter/jupyter_client/blob/master/README.md) on how to set up a development environment.
--- a/README.md
+++ b/README.md
@@ -84,5 +84,5 @@ Use `make` to build the documentation in HTML:
 
 Point your browser to the following URL to access the generated documentation:
 
-_file:///my/projects/jupyter_client/docs/_build/html/index.html_
+_file:///my/projects/jupyter\_client/docs/\_build/html/index.html_
 
--- a/README.md
+++ b/README.md
@@ -62,18 +62,18 @@ if you are using an Anaconda distribution, with `conda`.
 
 ## Run the Test Suite
 
-Use `pytest` to run the test suite:
+Use `py.test` to run the test suite:
 
     cd /my/projects/jupyter_client/
-    pytest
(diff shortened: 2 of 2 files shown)


rolweber:
Just noticed that both `pytest` and `py.test` work, so the last commit was superfluous. But since both comments in the issue used the spelling with dot, I assume that is the preferred one.

takluyver:
Thanks for opening this :-)

I think it's worth having some brief notes about how to run the tests and things, but I'm a bit concerned about putting this much information there. Most of this info is essentially the same for all our repos, and for many other Python packages out there, so we could end up with it duplicated in a lot of places, and duplicated content is at risk of getting out-of-date when things change.

Maybe we could simplify this to a couple of brief notes on how to do a development install and run the tests, and link to the [Jupyter contributor information](http://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html) for more details. Those docs may have room for improvement as well, but at least they're just in one [place](https://github.com/jupyter/jupyter/tree/master/docs/source).

rolweber:
I basically collected the handful of commands that I expect to need on a regular basis, then put in some prose and external links to make it more readable. If I have to study the general developer documentation first to filter out duplicate information, that's gonna take a few days. Our README should definitely have pointers there, I agree.

Looking at the documentation part of [Getting Started](http://jupyter.readthedocs.io/en/latest/contrib_docs/getting-started.html#edit-the-documentation-source-file)...
- "Source files for projects are typically found in the project’s docs/source directory."  
  * Jupyter Client has them in `docs` rather than `docs/source`. We need to mention that.
- "In addition, you may need the following packages: sphinxcontrib-spelling, sphinx_rtd_theme, nbsphinx, pyenchant, recommonmark and jupyter_sphinx_theme, which can be installed via `pip install sphinxcontrib-spelling sphinx_rtd_theme nbsphinx pyenchant recommonmark jupyter_sphinx_theme`."  
  "If you are on Linux, you may also need to install the Enchant C library by running `sudo apt-get install enchant`."
  * Do you expect a beginner to even _try_ after reading this? I gave up way before `sudo`.

I recommend to put the frequently needed commands directly into the README for copy&paste, and add pointers to more extensive documentation for those willing to take the next steps. That should keep the learning curve smooth.

I'll take another stab at it later this week.

rolweber:
I've trimmed down the prose and added relevant links to the Jupyter Guides. Please have another look.

Carreau:
Thanks a lot, that's a big improvement. If we want to do better we can still refine that in subsequent PRs.

PR #274, opened 2017-07-04
Add a `metadata` field to kernelspecs.

### LINKED ISSUE #273: Support metadata in kernel specs
Currently, there's no easy way to track metadata about a kernel beyond what language it supports. Having a display name is helpful, but it'd be nice to allow for programmatic access for more information about the kernel.

For example, internally we use Jupyter in a context where different backends have different functionality available (eg access to different libraries or hardware). It'd be nice to allow a FE to filter the list of available backends if we know a certain library is needed.

Similarly, as new versions of the Jupyter protocol roll out, some FEs will be in a situation where they might need to communicate with backends which speak different protocol versions. Having access to this info programmatically would be nicer than a trial-and-error approach.

/cc @rgbkrk

craigcitro:
Currently, the only way for a kernel to offer additional information about its
capabilities is to encode information into the kernel name or display name;
this information can be useful to clients, especially for things like
filtering a list of kernels.

This adds support for a new kernelspec dict field, `metadata`. This allows
kernels to add additional information, which clients can then consume as
needed.

Fixes #273. /cc @rgbkrk

### DIFF
--- a/docs/api/kernelspec.rst
+++ b/docs/api/kernelspec.rst
@@ -25,6 +25,11 @@ kernelspec - discovering kernels
       The name of the language the kernel implements, to help with picking
       appropriate kernels when loading notebooks.
 
+   .. attribute:: metadata
+
+      Additional kernel-specific metadata; clients can use this as needed,
+      for instance to aid in kernel selection and filtering.
+
    .. attribute:: resource_dir
 
       The path to the directory with this kernel's resources, such as icons.
--- a/docs/kernels.rst
+++ b/docs/kernels.rst
@@ -135,6 +135,8 @@ JSON serialised dictionary containing the following keys and values:
 - **env** (optional): A dictionary of environment variables to set for the kernel.
   These will be added to the current environment variables before the kernel is
   started.
+- **metadata** (optional): A dictionary of additional attributes about this
+  kernel; used by clients to aid clients in kernel selection.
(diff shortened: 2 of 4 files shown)


takluyver:
:+1: I think this will also be useful for the machinery I'm proposing in #261.

craigcitro:
Ah, @takluyver, this does seem like it'd be exactly the information you'd want to return as kernel info in `find_kernels()` in that PR. 

That actually brings up an interesting parallel I hadn't realized at first -- internally at Google, we don't use conda/virtualenv/etc at all, but the "different libraries" I alluded to in #273 are a *really* close analogue to different virtualenvs or conda environments, where each has a fixed set of libraries installed (possibly with particular build configurations for C++ libraries, that sort of thing). So I suspect we're actually solving roughly the same problem -- meaning I can hopefully bootstrap off some of what you're doing in #261. 😉

rgbkrk:
Since so far there is agreement on this one, its been open a few days, and this fits in with the way we do metadata elsewhere, I'm going to merge this with the acknowledgement that metadata is going to be the wild west just like in the notebook format.

rgbkrk:
Like we've done in the notebook format (especially per cell metadata) we've explicitly stated that usage of metadata should be namespaced. That could be worth adding on here too.

PR #275, opened 2017-07-05
Switch to `tox` for testing, docs, and coverage.

craigcitro:
This just moves all manual steps into `tox`:
* tests are run in `tox`
* Travis uses `tox` rather than a separate config
* docs and coverage are handled via `tox`
* updated `README` to mention `tox` where appropriate.

I made two other small cleanups in the process:
* small dep cleanup (`pytest` was required but never mentioned)
* added two more entries to `.gitignore`

I've also used 3.6 as the "default" version, in the sense that coverage and
doc building use 3.6 as the base version, rather than 2.7.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ install_requires = setuptools_args['install_requires'] = [
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['ipykernel', 'ipython', 'pytest'],
+    'test': ['ipykernel', 'ipython', 'mock', 'pytest', 'pytest-warnings'],
 }
 
 if 'setuptools' in sys.modules:
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ install_requires = setuptools_args['install_requires'] = [
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['ipykernel', 'ipython', 'mock', 'pytest', 'pytest-warnings'],
+    'test': ['ipykernel', 'ipython', 'mock', 'pytest'],
 }
 
 if 'setuptools' in sys.modules:
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ __pycache__
 \#*#
 .#*
 .coverage
+.cache
+absolute.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,27 @@
 language: python
-python:
-    - "nightly"
-    - '3.6-dev'
-    - 3.5
-    - 3.4
-    - 3.3
-    - 2.7
+matrix:
+  include:
+    - python: 3.6
+      env: TOX_ENV=docs
+    - python: 3.6
+      env: TOX_ENV=cover
+    - python: 2.7
+      env: TOX_ENV=py27
+    - python: 3.3
+      env: TOX_ENV=py33
+    - python: 3.4
+      env: TOX_ENV=py34
+    - python: 3.5
+      env: TOX_ENV=py35
+    - python: 3.6-dev
+      env: TOX_ENV=py36
+  allow_failures:
+    - python: nightly
+      env: TOX_ENV=py37
 sudo: false
 install:
-  - pip install --upgrade setuptools pip
-  - pip install --upgrade --pre -e .[test] pytest-cov pytest-warnings codecov
+  - pip install --upgrade tox
 script:
-  - py.test --cov jupyter_client jupyter_client
+  - tox -e "${TOX_ENV}"
 after_success:
-  - codecov
-matrix:
-    allow_failures:
-        - python: nightly
+  - if [[ "${TOX_ENV}" == "cover" ]]; then tox -e codecov; fi
--- a/README.md
+++ b/README.md
@@ -26,19 +26,18 @@ and download the dependencies of code and test suite by executing:
     pip install -e .[test]
     py.test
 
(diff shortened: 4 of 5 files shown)


setup.py
```diff
@@ -82,7 +82,7 @@
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['ipykernel', 'ipython', 'pytest'],
+    'test': ['ipykernel', 'ipython', 'mock', 'pytest', 'pytest-warnings'],
```
willingc: I believe that 'pytest-warnings' is deprecated and the functionality is included in pytest >= 3.1. Is there a particular reason to include it here?

Carreau:
Thanks @willingc for pinging me and thanks @craigcitro for the pull-request. 

I just want to note before hand that I'm biased against tox, having had several bead experience with it.  So I'll try to stay a bit away from making a decision in here.

For example this PR does not work  on my machine and have error messages interleaved with success messages. The issue are about dll/so which I'm unsure how to fix and I am concern to expose that to our users. If it's tox not playing well with conda then this is really concerning as many of our users are using conda. We should figure that out. 

It is true the we should make it easier to run test locally. Though I'm unsure the we want to run the all test-suite on all interpreters locally, it can takes quite some time. Also I am not a big fan on isolating all tests in tox, as I may have other dev packages installed locally and want to test things on my current env, with current patch to all the project I'm using. I guess it's just a question of choice which is documented so I guess it's fine. 

Thanks you for configuring it to run separate travis Job, that is much better than other project running one single tox with all the jobs, in particular to restart a single one it will be useful. 

Building the doc is a nice touch which is appreciated. 

> I've also used 3.6 as the "default" version, in the sense that coverage and
doc building use 3.6 as the base version, rather than 2.7.

Yes, thanks you we missed that. 

> small dep cleanup (pytest was required but never mentioned)
> added two more entries to .gitignore

Yes, thanks, I didn't even remembered we created an absolute.json file in the test and was wondering what it was.

So globally positive to get that in, I would really like to figure out why tox fails out of the box on my machine.

Carreau:
Well looking around and using google it seem that indeed tox does not play well with conda. 

1) you can't `conda install tox`
2) tox requires virtualenv
3) apparently depending on the platform (?) you may need to use conda-virtualenv or the env creation may fail. [SO accepted answer and first comment](https://stackoverflow.com/questions/30555943/is-it-possible-to-use-tox-with-conda-based-python-installations)  

So at least on MacOS + Conda you need to `pip install tox`, `pip uninstall virtualenv` and then `conda install virtualenv` or tox may fail.

At least I've figured out why tox fails on my machine – and now I want to be cautious with that. It seem that pip install virtualenv only fails *sometime*, which is a bit scary.

If a couple of conda / osx  users could just test that and tell me that they have no issue I would be reassured and put my local failure as a weird issue on my system.

Thanks a lot for taking the time to improve our testing infrastructure !

willingc:
> If it's tox not playing well with conda then this is really concerning as many of our users are using conda. We should figure that out.

I was using pip so it's entirely possible that it may not play nicely with conda. 

FYI, @craigcitro Min is on vacation this week. Ideally, I would like his input as well as Thomas and Matthias' views. Thanks so much for the PR.

Carreau:
I've added them as reviewer :-)

takluyver:
To be honest, tox is one of those things that I suspect is a good idea in principle, but whenever I come across a project using it, I think "myeh, another layer to understand".

I'm -0.5 because I feel like tox adds more complexity than it takes care of. But if other people want it, I'm OK with it going in.

setup.py
craigcitro: Nope -- I was just blindly copy-pasting it in from elsewhere. 😁 Dropped.

craigcitro:
Yeah, "here try my new tool it's the best!" is always a mixed bag. 😀 

So I think there are two questions:
1. Do we think using `tox` simplifies the travis setup?
1. Do we want to keep a `tox.ini` file around for folks who want to use it, knowing that it may get out of date?

In particular, I could start by just walking back the `README.md` changes, or also the `.travis.yml` changes.

takluyver:
1. It looks like it would be simpler if we went all-in on tox, i.e. have one Travis build which invokes tox to do N different envs. As things stand, with N Travis builds each doing one tox env, I think it's more complex than omitting tox entirely.
2. I'm inclined to say that we should only keep it if we're intending to use it. But I don't feel strongly about that.

willingc:
I'm inclined to agree with Thomas' suggestion. One area (documentation builds and linkchecks) could be simplified by using tox locally (not Travis).

I could see not changing the Travis approach at this point. But I would be fine with keeping a scaled down tox.ini that focuses on docs locally (or even tests locally if not using conda).

craigcitro:
OK, I'm going to close this for now. It sounds like maybe we want to explore `tox` one day, but right now, it's more work than it's worth, given that the devs can't use it themselves (since, eg, it doesn't work with `conda`).

I've pulled all the useful bits out into #277.

Carreau:
Thanks, I opened #278 to track that and see if we can figure out the conda story.

> It looks like it would be simpler if we went all-in on tox, i.e. have one Travis build which invokes tox to do N different envs. As things stand, with N Travis builds each doing one tox env, I think it's more complex than omitting tox entirely.

That would be horrible (IMHO) as the build would be sequential so way slower, and you couldn't restart a single job. Plus it would make it extremely hard to see when you break only a specific python version.

Issue #276, opened 2017-07-05
Port already in use when kernel starts


rolweber:
Summary: Before starting a kernel, Jupyter picks some unused port numbers and writes them into the connection file. Then it releases those port numbers, so the starting kernel can use them. During that interval, some other process opening a socket connection can allocate one (or more) of the ports. The starting kernel finds a port blocked and runs into some error condition. In the best of cases, Jupyter detects that the kernel doesn't come up and restarts it. But the restart uses the same connection file and port numbers, so the problem persists.

This is a reincarnation of https://github.com/jupyter/kernel_gateway/issues/131, which was closed there because it is an upstream problem. @minrk suggested to restart the kernel with newly selected port numbers if this error condition - failure on initial start of the kernel - is detected.

Unless there are other suggestions to deal with the situation, I'd like to work on a PR that implements this behavior. I'll be on vacation next week, but I'm opening the issue already now, to gather as much feedback as possible.

----

Looking at the code, I think that the `ConnectionFileMixin` in [connect.py](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/connect.py) will need some attention. It has to remember which of the port numbers are configured and which are chosen at random, so that only the latter are regenerated.
Another place I need to look at is the `KernelManager` in [manager.py](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py). I thought about deleting the kernel and starting a different one, with possibly a different ID. But there are event handling methods in the class, so it's probably easier to keep the instance and generate a new connection file. Maybe I should introduce a `relaunch_kernel`, in addition to `start_kernel` and `restart_kernel`?
But I'm not sure where to detect and handle the actual error condition. Is this something that should be done in jupyter_client itself? Or rather leave it to notebook, kernel_gateway, and other applications using the jupyter_client? Your suggestions would be most welcome :-)

rolweber:
[consoleapp.py](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/consoleapp.py) starts kernels, maybe that's the place to detect and handle failure of initial kernel startup.
[jupyter/notebook](https://github.com/jupyter/notebook/search?utf8=%E2%9C%93&q=start_kernel&type=) and [jupyter/kernel_gateway](https://github.com/jupyter/kernel_gateway/search?utf8=%E2%9C%93&q=start_kernel&type=) call the `start_kernel` method in the kernel manager though. 

[...]

PR #279, opened 2017-07-06
Port already in use

rolweber:
See issue #276.

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -423,7 +423,8 @@ class ConnectionFileMixin(LoggingConfigurable):
     def _record_random_port_names(self):
         """Records which of the ports are randomly assigned.
 
-        Records on first invocation. Does nothing on later invocations."""
+        Records on first invocation, if the transport is tcp.
+        Does nothing on later invocations."""
 
         if self.transport != 'tcp':
             return
@@ -435,6 +436,20 @@ class ConnectionFileMixin(LoggingConfigurable):
             if getattr(self, name) <= 0:
                 self._random_port_names.append(name)
 
+    def cleanup_random_ports(self):
+        """Forgets randomly assigned port numbers and cleans up the connection file.
+
+        Does nothing if no port numbers have been randomly assigned.
+        In particular, does nothing unless the transport is tcp.
+        """
+
+        if not self._random_port_names:
+            return
+
+        for name in self._random_port_names:
+            setattr(self, name, 0)
+
+        self.cleanup_connection_file()
 
     def write_connection_file(self):
         """Write connection info to JSON dict in self.connection_file."""
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -333,6 +333,9 @@ class ConnectionFileMixin(LoggingConfigurable):
     control_port = Integer(0, config=True,
             help="set the control (ROUTER) port [default: random]")
 
+    # names of the ports with random assignment
+    _random_port_names = None
+
     @property
     def ports(self):
         return [ getattr(self, name) for name in port_names ]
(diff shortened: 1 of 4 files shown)


rolweber:
This PR is ready for a first round of reviews and feedback. Summary:
- `connect.py`: The ConnectionFileMixin tracks which ports are randomly assigned, and knows how to unassign them.
- `manager.py`: The KernelManager.restart_kernel function has a new, optional parameter that tells whether to choose new random ports or keep the old ones.
- `restarter.py`: The KernelRestarter will keep the random ports as soon as the kernel has been detected alive. If the kernel must be restarted without ever being alive, new ports will be chosen.

It depends on the kernel implementation whether that addresses the port-in-use problem. If a kernel first gets all ports and does not respond to heartbeats unless that succeeds, it will be restarted with new ports. But if there are kernel implementations out there which respond to a heartbeat and therefore appear alive, although they couldn't get one of the other ports, they will be restarted with the same ports. But there's only so much that can be detected on the Jupyter client side.

minrk:
Nice! I think this is a good idea. We will have to think a little bit carefully about when newports is invoked automatically to avoid breaking things in surprising ways.

In particular, we should check what's going to happen in the notebook when this event occurs. What I would like to avoid is this sequence of events:

1. new kernel is requested
2. websocket channel connects to assigned port
3. kernel fails and is restarted with a new port
4. kernel is running and reports okay, but notebook is connected to the wrong port and wasn't notified that the port changed

This would be frustrating and confusing to have everything claim to be working, but have execution silently fail.

It's okay if this is the starting point because it's replacing an already broken case of a dead kernel, but I want to make sure that the right notifications/events are in place to handle it better in the notebook (i.e. ensure that connected channels reconnect to the new ports).

rolweber:
@minrk: Thanks for having a look. I agree that we need to think through some error scenarios.

The websocket channel connects to the notebook server, not to the kernel, right?
All kernel sockets are for ZeroMQ communication, or am I missing something?

minrk:
@rolweber when a websocket connects to the server, a corresponding zmq socket is connected from the server to the kernel. I want to know what will happen if the connection happens before the restart (i.e. is it possible that the zmq socket is connected to the first port and not the new port).

rolweber:
I studied the code of class [`ZMQChannelsHandler`](https://github.com/jupyter/notebook/blob/5.0.0/notebook/services/kernels/handlers.py#L100) in [`notebook/services/kernels/handlers.py`](https://github.com/jupyter/notebook/blob/5.0.0/notebook/services/kernels/handlers.py), along with some base classes and mixins. Afaict, ZMQ connections to the kernel are created in [`create_stream`](https://github.com/jupyter/notebook/blob/5.0.0/notebook/services/kernels/handlers.py#L126), which calls one of the [`connect_*`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/connect.py#L508) methods in the `ConnectionFileMixin`. These methods use the current port number as stored in the mixin attributes.

`create_stream` registers a handler for kernel restart events, [`on_kernel_restarted`](https://github.com/jupyter/notebook/blob/5.0.0/notebook/services/kernels/handlers.py#L433), which sends a status update about the restart. I assume that the status message is going to the browser. From there on, it's the JavaScript code in the browser that has to handle the event.

Without the change I'm proposing here, the ZMQ connections between the notebook server and the kernel already have to be re-established, because they were dropped when the old kernel got killed. With my change, re-establishing connections will pick up the new port numbers. Disclaimer: AFAICT.

rolweber:
The `KernelRestarter` already has some config parameters. I could add another one that enables the new behavior, while keeping the old behavior as the default.

rolweber:
If we feel it is safe, we could also set the default to the new behavior and keep the config parameter only as a fallback in case of unexpected problems.

rolweber:
I'm back from vacation. Any suggestions on how I can move this PR further towards merging?

minrk:
Thanks for your detailed look into the notebook behavior! I think we can make this True by default based on your findings. Then I think it's good to go.

rolweber:
Thanks for reviewing! I switched the default.

PR #280, opened 2017-08-02
Clarify recommended use of metadata in docs.

craigcitro:
(This is a followup to #274.)

/cc @rgbkrk

### DIFF
--- a/docs/api/kernelspec.rst
+++ b/docs/api/kernelspec.rst
@@ -30,6 +30,12 @@ kernelspec - discovering kernels
       Additional kernel-specific metadata; clients can use this as needed,
       for instance to aid in kernel selection and filtering.
 
+      Metadata added here should be namespaced for the tool reading and
+      writing that metadata. Concretely, if you're adding a new field called
+      :code:`supported_versions` which your tool recognizes, then you should
+      add it as :code:`"mytool": {"supported_versions": [1, 2]}`, **not** as a
+      top-level field called :code:`supported_versions`.
+
    .. attribute:: resource_dir
 
       The path to the directory with this kernel's resources, such as icons.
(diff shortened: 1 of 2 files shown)


docs/api/kernelspec.rst
```diff
@@ -30,6 +30,10 @@ kernelspec - discovering kernels
       Additional kernel-specific metadata; clients can use this as needed,
       for instance to aid in kernel selection and filtering.
 
+      As with all other metadata fields, jupyter clients should use
+      app-specific submessages as opposed to storing their own top-level
```
takluyver: submessages -> subdictionaries or subobjects?

docs/api/kernelspec.rst
```diff
@@ -30,6 +30,10 @@ kernelspec - discovering kernels
       Additional kernel-specific metadata; clients can use this as needed,
       for instance to aid in kernel selection and filtering.
 
+      As with all other metadata fields, jupyter clients should use
```
takluyver: Do we specifically mean 'clients' here, or just third party tools in general?

rgbkrk: Clients as in any _jupyter protocol_ frontend, not necessarily a jupyter _branded_ client.

docs/api/kernelspec.rst
rgbkrk: What does this mean?

takluyver: The kernelspec is not a message, so I wouldn't call parts of its data 'submessages'.

docs/api/kernelspec.rst
craigcitro: Improved (I hope). LMK if this is clearer.

docs/api/kernelspec.rst
craigcitro: Add an example, in the hopes that would clarify things. (Happy to tweak more if it doesn't.)

craigcitro:
Sorry for the slow reply; I think the filters I set up for my github emails are working a little **too** well. 😊

minrk:
Thanks!

PR #284, opened 2017-08-08
unconditionally use setuptools

minrk:
follow pattern of other packages, requiring setuptools and disabling implicit bdist_egg

follow-up to #282 #283, resolving any inconsistencies for installing in different contexts. Prevents rare no-setuptools installs from ever being different from normal installs.


Carreau:
+1 did you ment to remove the jupyter-kernelspec and jupyter-run scripts ?

PR #201, opened 2016-09-13
set sticky bit on connection files

minrk:
avoids periodic cleanup of runtime directory

closes #199

PR #286, opened 2017-08-09
Soften and suppress warnings about sticky bits

minrk:
- specifically silence EPERM on the containing directory,
  which can happen when using tempfiles.
- when warning is triggered, explain that it's not a big deal and what might happen.

and actually set the sticky bit, which I just realized we weren't doing because I used the wrong variable to update the permissions.

cf discussion in #201

cc @takluyver @AlJohri


AlJohri:
thanks y'all! just wanted to check if this is already released - ran into the issue on 5.3.1

takluyver:
It should be - you can see the milestone is set to 5.2.

Issue #186, opened 2016-08-11
Public API for Customizing Kernel Restart Behavior


ssanderson:
On Quantopian, we regularly (multiple times per day at our current usage) run into the following pathological case with the default kernel restart behavior:
1. A user (running in a docker container launched by Dockerspawner) has many notebooks open, each of which consumes a significant amount of memory.
2. The user opens a new notebook, and the creation of their kernel hits a memory limit configured in docker, causing it to get OOM-killed.  We load a nontrivial amount of data in some of our extension setup, so a user with tens of megabytes left under the cap can OOM before they get a chance to run another cell.  Depending on how happy AWS is on any given day, and on how nice docker is feeling, it can take 10+ seconds to for the kernel to die.
3. The KernelRestarter notices that the kernel process has died and restarts it.  The restart attempt counter is set to 1.
4. Kernel survives long enough that the next poll succeeds, and we reset the restart attempt counter to zero.
5. Kernel gets OOM-killed again.
6. Repeat steps 3-5 until someone manually intervenes.

The simplest solution for us to to increase the poll interval on KernelRestarter to be long enough that a nearly-OOM kernel reliably dies before we poll again.  This is tricky to tune, however, and it has negative UX impact if the kernel crashes for other reasons.  

What I'd really like to be able to do is to customize the kernel restart logic to include heuristics for detecting and giving up early in the nearly-OOM case.  As far as I can tell, however, there's no way to provide a custom subclass of `KernelRestarter` in a config file, because the initialization logic for `IOLoopKernelManager` looks like:

```
    _restarter = Instance('jupyter_client.ioloop.IOLoopKernelRestarter', allow_none=True)

    def start_restarter(self):
        if self.autorestart and self.has_kernel:
            if self._restarter is None:
                self._restarter = IOLoopKernelRestarter(
                    kernel_manager=self, loop=self.loop,
                    parent=self, log=self.log
                )
            self._restarter.start()

    def stop_restarter(self):
        if self.autorestart:
            if self._restarter is not None:
                self._restarter.stop()
```

So I think with the way things stand right now, my best option is to override `MultiKernelManager.kernel_manager_class`, which is explicitly configurable.  Two questions about this:
1. Is it reasonable to override the whole kernel manager to customize the restart logic?
2. Is there interest in a public API for just configuring the restarting behavior?

Carreau:
> Is it reasonable to override the whole kernel manager to customize the restart logic?

[...]

PR #290, opened 2017-08-26
ENH: Make KernelRestarter class configurable.

ssanderson:
Make the `KernelRestarter` class used by `IOLoopKernelManager` configurable.

### DIFF
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -17,7 +17,8 @@ from zmq.eventloop import ioloop
 from zmq.eventloop.zmqstream import ZMQStream
 
 from traitlets import (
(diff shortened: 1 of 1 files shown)


ssanderson:
xref: https://github.com/jupyter/jupyter_client/issues/186

ssanderson:
@minrk do you think there would be interest in taking the throttling-based restart implementation posted at https://github.com/jupyter/jupyter_client/issues/186#issuecomment-241904699, either as a replacement for or an alternative to the current N-consecutive-attempts implementation?

minrk:
@ssanderson as discussed in person: yes, I think the restarted-budget logic you mentioned makes sense to bring in as the default behavior.

PR #304, opened 2017-11-10
prepare for tornado 5

minrk:
- use IOLoop.current over IOLoop.instance
- drop removed `loop` arg from PeriodicCallback
- deprecate now-unused IOLoopKernelRestarter.loop

pyzmq 17 will be needed to be fully compatible with tornado 5, which runs on asyncio on Python 3.


takluyver:
Test failure in `test_tracking`.

takluyver:
@meeseeksdev backport

PR #306, opened 2017-11-22
extend special handling of sys.executable to pythonX[.Y]

minrk:
this should allow ipykernel's wheel-installed specs to specify `python3` or `python2` and prevent python2 kernels from launching with sys.executable if the Python version is 3.

The new KernelFinder machinery may make this obsolete, but I think we still need this for ipykernel's wheels to avoid launching with sys.executable.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -11,11 +11,6 @@ import re
 import signal
 import sys
 import time
-import warnings
-try:
-    from queue import Empty  # Py 3
-except ImportError:
-    from Queue import Empty  # Py 2
 
 import zmq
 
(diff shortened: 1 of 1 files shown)


takluyver:
Yup, I think this makes sense.

Do you want anyone specific to consider this before we merge? I'm happy to press the button.

minrk:
I think this is pretty straightforward, so I'm happy as-is. The release process is the trickier part - what version of jupyter-client this goes in (since you are working on big 6.0 things right now), when to adopt it in ipykernel, etc.

takluyver:
I think the big changes are still likely to take a while to settle, so it's worth doing a 5.2 release with this and other improvements. I might make a 5.x branch and back out the new discovery machinery, since that's part of the big changes I'm still working on.

takluyver:
@meeseeksdev backport

PR #314, opened 2017-12-15
kill process group when killing kernel

minrk:
if killpg is available. Keeps kill_kernel consistent with other signals.

This should cleanup process groups (e.g. multiprocessing subprocesses) and make EADDRINUSE less likely during restart (ipython/ipykernel#274)
when kernels don't cleanup after themselves (ipykernel#290).


takluyver:
@meeseeksdev backport

PR #325, opened 2017-12-30
Parenthesize conditional requirement in setup.py

### LINKED ISSUE #324: extra_require not respected with .whl (bug in wheel package?)
See discussion in https://github.com/pypa/setuptools/issues/1242
Pr on it's way. We should likely release a 5.2.1

Carreau:
Du to a likely bug in wheel, the conditional dependency on pytest ends
up being unconditional. Seem like adding parenthesis fix that (as a work
around).

See https://github.com/pypa/setuptools/issues/1242
Closes #324

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,13 @@
 Changes in Jupyter Client
 =========================
 
+5.2.1
+=====
+
+- Add parenthesis to conditional pytest requirement to work around a bug in the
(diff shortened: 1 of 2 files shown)


Carreau:
I've create a 5.2.x branch (from 5.2.0, please verify), created a 5.2.1 milestone and configured it (see milestone description) to ask MrMeeseeks to backport to 5.2.x on merge.

PR #328, opened 2018-01-03
Exclude build docs from sdist.

Carreau:
This shrinks the sdist from 2MB to ~250KB... just realized that after
uploading 5.2.1 took way too long. Apparently 5.2.0 alsho shipped built
docs.


blink1073:
Thanks!

PR #335, opened 2018-01-10
Migrate SSH tunneling from pyzmq

dmanikowski-reef:
As discussed in https://github.com/ipython/ipython/issues/10934

As this part of code is mainly used in IPython we
agreed to move it from pyzmq. It will be easier to
maintain here and some planned changes in this code
will be easier to apply and release.

Also some code reformatting was applied in `connect.py` file.

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -35,9 +35,9 @@ from jupyter_core.paths import jupyter_data_dir, jupyter_runtime_dir
 
 
 def write_connection_file(fname=None, shell_port=0, iopub_port=0, stdin_port=0, hb_port=0,
-                         control_port=0, ip='', key=b'', transport='tcp',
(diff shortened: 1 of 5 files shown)


jupyter_client/connect.py
```diff
@@ -249,7 +249,7 @@ def tunnel_to_kernel(connection_info, sshserver, sshkey=None):
     (shell, iopub, stdin, hb) : ints
         The four ports on localhost that have been forwarded to the kernel.
     """
-    from zmq.ssh import tunnel
+    from jupyter_core.ssh import tunnel
```
minrk: This should now be `from .ssh`

dmanikowski-reef: Done.

minrk:
Thanks!

PR #336, opened 2018-01-16
handle classes having been torn down in atexit

### LINKED ISSUE #323: HBChannel._exiting = True, AttributeError: 'NoneType' object has no attribute '_exiting'
I get the following problem after running `python setup.py test`. All the tests succeed, but this error message appears in the end.

```
Error in sys.exitfunc:
  File "/root/miniconda/envs/test_env/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/root/miniconda/envs/test_env/lib/python2.7/site-packages/jupyter_client/channels.py", line 83, in _notice_exit
    HBChannel._exiting = True
AttributeError: 'NoneType' object has no attribute '_exiting'
```

minrk:
we could probably avoid this if we registered/unregistered atexit callbacks for instances instead of registering it once for classes at import time

closes #323

Perhaps more hygenic, if more complex alternative that would use instances:

- register atexit callback in `__init__`
- use a weakref to avoid keeping a reference forever
- unregister callback in `__del__`

Attaching it to an instance should prevent any process-teardown issues, since classes don't get deleted while they still have running instances (I'm pretty sure).

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -80,7 +80,10 @@ class HBChannel(Thread):
     @staticmethod
     @atexit.register
     def _notice_exit():
-        HBChannel._exiting = True
+        # Class definitions can be torn down during interpreter shutdown.
+        # We only need to set _exiting flag if this hasn't happened.
+        if HBChannel is not None:
+            HBChannel._exiting = True
 
     def _create_socket(self):
         if self.socket is not None:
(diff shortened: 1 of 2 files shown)


takluyver:
I'm happy enough with this simple but somewhat dirty approach.

PR #339, opened 2018-01-23
avoid calling private method in subclasses of KernelSpecManager

### LINKED ISSUE #338: get_all_specs not compatible with subclassing
[nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels) subclasses KernelSpecManager to find conda kernels. It does so by overriding `find_kernel_specs` to return its name list and resource directories, and `get_kernel_spec` to return the specs themselves.

We've added `.get_all_specs` as a simpler API to get all of the specs without separately finding and loading them. But we've done this in such a way that `.get_all_specs` calls `.find_kernel_specs`, a public API, and `_get_kernel_spec_by_name`, a private method. This means that implementations that override `.find_kernel_specs` in such a way that `_get_kernel_spec_by_name` won't work, then `.get_all_specs` is broken for any such subclass (this includes nb_conda_kernels 2.1).

Now, one option is for custom kernelspec managers to implement all three methods - find_kernel_specs, get_all_specs, and get_kernel_spec, but this is a backward-incompatible change, and I think we should consider it a bug.

I think we should ensure that `.get_all_specs` works for KSM subclasses that have only implemented the public methods `.find_kernel_specs` and `.get_kernel_spec` to fix this backward-incompatibility, and push that out in a bugfix release.

cc @dartdog who reported the issue

minrk:
on the result of a public overrideable method, which breaks subclasses that don't override get_all_specs

rather than falling back on the public method, which is slower because it looks up the location of a directory we have already found, reserve this optimization for the base class itself, which is the only place this is a safe method to call.

closes #338

I think we should ship a 5.2.2 bugfix upon merging this one and confirming it fixes nb_conda_kernels

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -254,7 +254,14 @@ class KernelSpecManager(LoggingConfigurable):
         res = {}
         for kname, resource_dir in d.items():
             try:
-                spec = self._get_kernel_spec_by_name(kname, resource_dir)
+                if self.__class__ is KernelSpecManager:
+                    spec = self._get_kernel_spec_by_name(kname, resource_dir)
+                else:
+                    # avoid calling private methods in subclasses,
+                    # which may have overridden find_kernel_specs
+                    # and get_kernel_spec, but not the newer get_all_specs
+                    spec = self.get_kernel_spec(kname)
+
                 res[kname] = {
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks for tracking this down!

PR #341, opened 2018-01-24
changelog for 5.2.2

minrk:
will release 5.2.2 from 5.x after this. I've merged 5.2.x into 5.x, since it only contained the one 5.2.1 change, and 5.x only has changes for 5.2.2.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,12 +4,28 @@
 Changes in Jupyter Client
 =========================
 
+5.2.2
+=====
+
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! I'll start releasing 5.2.2 now.

minrk:
Released to PyPI, git

takluyver:
Thanks Min! I see you've already dealt with conda-forge as well (https://github.com/conda-forge/jupyter_client-feedstock/pull/16).

PR #343, opened 2018-01-27
import warnings for 5.2.x

rgbkrk:
This might need to be targeted on 5.x. We were running into an issue where a warning was being issued without `import warnings` being done at the top.


rgbkrk:
Switched the base branch to `5.x`.

takluyver:
Thanks, makes sense.

PR #344, opened 2018-01-27
return None kernel_spec if kernel_name not set

rgbkrk:
This restores the ability of using the deprecated `kernel_cmd`, which previously allowed you to run an arbitrary kernel without having to write a kernelspec. Without this, when using `kernel_cmd`, a lookup on an empty `kernel_name` (with `""` as value) was being performed.


takluyver:
Thanks!

PR #345, opened 2018-01-29
Add comm-open and comm-msg methods to client

eddiejessup:
Let users of the Jupyter client send messages to open new comms, and send messages to comms. These methods implement the specification as described [here](https://jupyter-client.readthedocs.io/en/stable/messaging.html#custom-messages).

The tests don't look for a response, because one isn't expected for either message type.

Since the methods to send messages on the shell channel seemed to have duplicated logic, I factored it out into a helper method `_send_shell_message`. None of the methods that now use it should change their behavior as a result.

### DIFF
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -160,6 +160,8 @@ class BlockingKernelClient(KernelClient):
     inspect = reqrep(KernelClient.inspect)
     kernel_info = reqrep(KernelClient.kernel_info)
     comm_info = reqrep(KernelClient.comm_info)
+    comm_open = reqrep(KernelClient.comm_open)
+    comm_message = reqrep(KernelClient.comm_message)
     shutdown = reqrep(KernelClient.shutdown)
 
 
--- a/jupyter_client/blocking/client.py
+++ b/jupyter_client/blocking/client.py
@@ -162,6 +162,7 @@ class BlockingKernelClient(KernelClient):
     comm_info = reqrep(KernelClient.comm_info)
     comm_open = reqrep(KernelClient.comm_open)
     comm_message = reqrep(KernelClient.comm_message)
+    comm_close = reqrep(KernelClient.comm_close)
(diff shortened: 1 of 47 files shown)


vidartf:
I'm not in a position to say whether this should go in or not, but I will point out that it should probably also include the `comm-close` message, and support buffers for the messages (not sure if it currently does or not).

eddiejessup:
Hey @vidartf, thanks for taking a look!

About `comm-close`, I actually didn't even see this was part of the protocol, though now it seems obvious. I'll add that.

About buffers, sorry I'm not sure what it means to support buffers for the messages. I've had a grep through the code but can't find anything very relevant. Any chance you could clarify what this would involve?

jupyter_client/blocking/client.py
```diff
@@ -160,6 +160,9 @@ def _recv_reply(self, msg_id, timeout=None):
     inspect = reqrep(KernelClient.inspect)
     kernel_info = reqrep(KernelClient.kernel_info)
     comm_info = reqrep(KernelClient.comm_info)
+    comm_open = reqrep(KernelClient.comm_open)
+    comm_message = reqrep(KernelClient.comm_message)
+    comm_close = reqrep(KernelClient.comm_close)
```
takluyver: These probably shouldn't be wrapped in `reqrep`, because they don't (in general) have a reply to wait for.

takluyver:
Re buffers: they're mentioned and the end of [this section](https://jupyter-client.readthedocs.io/en/stable/messaging.html#the-wire-protocol). But I think it would require changes in a few levels to pass them through, so feel free to leave it to a separate PR.

eddiejessup:
OK I get what you mean now. Yeah, to be honest I don't feel comfortable enough with the serialization-level stuff to expect to implement this well.

If you guys are happy to merge this in as-is, then maybe later on I or someone else can add support for buffers, I'm happy with that.

PR #352, opened 2018-03-08
tornado 5 fixes in ThreadedClient

minrk:
The main issue was failing to schedule `ioloop.stop` in the ioloop thread, which is required when tornado is running on asyncio in order to wake the thread. The result of not doing this is hanging forever when trying to exit, e.g. in QtConsole.

There is further cleanup of threadsafety issues with respect to asyncio and tornado objects.

closes https://github.com/jupyter/qtconsole/issues/275

cc @ccordoba12

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -182,7 +182,7 @@ class IOLoopThread(Thread):
         :meth:`~threading.Thread.start` is called again.
         """
         if self.ioloop is not None:
-            self.ioloop.stop()
+            self.ioloop.add_callback(self.ioloop.stop)
         self.join()
         self.close()
 
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -3,7 +3,8 @@
 from __future__ import absolute_import
 import atexit
 import errno
-from threading import Thread
+import sys
+from threading import Thread, Event
 import time
 
 # import ZMQError in top-level namespace, to avoid ugly attribute-error messages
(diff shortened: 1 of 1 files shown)


jupyter_client/threaded.py
```diff
@@ -156,8 +163,26 @@ def _notice_exit():
...
+        Thread.start(self)
+        self._start_event.wait()
+
     def run(self):
         """Run my loop, ignoring EINTR events in the poller"""
+        if 'asyncio' in sys.modules:
+            # tornado may be using asyncio,
+            # ensure an eventloop exists for this thread
+            import asyncio
+            asyncio.set_event_loop(asyncio.new_event_loop())
```
takluyver: It seems odd that we have to do this. Is there a neater way that we can use in the future when we can assume that tornado is using asyncio?

ccordoba12:
I tested this against qtconsole and Spyder and it works as expected. Thanks @minrk!

jupyter_client/threaded.py
minrk: The situation: asyncio will not create an eventloop in a thread, you have to tell it to, at least in certain circumstances. Tornado is running on top of asyncio, but won't create the required asyncio eventloop if there isn't one (asyncio.get_event_loop will fail in threads that haven't initialized eventloops explicitly). I'm not sure if this is a tornado bug or not.

We can ask if the tornado configured IOLoop class is a subclass of AsyncIOLoop, which is the tornado asyncio wrapper implementation. If we assume that's going to be the case and not some other weird requires-asyncio implementation that doesn't subclass it.

Note that this is just instantiating an object, not starting it or anything, so it's a pretty minimal operation and harmless if the thread-local eventloop goes unused. I'm not quite sure what the cleanest solution is for "create the asyncio eventloop that tornado may need in this thread only if tornado is actually going to need it", because we would need more detailed try-except for importing tornado.platform.asyncio.AsyncIOLoop which may not be defined (e.g. python 2 or older tornado) in order to call `issubclass()`.

vidartf: You could consider setting [our own policy](https://docs.python.org/3/library/asyncio-eventloops.html#asyncio.AbstractEventLoopPolicy.get_event_loop), so that we handle the calls to `get_event_loop`.

stonebig:
will there be a patch release soon so everybody can enjoy Tornado5 ?

jupyter_client/threaded.py
minrk: We shouldn't set our own policy in jupyter_client because that would preclude applications setting their own policies. We could set our own in qtconsoleapp, but can't rely on that in the qtconsole kernelmanager, which can be used in other applications like Spyder.

minrk:
@stonebig yes, hopefully tomorrow. I'll open a PR with a changelog for 5.2.3 and it should be out shortly.

PR #353, opened 2018-03-09
Wrap setting of kernel_id with method that can then be overridden in …

kevin-bates:
…subclasses.

A recent requirement for [Jupyter Enterprise Gateway](https://github.com/jupyter-incubator/enterprise_gateway) is for clients to be able to
specify the kernel_id for new kernels.  Although `jupyter_client.start_kernel()` will
honor client-provided kernel_ids, Notebook's override of `start_kernel()` changes
the semantics of a non-null kernel_id in the argument list to mean an existing
(persisted) kernel should be _started_.  As a result, applications that derive
from the kernel management infrastructure beyond Notebook cannot influence the
derivation of the kernel's id via the existing argument list behavior.

By introducing the `determine_kernel_id()` method, subclasses are able to derive
the kernel's id however they wish.

With the ability to know the kernel's id prior to its invocation, a number of
things can be done that wouldn't be possible otherwise.  For example, this
provides the ability to setup a shared filesystem location possibly pre-populated
with data relative to what the request (i.e., kernel) is going to need.

### DIFF
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -90,7 +90,7 @@ class MultiKernelManager(LoggingConfigurable):
 
         The kernel ID for the newly started kernel is returned.
         """
-        kernel_id = kwargs.pop('kernel_id', unicode_type(uuid.uuid4()))
+        kernel_id = self.determine_kernel_id(**kwargs)
         if kernel_id in self:
             raise DuplicateKernelError('Kernel already exists: %s' % kernel_id)
 
@@ -315,3 +315,13 @@ class MultiKernelManager(LoggingConfigurable):
         =======
         stream : zmq Socket or ZMQStream
         """
+
+    def determine_kernel_id(self, **kwargs):
+        """
+        Returns the kernel_id to use for this request.  If kernel_id is already in the arguments list,
+        that value will be used.  Otherwise, a newly generated uuid is used.  Subclasses may override
+        this method to substitute other sources of kernel ids.
+        :param kwargs:
+        :return: string-ized version 4 uuid
+        """
+        return kwargs.pop('kernel_id', unicode_type(uuid.uuid4()))
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -86,11 +86,11 @@ class MultiKernelManager(LoggingConfigurable):
         """Start a new kernel.
 
         The caller can pick a kernel_id by passing one in as a keyword arg,
-        otherwise one will be picked using a uuid.
+        otherwise one will be generated using new_kernel_id().
 
         The kernel ID for the newly started kernel is returned.
         """
-        kernel_id = self.determine_kernel_id(**kwargs)
+        kernel_id = kwargs.pop('kernel_id', self.new_kernel_id(**kwargs))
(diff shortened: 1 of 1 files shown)


jupyter_client/multikernelmanager.py
```diff
@@ -315,3 +315,13 @@ def connect_hb(self, kernel_id, identity=None):
         =======
         stream : zmq Socket or ZMQStream
         """
+
+    def determine_kernel_id(self, **kwargs):
```
minrk: Maybe simpler verb 'new_kernel_id'?

jupyter_client/multikernelmanager.py
```diff
@@ -90,7 +90,7 @@ def start_kernel(self, kernel_name=None, **kwargs):
 
         The kernel ID for the newly started kernel is returned.
         """
-        kernel_id = kwargs.pop('kernel_id', unicode_type(uuid.uuid4()))
+        kernel_id = self.determine_kernel_id(**kwargs)
```
minrk: Should `determine_kernel_id` be called if kernel_id is specified? It seems like it should only be called if kernel_id is not specified.

jupyter_client/multikernelmanager.py
kevin-bates: I'm okay with changing the verb in the method name - please see next comment.

jupyter_client/multikernelmanager.py
kevin-bates: @minrk - thank you for taking a look at this.

I was originally thinking that the override should be called unconditionally with the base implementation honoring the existence of the parameter, but that probably leaves too much room for mistakes, so I'd be okay with calling the method only when `kernel_id is None`.

Not sure if its my Java background (very new to python), but I don't like using `new` as a verb.  Does something like `init_kernel_id(**kwargs)` sound any better?   

Would you mind suggesting the appropriate *python way* for constructing the conditional call based on the `pop` result?

minrk: I would write it like this:

```python
if 'kernel_id' in kwargs:
    kernel_id = kwargs.pop('kernel_id')
else:
    kernel_id = self.new_kernel_id()
```

By Python (or maybe just Jupyter) convention, I like `new` better than `init`, because `init` implies a stateful change. I would expect `.init_kernel_id()` to *set* `self.kernel_id`, not return a value I'm free to use. The verb I would use for a stateless callable that returns a new kernel id is 'new'. 'generate' works, too, but it's more verbose than I would like.

kevin-bates: Thanks for the explanation on `new` vs. `init` - that makes sense.  Regarding the example (and a better understanding of python args), I figured things would be more compact and  something like the original would have been suggested:
```python
kernel_id = kwargs.pop('kernel_id', self.new_kernel_id(**kwargs))
```
I'd prefer to push this version, but would be happy to submit the expanded form if you feel strongly about it.

minrk: It doesn't make much of a difference here, but I find clarity is typically preferable to compactness. The `if/else` is also ever-so-slightly more efficient, since `new_kernel_id` is only called if needed, rather than this PR, where it's called unconditionally, even if the result is just discarded.

PR #357, opened 2018-03-11
Add setup() args for new pypi.org

willingc:
- Adds additonal python 3 versions
- Adds Jupyter framework to classifiers :-)
- Adds project_urls to provide links to end users to docs, source, issue tracker

https://packaging.python.org/tutorials/distributing-packages/#setup-args


setup.py
```diff
@@ -60,22 +60,35 @@ def run(self):
...
+        'Operating System :: MacOS',
+        'Operating System :: Microsoft :: Windows',
+        'Operating System :: POSIX :: Linux',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
```
takluyver: I feel like I harp on about this, but my preference is to use the `2` and `3` classifiers and leave the more specific ones out. They only get out of date, as these have.

The `python_requires` parameter, which I see you've also added, is a better way of conveying which minor versions it's meant to work with.

setup.py
```diff
@@ -60,22 +60,35 @@ def run(self):
...
     classifiers     = [
+        'Framework :: Jupyter',
         'Intended Audience :: Developers',
+        'Intended Audience :: Education',
         'Intended Audience :: System Administrators',
         'Intended Audience :: Science/Research',
         'License :: OSI Approved :: BSD License',
+        'Operating System :: MacOS',
+        'Operating System :: Microsoft :: Windows',
+        'Operating System :: POSIX :: Linux',
```
takluyver: I'd expect it to work on any platform where Python and ZMQ are available, which is probably a much bigger group than these three. Maybe use `Operating System :: OS Independent` instead? Or leave it out altogether - I think OS-independent is the default assumption for Python modules.

setup.py
```diff
@@ -85,6 +98,7 @@ def run(self):
         'entrypoints',
         'tornado>=4.1',
     ],
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
```
minrk: I think we can be optimistic and remove '<4' here. Putting this here prevents people from even attempting to install the package with a hypothetical Python 4 without us making a new release.

minrk:
@willingc great! I can release 5.2.3 with this after the above comments are addressed.

setup.py
willingc: Sure.

setup.py
```diff
@@ -60,22 +60,29 @@ def run(self):
...
+        'Tracker': 'https://github.com/jupyter/jupyter_client/issues',
+    },
     classifiers     = [
+        'Framework :: Jupyter',
         'Intended Audience :: Developers',
+        'Intended Audience :: Education',
         'Intended Audience :: System Administrators',
         'Intended Audience :: Science/Research',
         'License :: OSI Approved :: BSD License',
+        'Operating System :: Independent',
```
takluyver: The spelling is `Operating System :: OS Independent` . Tautological, I know, but I think PyPI refuses uploads if the classifiers don't match [this list](https://pypi.org/pypi?%3Aaction=list_classifiers).

willingc: Good catch. Fixing.

willingc:
Hang tight. I need to fix the version error for Python 2.

setup.py
```diff
@@ -85,6 +92,7 @@ def run(self):
         'entrypoints',
         'tornado>=4.1',
     ],
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <=4',
```
minrk: We can remove `<=4`. The rest is already correctly limiting to 2.7, 3.3+. This should be:

```python
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*',
```

adding `>=3.3` would exclude 2.7, since all of these conditions are AND-ed together.

willingc: Good point. I need more coffee.

minrk:
Thanks!

PR #361, opened 2018-03-14
Fix leak of IOLoopKernelManager object

kevin-bates:
After analyzing various leaked items when running either Notebook or
Jupyter Kernel Gateway, one item that recurred across each kernel
startup and shutdown sequence was an instance of IOLoopKernelManager.
(Of course, when using JKG, this instance was KernelGatewayIOLoopKernelManager
since it derives from the former.)

The leak is caused by the circular references established in the `self._restarter`
and `self.session.parent` members. This change breaks the circular reference when
the restarter is stopped and during `cleanup()` of the kernel manager.  Once the
references are broken, the kernel manager instance can be garbage collected.

### DIFF
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -54,6 +54,7 @@ class IOLoopKernelManager(KernelManager):
         if self.autorestart:
             if self._restarter is not None:
                 self._restarter.stop()
+                self._restarter = None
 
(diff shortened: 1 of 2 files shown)


kevin-bates:
@minrk - thank you for merging these PRs.  Can we assume that 5.2.4 will be released in a couple weeks given the other 5.2.x releases have been about 3-4 weeks apart?

PR #366, opened 2018-03-23
Undeprecate KernelManager.extra_env

### LINKED ISSUE #365: Please fix breakage caused by removal of `extra_env`
It used to be possible to do something like the following:

```
self.repl_kernel_manager = QtKernelManager(extra_env=self.envars)
self.repl_kernel_manager.start_kernel()
```

...and the dict `self.envars` passed in as `extra_env` would add environment variables to `os.environ` before the kernel started.

(See: https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L244)

This appears to have been removed with no mention in the changes. :-(

Can this *very useful* feature be re-added please..? I realise I could use `kernel.json` to specify this sort of thing, but in my case the application allows users to update environment variables on the fly and I want this reflected when they run their code with Jupyter.

Does this make sense..?

ntoll:
This fixes #365 (i.e. an undocumented breakage to the KernelManager API).

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -18,7 +18,7 @@ import zmq
 from ipython_genutils.importstring import import_item
 from .localinterfaces import is_local_ip, local_ips
 from traitlets import (
-    Any, Float, Instance, Unicode, List, Bool, Type, DottedObjectName
+    Any, Float, Instance, Unicode, List, Bool, Type, DottedObjectName, Dict
 )
 from jupyter_client import (
     launch_kernel,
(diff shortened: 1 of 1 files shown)


minrk:
I think something is confused. What release of jupyter-client has KernelManager.extra_env that's been removed? I only see it *added* in yet-to-be-released 6.0, not removed in any release.

ntoll:
Hi @minrk the confusion is on my part. I was looking at the version numbers and the one in `master` (where the `extra_env` work is) is less than the one that is released. ;-)

PR #379, opened 2018-05-15
make KernelManager configurable for all who inherit JupyterConsoleApp

mpacer:


mpacer:
@rgbkrk @takluyver @Carreau @minrk Do any of you have any comments on this?

rgbkrk:
My only real question is if we have to bear in mind what the NotebookApp or others already use for this field. If it overlaps, does it maintain the same purpose? (I do realize the notebook app has the two related classes, one for the single base and one for the mapping kernel manager).

mpacer:
The `NotebookApp` doesn't use this field — it is using `NotebookApp.kernel_manager_class` which does not inherit from `JupyterConsole` and so would have no way to access `JupyterConsole.kernel_manager_class`. 

As far as I can tell there are two main uses of the `JupyterConsole` application: 
- `juptyer_console` (which does not overwrite this attribute and so will inherit it appropriately)
- `qtconsole` (which overwrites this attribute [to be `QtKernelManager`](https://github.com/jupyter/qtconsole/blob/742f04d31c1869c4909b07c338cd4701a5e0080e/qtconsole/qtconsoleapp.py#L147
) thus will not be able to inherit this value)

Are there other known uses of the `JupyterConsole` application?

PR #380, opened 2018-05-16
update python version testing

Carreau:
3.3is EOL since last year, and pip print warnings in the test-suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
     - "nightly"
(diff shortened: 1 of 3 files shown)


Carreau:
Note that if we drop 3.3 in testing we might want to explicitly remove 3.3 compatibility in requires_python and maybe remove some conditionals.

cc @mpacer

mpacer:
rebased on master to resolve merge conflict

PR #384, opened 2018-06-06
Don't include extra buffers in message signature

### LINKED ISSUE #383: Documentation and implementation disagree on HMAC while using raw data buffers.
It seems that the [wire protocol documentation](http://jupyter-client.readthedocs.io/en/stable/messaging.html#the-wire-protocol) and the implementation of [`jupyter_client`'s message signing](https://github.com/jupyter/jupyter_client/blob/01fc51bb0d42955eace49c002e92b0f3c87eb130/jupyter_client/session.py#L759) disagree on how the HMAC should be computed in the presence of extra data buffers 
 (the two definitions are equivalent if there are no extra data buffers).

The documentation says:

> The signature is the HMAC hex digest of the concatenation of:
> * A shared key (typically the key field of a connection file)
> * The serialized header dict
> * The serialized parent header dict
> * The serialized metadata dict
> * The serialized content dict
> In Python, this is implemented via:
> 
> ```python
> # once:
> digester = HMAC(key, digestmod=hashlib.sha256)
> 
> # for each message
> d = digester.copy()
> for serialized_dict in (header, parent, metadata, content):
>     d.update(serialized_dict)
> signature = d.hexdigest()
> ```

whereas the implentation is
```python
def send_raw(self, stream, msg_list, flags=0, copy=True, ident=None):
    """
    msg_list : list
        The serialized list of messages to send. This only includes the
        [p_header,p_parent,p_metadata,p_content,buffer1,buffer2,...] portion of
        the message.
    """
    to_send = []
    # clip
    to_send.append(self.sign(msg_list))
    # clip
```

which calculates the signature including the buffers as well.

Should the documentation be updated?

travigd:
Closes #383 and makes `Session.send_raw` compliant with the protocol wire specification (w.r.t. signing optional binary data buffers at the end of messages).

Not sure if this would warrant test cases on the way signatures are handled in the presence of data buffers.

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -779,7 +779,8 @@ class Session(Configurable):
             to_send.extend(ident)
 
         to_send.append(DELIM)
-        to_send.append(self.sign(msg_list))
+        # Don't include buffers in signature (per spec).
+        to_send.append(self.sign(msg_list[0:4]))
         to_send.extend(msg_list)
         stream.send_multipart(to_send, flags, copy=copy)
 


minrk:
Thanks!

PR #388, opened 2018-07-13
Updates to the kernel messaging spec

jasongrout:
These came out of conversations with @minrk.

CC also @SylvainCorlay, who I think has been advocating for the deprecation of sending a shutdown message on the shell channel.

Marking as work-in-progress since I plan to work on this more in the near future. But I also think these changes are pretty self-contained, so I wouldn't regret us merging these now and submitting another PR later either.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1228,7 +1228,8 @@ Message type: ``status``::
     content = {
         # When the kernel starts to handle a message, it will enter the 'busy'
         # state and when it finishes, it will enter the 'idle' state.
-        execution_state : ('busy', 'idle', other optional states)
+        # The kernel will publish state 'starting' exactly once at process startup.
+        execution_state : ('busy', 'idle', 'starting')
     }
 
 When a kernel receives a request and begins processing it,
@@ -1238,10 +1239,6 @@ and has finished publishing associated IOPub messages, if any,
 it shall publish a status message with ``execution_state: 'idle'``.
 Thus, the outputs associated with a given execution shall generally arrive
 between the busy and idle status messages associated with a given request.
-
-A kernel may send optional status messages with execution states other than
-`busy` or `idle`. For example, a kernel may send a status message with a
-`starting` execution state exactly once at process startup.
 
 .. note::
 
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1205,6 +1205,8 @@ Message type: ``error``::
 
     ``pyerr`` renamed to ``error``
 
+.. _status:
+
 Kernel status
 -------------
 
@@ -1215,8 +1217,7 @@ Message type: ``status``::
     content = {
         # When the kernel starts to handle a message, it will enter the 'busy'
         # state and when it finishes, it will enter the 'idle' state.
-        # The kernel will publish state 'starting' exactly once at process startup.
-        execution_state : ('busy', 'idle', 'starting')
+        execution_state : ('busy', 'idle', other optional states)
     }
 
 When a kernel receives a request and begins processing it,
@@ -1226,6 +1227,10 @@ and has finished publishing associated IOPub messages, if any,
 it shall publish a status message with ``execution_state: 'idle'``.
 Thus, the outputs associated with a given execution shall generally arrive
 between the busy and idle status messages associated with a given request.
+
+A kernel may send optional status messages with execution states other than
+`busy` or `idle`. For example, a kernel may send a status message with a
+`starting` execution state exactly once at process startup.
 
 .. note::
 
@@ -1242,14 +1247,6 @@ between the busy and idle status messages associated with a given request.
 
     Busy and idle messages should be sent before/after handling every request,
     not just execution.
-
-.. note::
-
-    Extra status messages are added between the notebook webserver and websocket clients
-    that are not sent by the kernel. These are:
-
-    - restarting (kernel has died, but will be automatically restarted)
-    - dead (kernel has died, restarting has failed)
 
 Clear output
 ------------
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -124,15 +124,26 @@ A message is defined by the following four-dictionary structure::
 
 .. note::
 
-A client session value, in message headers from a client, should be unique among
-all clients connected to a kernel and should be constant over the lifetime of
-the client. A kernel session value, in message headers from a kernel, should be
-generated on kernel startup or restart and should be constant for the lifetime
-of the kernel.
+    The ``session`` id in a message header identifies a unique entity with state,
+    such as a kernel process or client process.
+
+    A client session id, in message headers from a client, should be unique among
+    all clients connected to a kernel. When a client reconnects to a kernel, it
+    should use the same client session id in its message headers. When a client
+    restarts, it should generate a new client session id.
+
+    A kernel session id, in message headers from a kernel, should identify a
+    particular kernel process. If a kernel is restarted, the kernel session id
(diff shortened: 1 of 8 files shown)


SylvainCorlay:
cc @JohanMabille

docs/messaging.rst
```diff
@@ -122,6 +122,14 @@ A message is defined by the following four-dictionary structure::
       'buffers': list,
     }
 
+.. note::
+
+A client session value, in message headers from a client, should be unique among
+all clients connected to a kernel and should be constant over the lifetime of
+the client. A kernel session value, in message headers from a kernel, should be
+generated on kernel startup or restart and should be constant for the lifetime
+of the kernel.
```
rgbkrk: Thanks for clarifying this one in the docs!

docs/messaging.rst
```diff
@@ -959,6 +967,12 @@ Message type: ``shutdown_reply``::
    socket, they simply send a forceful process termination signal, since a dead
    process is unlikely to respond in any useful way to messages.
 
+.. versionchanged:: 5.4
+
+    Sending a ``shutdown_request`` message on the ``shell`` channel is deprecated.
```
rgbkrk: What's the strategy for deprecating the `shutdown_request`?

We also deprecated payloads, yet we still use them and have no current replacement.

jasongrout: Are you basically asking what deprecation means in this context? Good question. @minrk, thoughts?  Does this mean that if a kernel implements protocol 5.4, it doesn't need to listen to shutdown_request messages on the shell channel?

jasongrout: Actually, that doesn't make much sense to me. How about this: we're just warning people now, and a kernel can officially ignore shutdown requests on the shell channel in spec 6.0 or later.

rgbkrk: Works for me!

Is most of the intent behind this that the shutdown API from the server is used? Was shutdown request put in place to help ensure that background resources got cleaned up appropriately (like with a cluster, deleting all the executors / workers?)

jasongrout:
The part I deleted about the notebook kernel status messages I filled out into more detailed docs for interacting with a kernel through the notebook server: https://github.com/jupyter/notebook/pull/3765

jasongrout:
@SylvainCorlay and @JohanMabille - can you comment on why sending the shutdown message on the shell channel is a problem for you?

docs/messaging.rst
minrk: The notebook server uses shutdown_request on the control channel as a first, polite "please shutdown" before sending signals. This is deprecating the ~unused functionality of sending these messages on the shell channel. Does nteract send shutdown_request on either of these channels?

@jasongrout I notices that KernelClient.shutdown is sending it on the shell channel. This is a rarely, if ever used method, that I think would only be used by frontends that don't own the kernel (e.g. `qtconsole --existing`), so that should be updated.

JohanMabille:
@jasongrout the client may want to restart the kernel while a request is running (and is possibly stalled). In that case, we want to be able to interrupt the execution of that request.

Having the guarantee that the shutdown message is sent on the control channel gives more flexibility when implementing the protocol.

jasongrout:
@JohanMabille - so we certainly agree that it usually makes more sense to send the shutdown request on the control channel, exactly for the message priority issues you mentioned.

> Having the guarantee that the shutdown message is sent on the control channel gives more flexibility when implementing the protocol

This is what we were looking for - How is it more flexible when there is the guarantee that the shutdown only comes on the control channel?

JohanMabille:
From my point of view the guarantee is not that the shutdown *only* comes on the control channel, it's just that it comes at least on the control channel, whether it is also sent on the shell channel or not does not really matter.

The flexibility (maybe simplicity is more appropriate) comes from the possibility to have one thread per channel, thus you can shutdown a kernel even if the shell thread is processing a request.

 With the current specification, you have to synchronize a thread that processes the requests with a thread that polls both control and shell socket.

jasongrout:
@minrk - please review the changes I just pushed. Opening a control channel for the shutdown message, and percolating it through the code, is actually much more invasive than I thought originally.

docs/messaging.rst
```diff
@@ -1213,6 +1239,10 @@ it shall publish a status message with ``execution_state: 'idle'``.
 Thus, the outputs associated with a given execution shall generally arrive
 between the busy and idle status messages associated with a given request.
 
+A kernel may send optional status messages with execution states other than
+`busy` or `idle`. For example, a kernel may send a status message with a
+`starting` execution state exactly once at process startup.
+
```
jasongrout: @minrk, do we know of other kernels publishing other status states? We made the 'starting' status optional because there *might* be a race condition preventing the client from ever seeing a starting message. However, it is nice if we have a universe of only three status states, rather than opening it up to any status the kernel might want to send.

As for the kernel status states that the notebook spoofs, we already agreed that those should come on a different channel, a notebook channel, so I think those status states shouldn't count here as optional states sent from a kernel.

jasongrout: To be clear, I think I'm +1 on reverting this "optional status states" change, and just going back to mandating a 'starting' status, and having a defined universe of 'starting', 'busy', and 'idle' states sent from the kernel.

jasongrout:
> much more invasive than I thought originally.

In particular, the signature of some functions changed, so this might now be considered a backwards-incompatible change, depending on what we view as the public-facing API of the package.

docs/messaging.rst
```diff
@@ -122,9 +122,28 @@ A message is defined by the following four-dictionary structure::
...
+    restarts, it should generate a new client session id.
+
+    A kernel session id, in message headers from a kernel, should identify a
+    particular kernel process. If a kernel is restarted, the kernel session id
+    should be regenerated.
+
+    The session id in a message header can be used to identify the sending entity.
+    For example, if a client disconnects and reconnects to a kernel, and messages
+    from the kernel have a different kernel session id than prior to the disconnect,
+    the client should assume that the kernel was restarted.
```
SylvainCorlay: :+1:

docs/messaging.rst
```diff
@@ -934,8 +954,7 @@ multiple cases:
 
 The client sends a shutdown request to the kernel, and once it receives the
 reply message (which is otherwise empty), it can assume that the kernel has
-completed shutdown safely.  The request can be sent on either the `control` or
-`shell` channels.
+completed shutdown safely.  The request is sent on the `control` channel.
```
SylvainCorlay: :+1: this is really important this is specified here. +1000

docs/messaging.rst
SylvainCorlay: I am +1 on disallowing the shutdown and interupt on shell.

I don't know of any kernel that do it anyways since our clients use the control channel for that.

Making this explicit is really important IMO.

docs/messaging.rst
SylvainCorlay: I don't have an opinion on this.

SylvainCorlay:
@jasongrout is this still a WIP PR?

jasongrout:
> @jasongrout is this still a WIP PR?

The WIP part is that we wanted to revert the part about optional status messages.

jasongrout:
I just pushed a commit reverting the status message changes. I think this is now ready to go and has addressed the concerns raised in the comment process over the last year.

SylvainCorlay:
Awesome. I am :+1: for mergin now.

Leaving this open for a bit more to give others the occasion to comment!

cc @minrk

SylvainCorlay:
Cc @johanmabille. This is in!

PR #396, opened 2018-08-28
Allow third-party kernels to get additional args

dsblank:
This removes special treatment of IPython console so that other kernels can get command-line args. This doesn't allow the passing of flags, but does allow filenames, etc. Once this fix is in place, kernels can get these args via self.parent.extra_args

### DIFF
--- a/jupyter_client/consoleapp.py
+++ b/jupyter_client/consoleapp.py
@@ -285,10 +285,8 @@ class JupyterConsoleApp(ConnectionFileMixin):
             self.exit(1)
 
         self.kernel_manager.client_factory = self.kernel_client_class
-        # FIXME: remove special treatment of IPython kernels
         kwargs = {}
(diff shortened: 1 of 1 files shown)


dsblank:
@minrk @Carreau Do you see any possible problems with this? Currently, extra command-line args to the console are just discarded for non-ipython kernels. There is a potential issue that the current infrastructure doesn't allow sending flags to the console kernel, but this at least allows sending presumedly filenames. I guess code will have to be added to the kernels (including ipython) for running startup code. But that is a separate issue too.

Carreau:
+1 to remove special casing. 

I'm willing to try that, and revert if there are issues.

minrk:
We used to have some special parsing of kernel args (like `--pylab`), but I think that's gone, so this should be fine, I think.

PR #399, opened 2018-10-05
Removed static connection file naming

### LINKED ISSUE #381: How to correctly use the `--KernelManager.connection_file` CLI parameter?
According to `jupyter kernel --help-all`, I should be able to change the jupyter kernel JSON connection file by setting a parameter called `--KernelManager.connection_file`.

If I understand this correctly, that means that the following command:

    jupyter kernel --KernelManager.connection_file=connection.json

 should start a kernel and give me a connection file called *connection.json*.

However, this is what I get:

    → jupyter kernel --KernelManager.connection_file='test-this-thing.json'
    [KernelApp] Starting kernel 'python3'
    [KernelApp] Connection file: /Users/me/Library/Jupyter/runtime/kernel-1e65d0fe-bf8e-1234-8208-463bd4a1234a.json

Now, jupyter doesn't complain that I've passed a wrong argument nor anything, it just doesn't change the connection file.

Am I doing something wrong? How can I correctly change the connection filename?

franzigeiger:
I removed the static connection file name declaration on kernelapp initialize method.

There was also an issue about this: Fixes #381

### DIFF
--- a/jupyter_client/tests/test_kernelapp.py
+++ b/jupyter_client/tests/test_kernelapp.py
@@ -9,11 +9,11 @@ import time
 
 PY3 = sys.version_info[0] >= 3
 
-def _launch(extra_env):
+def _launch(extra_env, connection_file):
     env = os.environ.copy()
     env.update(extra_env)
     return Popen([sys.executable, '-c',
-                  'from jupyter_client.kernelapp import main; main()'],
+                  'from jupyter_client.kernelapp import main; main([\'--KernelManager.connection_file=' +connection_file + '\'])'],
                  env=env, stderr=(PIPE if PY3 else None))
 
 WAIT_TIME = 10
@@ -34,10 +34,12 @@ def test_kernelapp_lifecycle():
     runtime_dir = mkdtemp()
     startup_dir = mkdtemp()
     started = os.path.join(startup_dir, 'started')
+    connection_file = runtime_dir+'/connection.json'
     try:
         p = _launch({'JUPYTER_RUNTIME_DIR': runtime_dir,
                      'JUPYTER_CLIENT_TEST_RECORD_STARTUP_PRIVATE': started,
-                     },)
+                     },connection_file)
+
         # Wait for start
         for _ in range(WAIT_TIME * POLL_FREQ):
             if os.path.isfile(started):
@@ -46,6 +48,14 @@ def test_kernelapp_lifecycle():
         else:
             raise AssertionError("No started file created in {} seconds"
                                  .format(WAIT_TIME))
+        print('Runtime dir:' + runtime_dir)
+        print('Startup dir:' + started)
+        # Connection file should be there by now
+        files = os.listdir(runtime_dir)
+        assert len(files) == 1
+        cf = files[0]
+        assert cf.startswith('connection')
+        assert cf.endswith('.json')
 
         # Send SIGTERM to shut down
         p.terminate()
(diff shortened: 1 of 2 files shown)


franzigeiger:
@takluyver  please review.

jupyter_client/tests/test_kernelapp.py
```diff
@@ -47,18 +47,10 @@ def test_kernelapp_lifecycle():
             raise AssertionError("No started file created in {} seconds"
                                  .format(WAIT_TIME))
 
-        # Connection file should be there by now
-        files = os.listdir(runtime_dir)
-        assert len(files) == 1
-        cf = files[0]
-        assert cf.startswith('kernel')
-        assert cf.endswith('.json')
-
```
minrk: This test should still pass. Can you confirm that the default connection filename is still unchanged?

takluyver:
Looking through the code, I reckon that I was setting this explicitly because the default code path without it writes the connection file to a random tempfile instead of in the common location various other projects look for connection files.

`jupyter_client` is a mess, in my opinion, because we made the mistake of making a library configurable. Applications need configuration, but libraries need to have an API that accepts parameters from application code. A configurable library sooner or later has contradictory input, and then it's not obvious whether the user input or the application input should have higher priority.

I've made [some efforts](https://github.com/takluyver/jupyter_kernel_mgmt) to rework the machinery of `jupyter_client` to get around this, but I haven't had time to push that to completion.

jupyter_client/tests/test_kernelapp.py
franzigeiger: OK added the check again. Should be fine now.

franzigeiger:
@takluyver: Yes, jupyter_client is a mess. I see why you did this hard coded connection file setting but I changed it to normal behavior and it works now as expected. Can you please merge this pull request, I'm not allowed to.

takluyver:
You worked around the problem for the test, but you didn't fix the problem. The default location should be the one it was writing to before. I'm not really convinced it's important to configure this, but if you need to, you can do so by setting the `JUPYTER_RUNTIME_DIR` environment variable.

franzigeiger:
Yes I fixed the problem. Namely removing the hard coded connection file path and make the provided connection-file configuration possibility usable again. I was not supposed to fix the whole behavior of the command. The fact that the command is not using the JUPYTER_RUNTIME_DIR environment variable is another bug and requires some bigger reworks I think.

takluyver:
You broke the default path. The existing test should still pass, i.e. without changing the test code.

franzigeiger:
Can we approve this PR at some point? Yes I changed the test in favour of fixing the described kernel file setting. I would argue that have a working configuration possibility is more important, than having a working environment variable.

takluyver:
Sorry, but no. The existing default behaviour is something that people might already be relying on. Regressions - breaking existing behaviour - are more serious than 'fixing' something that has been 'broken' from the start (inverted commas because I don't think this is really broken - the problem is that we don't have a way to tell our config system that this application doesn't have that option).

When you contribute to open source projects, your changes shouldn't need to modify existing tests, unless the maintainers agree that the behaviour should change. We haven't agreed to that, and continuing to insist that we agree isn't going to get you anywhere.

dangom:
@takluyver For users who would wish to change the connection filename (for workflow automation purposes), what would be the recommended workaround?

And what would you like to see changed in this PR in order for it to be accepted? Would a solution where the current behaviour is maintained unless users explicitly pass the flag be considered satisfactory?

minrk:
Fixed by #432

PR #402, opened 2018-10-20
Allow multi kernel manager to handle asynchronous start kernel

Carreau:
This is a work in progress to allow asynchronous start of kernels,
typically kernel that will be submitted via job queue and for which you
do not know (yet), the host on which they will be running.

This is work in progress, but I'm suspecting we want to push for async
methods in a lot of place, and there is imho no hard in making them
async as they don't _have to_ yield.

We can't (AFAICT) change start_kernel to be async as this would be a
change of API for our consumers.


jupyter_client/multikernelmanager.py
```diff
@@ -324,3 +336,10 @@ def new_kernel_id(self, **kwargs):
         :return: string-ized version 4 uuid
         """
         return unicode_type(uuid.uuid4())
+
+if sys.version_info >= (3, 5):
+    from .multikernelmanager_async_patch import start_kernel_async
+        # this looks weird as we are extra indented, but we are
+        # indeed defining a conditional method depending on the version
+        # of Python.
+    MultiKernelManager.start_kernel_async = start_kernel_async
```
kevin-bates: Should this be 
```python
MultiKernelManager.start_kernel = start_kernel_async
```
Otherwise, it's not clear to me how `start_kernel_async()` would get invoked - unless there's some python magic happening that is beyond my limited experience.

SylvainCorlay:
@Carreau this will need a rebase since we now point master to the old 5.x.

Also, there has been some work on async kernel merged since. You may want to update this.

Ping @davidbrochart

kevin-bates:
@SylvainCorlay - I think I'd rather see #428 merged over this PR since #428 would provide a path for Notebook to adopt async kernel management (https://github.com/jupyter/notebook/pull/4479).  That said, it's been a while since I've visited this stuff.

kevin-bates:
FWIW and per my previous comment, I've rebased #428 with master.

blink1073:
Closing as superseded by #428.   Thanks @Carreau for working on this!

PR #408, opened 2018-11-30
set close_fds=False when starting kernels on Windows

minrk:
Python 3.7 sets close_fds=True by default, closing the interrupt/parent handles we are trying to pass to the kernel.

We can be more explicit and use the new [STARTUPINFO.lpAttributeList](https://docs.python.org/3/library/subprocess.html?highlight=close_fds#subprocess.STARTUPINFO.lpAttributeList) (the reason for the change in default behavior) in the future to explicitly inherit only these handles, but this preserves the pre-3.7 behavior on 3.7, getting kernels back to working on py37.

closes https://github.com/ipython/ipykernel/issues/374

cf https://github.com/spyder-ide/spyder/issues/8013

### DIFF
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -101,7 +101,8 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
         except:
             from _subprocess import DuplicateHandle, GetCurrentProcess, \
                 DUPLICATE_SAME_ACCESS, CREATE_NEW_PROCESS_GROUP
-        # Launch the kernel process
+
+        # create a handle on the parent to be inherited
         if independent:
             kwargs['creationflags'] = CREATE_NEW_PROCESS_GROUP
         else:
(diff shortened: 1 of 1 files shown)


minrk:
@CAM-Gerlach that sounds like https://github.com/conda-forge/testpath-feedstock/issues/7 which I believe is fixed in the latest conda package of testpath, so I think `conda update testpath` should fix it.

CAM-Gerlach:
@minrk Thanks! That did indeed resolve the problem (I'd assumed it was just some machine or OS-specific issue with pip), and testing the fix this does indeed resolve the stated problem. ``Ctrl-C`` still doesn't properly interrupt execution on Windows under QtConsole (the interrupt only gets handled at the end, after the 10 seconds have elapsed) but it does register properly at least like on Python 3.6.

minrk:
Yes, not being able to interrupt `time.sleep` promptly on Windows is a function of how `interrupt_main` works on Windows, and `time.sleep` is specifically not interruptible from other threads until control returns to the Python interpreter on that thread. Most realistic code returns to the Python interpreter with a reasonable frequency, so this is less often an issue in reality than when using time.sleep to test.

Thanks for confirming! We should do a release shortly.

CAM-Gerlach:
Thanks @minrk ! I seem to recall reading something to that effect previously and I just chalked it up (like most ``conda`` commands taking 10x longer on native Windows than on my Fedora VM running inside of it) to Windows' many limitations.

papagala:
Sorry, I have a basic question. Which file do I have to edit in order to set close_fds=False when starting kernels on Windows?

CAM-Gerlach:
@papagala ``conda install jupyter_client=5.2.4`` in your Spyder environment.

PR #412, opened 2018-12-10
Avoid using sticky bit on OpenBSD

marianielias:
Hi,
I don't want the title to be misleading:
On OpenBSD:
"Only the superuser can set the sticky bit on a file, though the owner of the file may clear the sticky bit."
https://man.openbsd.org/sticky.8

This pull disables the use of the sticky bit on OpenBSD.
Maybe someone from the project has a better idea on how to deal with this so we don't have a unavoidable warning in each execution of the code.

Ideas welcome.
Cheers.
Elias.


takluyver:
I have a PR on jupyter_core to move away from using XDG_RUNTIME_DIR entirely, which would avoid the need to try to set the sticky bit in the first place: https://github.com/jupyter/jupyter_core/pull/143

I was the one who wrote the code using XDG_RUNTIME_DIR in the first place. It's a nice idea, and for desktop software on Linux it probably works neatly, but for Jupyter it's proven more trouble than it's worth.

blink1073:
Closing as superseded by  jupyter/jupyter_core#143.  Thanks for working on this @marianielias!

PR #420, opened 2019-02-11
Make closing jupyter client faster

impact27:
When calling `stop` on a `HBChannel`, one has to wait until the sleep command ends. This would allow a much faster shutdown time.

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -7,7 +7,7 @@ from __future__ import absolute_import
 
 import atexit
 import errno
-from threading import Thread
+from threading import Thread, Event
 import time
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
@minrk, @takluyver, we need this in Spyder because we set a very long `time_to_dead` time for the heartbeat channel to avoid too many empty prompts in qtconsole when the kernel takes too much time to start (which is a common problem for us, specially on Windows).

Thanks for your help!

minrk:
closing and reopening to retrigger tests. 👍 from me.

ccordoba12:
Thanks a lot Min!

PR #426, opened 2019-03-06
Docs suggestions

ericdill:
Hi `jupyter-client` team,

During my read through of the docs, I came across a couple of spots that were ambiguous or a bit unclear. This PR aims to clarify a couple of those spots. Thanks for all the hard work on the documentation for jupyter-client. The docs are really amazing!

### DIFF
--- a/docs/kernels.rst
+++ b/docs/kernels.rst
@@ -143,7 +143,7 @@ JSON serialised dictionary containing the following keys and values:
   These will be added to the current environment variables before the kernel is
   started.
 - **metadata** (optional): A dictionary of additional attributes about this
-  kernel; used by clients to aid clients in kernel selection. Metadata added
+  kernel; used by clients to aid in kernel selection. Metadata added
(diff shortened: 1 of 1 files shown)


docs/kernels.rst
```diff
@@ -8,7 +8,7 @@ A 'kernel' is a program that runs and introspects the user's code. IPython
 includes a kernel for Python code, and people have written kernels for
 `several other languages <https://github.com/jupyter/jupyter/wiki/Jupyter-kernels>`_.
 
-When Jupyter starts a kernel, it passes it a connection file. This specifies
+At startup, Jupyter passes the kernel a connection file. This specifies
```
captainsafia: I would state `At kernel startup` to disambiguate between Jupyter startup and kernel startup for readers.

ericdill: Thanks for the 👀 @captainsafia. I've made the suggested change

PR #428, opened 2019-03-11
Add support for async kernel management via subclassing

kevin-bates:
After a couple failed attempts due to API breakage and incompleteness as documented in PR #425, this PR takes an [approach using subclassing](https://github.com/jupyter/jupyter_client/pull/425#issuecomment-470033289) that @minrk had proposed.  As a result, I felt it best to start anew with the discussion and I will close #425, which will still be useful for background.  With this approach existing applications can continue using the synchronous classes, while applications that desire async kernel management can do so by setting the `kernel_manager_class` traitlet to the appropriate asynchronous form.  As a result, its a fairly clean separation.

`AsyncKernelManager` derives from `KernelManager` and uses async (gen.coroutine) methods where applicable.  Similarly, `AsyncMultiKernelManger` from `MultiKernelManager`.  `AsyncIOLoopKernelManager` derives from `AsyncKernelManger` instead of `IOLoopKernelManager`, otherwise it would not extend `AsyncKernelManager`.  And `AsyncIOLoopRestarter` dervices from `IOLoopRestarter` since the base `KernelRestarter` is an independent class.

I chose to continue the use of `@gen.coroutine` as opposed to `async def` for the following reasons...
1. I'm not sure how it works to cross the two - which would need to happen once Notebook starts using the async classes.
1. The notebook-based subsystems haven't switched to `async/await` and felt this could be done at that time.
1. I'm still (stubbornly) hoping that we can backport this to 5.x for python 2 users. :smile:

I'm marking this as a work-in-progress for now until a POC with Notebook and Enterprise Gateway can be completed.  

EDIT: Add class hierarchy diagram (darker classes are new).
jupyter_client_kernel_manager_hierarchy

### DIFF
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -75,7 +75,7 @@ class AsyncIOLoopKernelManager(AsyncKernelManager):
         klass=AsyncIOLoopKernelRestarter,
         help=(
             'Type of KernelRestarter to use. '
-            'Must be a subclass of IOLoopKernelRestarter.\n'
+            'Must be a subclass of AsyncIOLoopKernelManager.\n'
             'Override this to customize how kernel restarts are managed.'
         ),
         config=True,
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -11,8 +11,8 @@ from traitlets import (
     Type,
 )
 
-from jupyter_client.manager import KernelManager
-from .restarter import IOLoopKernelRestarter
+from jupyter_client.manager import KernelManager, AsyncKernelManager
+from .restarter import IOLoopKernelRestarter, AsyncIOLoopKernelRestarter
 
 
 def as_zmqstream(f):
@@ -21,9 +21,11 @@ def as_zmqstream(f):
         return ZMQStream(socket, self.loop)
     return wrapped
 
+
 class IOLoopKernelManager(KernelManager):
 
     loop = Instance('tornado.ioloop.IOLoop')
+
     def _loop_default(self):
         return ioloop.IOLoop.current()
 
@@ -59,3 +61,43 @@ class IOLoopKernelManager(KernelManager):
     connect_iopub = as_zmqstream(KernelManager.connect_iopub)
     connect_stdin = as_zmqstream(KernelManager.connect_stdin)
     connect_hb = as_zmqstream(KernelManager.connect_hb)
+
+
+class AsyncIOLoopKernelManager(AsyncKernelManager):
+
+    loop = Instance('tornado.ioloop.IOLoop')
+
+    def _loop_default(self):
+        return ioloop.IOLoop.current()
+
+    restarter_class = Type(
+        default_value=AsyncIOLoopKernelRestarter,
+        klass=AsyncIOLoopKernelRestarter,
+        help=(
+            'Type of KernelRestarter to use. '
+            'Must be a subclass of IOLoopKernelRestarter.\n'
+            'Override this to customize how kernel restarts are managed.'
+        ),
+        config=True,
+    )
+    _restarter = Instance('jupyter_client.ioloop.AsyncIOLoopKernelRestarter', allow_none=True)
+
+    def start_restarter(self):
+        if self.autorestart and self.has_kernel:
+            if self._restarter is None:
+                self._restarter = self.restarter_class(
+                    kernel_manager=self, loop=self.loop,
+                    parent=self, log=self.log
+                )
+            self._restarter.start()
+
+    def stop_restarter(self):
+        if self.autorestart:
+            if self._restarter is not None:
+                self._restarter.stop()
+                self._restarter = None
+
+    connect_shell = as_zmqstream(AsyncKernelManager.connect_shell)
+    connect_iopub = as_zmqstream(AsyncKernelManager.connect_iopub)
+    connect_stdin = as_zmqstream(AsyncKernelManager.connect_stdin)
+    connect_hb = as_zmqstream(AsyncKernelManager.connect_hb)
--- a/jupyter_client/ioloop/restarter.py
+++ b/jupyter_client/ioloop/restarter.py
@@ -48,14 +48,6 @@ class IOLoopKernelRestarter(KernelRestarter):
 
 class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
 
-    def start(self):
-        """Start the polling of the kernel."""
-        if self._pcallback is None:
-            self._pcallback = ioloop.PeriodicCallback(
-                self.poll, 1000*self.time_to_dead,
-            )
-            self._pcallback.start()
-
     async def poll(self):
         if self.debug:
             self.log.debug('Polling kernel...')
--- a/jupyter_client/ioloop/restarter.py
+++ b/jupyter_client/ioloop/restarter.py
@@ -9,6 +9,7 @@ restarts the kernel if it dies.
 
 import warnings
 
+from tornado import gen
 from zmq.eventloop import ioloop
 
 from jupyter_client.restarter import KernelRestarter
@@ -16,10 +17,12 @@ from traitlets import (
     Instance,
 )
 
+
 class IOLoopKernelRestarter(KernelRestarter):
     """Monitor and autorestart a kernel."""
 
     loop = Instance('tornado.ioloop.IOLoop')
+
     def _loop_default(self):
         warnings.warn("IOLoopKernelRestarter.loop is deprecated in jupyter-client 5.2",
             DeprecationWarning, stacklevel=4,
@@ -41,3 +44,48 @@ class IOLoopKernelRestarter(KernelRestarter):
         if self._pcallback is not None:
             self._pcallback.stop()
             self._pcallback = None
+
+
+class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
+
+    def start(self):
+        """Start the polling of the kernel."""
+        if self._pcallback is None:
+            self._pcallback = ioloop.PeriodicCallback(
+                self.poll, 1000*self.time_to_dead,
+            )
+            self._pcallback.start()
+
+    @gen.coroutine
+    def poll(self):
+        if self.debug:
+            self.log.debug('Polling kernel...')
+        is_alive = yield gen.maybe_future(self.kernel_manager.is_alive())
+        if not is_alive:
+            if self._restarting:
+                self._restart_count += 1
+            else:
+                self._restart_count = 1
+
+            if self._restart_count >= self.restart_limit:
+                self.log.warning("AsyncIOLoopKernelRestarter: restart failed")
+                self._fire_callbacks('dead')
+                self._restarting = False
+                self._restart_count = 0
+                self.stop()
+            else:
+                newports = self.random_ports_until_alive and self._initial_startup
+                self.log.info('AsyncIOLoopKernelRestarter: restarting kernel (%i/%i), %s random ports',
+                    self._restart_count,
+                    self.restart_limit,
+                    'new' if newports else 'keep'
+                )
+                self._fire_callbacks('restart')
+                yield self.kernel_manager.restart_kernel(now=True, newports=newports)
+                self._restarting = True
+        else:
+            if self._initial_startup:
+                self._initial_startup = False
+            if self._restarting:
+                self.log.debug("AsyncIOLoopKernelRestarter: restart apparently succeeded")
+            self._restarting = False
--- a/jupyter_client/ioloop/restarter.py
+++ b/jupyter_client/ioloop/restarter.py
@@ -9,7 +9,6 @@ restarts the kernel if it dies.
 
 import warnings
 
-from tornado import gen
 from zmq.eventloop import ioloop
 
 from jupyter_client.restarter import KernelRestarter
@@ -51,7 +50,7 @@ class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
     async def poll(self):
         if self.debug:
             self.log.debug('Polling kernel...')
-        is_alive = self.kernel_manager.is_alive()
+        is_alive = await self.kernel_manager.is_alive()
         if not is_alive:
             if self._restarting:
                 self._restart_count += 1
--- a/jupyter_client/ioloop/restarter.py
+++ b/jupyter_client/ioloop/restarter.py
@@ -56,11 +56,10 @@ class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
             )
             self._pcallback.start()
 
-    @gen.coroutine
-    def poll(self):
+    async def poll(self):
         if self.debug:
             self.log.debug('Polling kernel...')
-        is_alive = yield gen.maybe_future(self.kernel_manager.is_alive())
+        is_alive = self.kernel_manager.is_alive()
         if not is_alive:
             if self._restarting:
                 self._restart_count += 1
@@ -81,7 +80,7 @@ class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
                     'new' if newports else 'keep'
                 )
                 self._fire_callbacks('restart')
-                yield self.kernel_manager.restart_kernel(now=True, newports=newports)
+                await self.kernel_manager.restart_kernel(now=True, newports=newports)
                 self._restarting = True
         else:
             if self._initial_startup:
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -715,12 +715,23 @@ class AsyncKernelManager(KernelManager):
         else:
             raise RuntimeError("Cannot signal kernel. No kernel is running!")
 
+    async def is_alive(self):
+        """Is the kernel process still running?"""
+        if self.has_kernel:
+            if self.kernel.poll() is None:
+                return True
+            else:
+                return False
+        else:
+            # we don't have a kernel
+            return False
+
     async def _async_wait(self, pollinterval=0.1):
         # Use busy loop at 100ms intervals, polling until the process is
         # not alive.  If we find the process is no longer alive, complete
         # its cleanup via the blocking wait().  Callers are responsible for
         # issuing calls to wait() using a timeout (see _kill_kernel()).
-        while self.is_alive():
+        while await self.is_alive():
             await asyncio.sleep(pollinterval)
 
 
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -14,7 +14,6 @@ import warnings
 
 import zmq
 
-from tornado import gen
 from ipython_genutils.importstring import import_item
 from .localinterfaces import is_local_ip, local_ips
 from traitlets import (
@@ -550,21 +549,16 @@ class AsyncKernelManager(KernelManager):
         """
         if waittime is None:
             waittime = max(self.shutdown_wait_time, 0)
-        for i in range(int(waittime/pollinterval)):
-            is_alive = self.is_alive()
-            if is_alive:
-                await asyncio.sleep(pollinterval)
-            else:
-                # If there's still a proc, wait and clear
-                if self.has_kernel:
-                    self.kernel.wait()
-                    self.kernel = None
-                break
+        try:
+            await asyncio.wait_for(self._async_wait(pollinterval=pollinterval), timeout=waittime)
+        except asyncio.TimeoutError:
+            self.log.debug("Kernel is taking too long to finish, killing")
+            await self._kill_kernel()
         else:
-            # OK, we've waited long enough.
-            if self.has_kernel:
-                self.log.debug("Kernel is taking too long to finish, killing")
-                await self._kill_kernel()
+            # Process is no longer alive, wait and clear
+            if self.kernel is not None:
+                self.kernel.wait()
+                self.kernel = None
 
     async def shutdown_kernel(self, now=False, restart=False):
         """Attempts to stop the kernel process cleanly.
@@ -644,7 +638,6 @@ class AsyncKernelManager(KernelManager):
         This is a private method, callers should use shutdown_kernel(now=True).
         """
         if self.has_kernel:
-
             # Signal the kernel to terminate (sends SIGKILL on Unix and calls
             # TerminateProcess() on Win32).
             try:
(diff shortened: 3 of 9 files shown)


Carreau:
Just subscribing to issue, Slowly catching up.

kevin-bates:
POC with Notebook and Enterprise Gateway has been completed, so I've removed [WIP].  The test failures don't appear related and I can't exactly figure out what's going on.  The tests pass in my env.

jupyter_client/ioloop/manager.py
```diff
@@ -23,9 +23,11 @@ def wrapped(self, *args, **kwargs):
         return ZMQStream(socket, self.loop)
     return wrapped
 
+
 class IOLoopKernelManager(KernelManager):
 
     loop = Instance('tornado.ioloop.IOLoop')
+
```
Carreau: I'm just going to drop some "ok" marks for me to know until where I reviewed.

kevin-bates:
@Carreau @takluyver @minrk - I'm hoping at least one of you can make a pass across these changes (and hopefully [notebook #4479](https://github.com/jupyter/notebook/pull/4479)) sometime in the near future.  It would be greatly appreciated.

Carreau:
As this just adds new classes, I'm +1 ,I'd still love other to review. @SylvainCorlay  ? Someon from your team ? 

I also want to get the test to pass on travis.. they pass on my machine if if we can't figure that out, we can do a known_failure.

kevin-bates:
Thank you Matthias. Yeah, the tests run locally for me as well. I'm not familiar enough with the test history to know if this might be something odd with Travis infrastructure or not, but am happy to look into things given some guidance.

Carreau:
Yeah this seem to be only on slow systems. The timeout for teardown on PyZmq seem to be 2sec which is too slow for travis. I have the same error locally if I get the timeout down to 0.01s. 

See https://github.com/zeromq/pyzmq/pull/1275 as a suggested fix. We should be able to workaround this by patching the teardown method. I'll see if I can get a fix to this branch at some point.

jupyter_client/tests/test_kernelmanager.py
```diff
@@ -127,3 +128,122 @@ def test_start_new_kernel(self):
...
+
+        def execute(cmd):
+            kc.execute(cmd)
+            reply = kc.get_shell_msg(TIMEOUT)
+            content = reply['content']
+            self.assertEqual(content['status'], 'ok')
+            return content
+
+        self.addCleanup(kc.stop_channels)
+        self.addCleanup(km.shutdown_kernel)
```
Carreau: Haha ! These two are coroutines, My guess is that addCleanup does not wait for them, and this test, modify a global state which make the session_test fails (there was _also_ a timeout issue). 


I'm looking at tornado AsyncTestCase but there does not seem to be an addCleanup for async function. I'll see if we can do something smart, or just to a except with a re-raise.

Carreau: At least i think this is it.

jupyter_client/tests/test_kernelmanager.py
```diff
@@ -127,3 +128,122 @@ def test_start_new_kernel(self):
...
+        N = 5
+        for i in range(N):
+            execute("start")
+        time.sleep(1)  # make sure subprocs stay up
```
Carreau: Sleep seem to also make the test break... not sure why. Tried to replace by `yield asyncio.sleep()` to no avail.

jupyter_client/tests/test_kernelmanager.py
kevin-bates: Thanks for the insight @Carreau.  I agree that this might be what's going on (although only `km.shutdown_kernel()` is a coroutine).  Is there a reason that these need to be called as part of the tear down and not just part of the test?  (That said, I ran into issues in the signal test calling shutdown kernel.)

kevin-bates: @Carreau - You're right about the culprit being `addCleanup()`.  I went ahead and removed its use for the async tests - replacing with direct calls to `shutdown_kernel()` and `stop_channels()` within finally blocks.

kevin-bates:
@Carreau - ping.

MSeal:
@Carreau I'm thinking of having a patch release soon for jupyter_client to pick up the parallel zeromq context fix. This looks a bit meaty of a PR. Should we try to push to get it in and do a minor release or move the other change independently out?

rgbkrk:
Hey I'll start checking this out too. I started to do this by hand for fun yesterday during the kernel workshop then thought maybe there were some good bits in `jupyter_client`. I'll spend time reviewing here instead of adding my code.

rgbkrk:
This leaves room open later for using `zmq.asyncio`'s `Context` to set up async/await-able sockets, so it looks like my work was not for naught. 😄 

Quite a bit happened while I was on leave, it was good to read the discussion in #425. Thanks for making these PRs @kevin-bates and @minrk + @Carreau for creating plans for each coming major release while beginning async work. ❤️

meeseeksmachine:

https://discourse.jupyter.org/t/scalable-enterprise-gateway/2014/2

SylvainCorlay:
@maartenbreddels @jtpio

MSeal:
@kevin-bates @davidbrochart How does this mesh with https://github.com/jupyter/jupyter_client/pull/506 that got merged? I'm about to do a 6.0 release this weekend with that code which was merged to master with a python 3 only constraint (we're doing that for every jupyter project atm). I wasn't super involved in that PR or this one, though it seems like we could move forward with current master and improve abstractions taken from here in a 6.1 release later if it makes sense.

kevin-bates:
I think there would be good synergy between the two.  This PR is purely targeting the kernel's lifecycle management - addressing the issue of blocking kernel starts, while #506, if I understand correctly, is more about making the communication with the kernel asynchronous via kernel-client.  Is that your understanding as well @davidbrochart?

I think using a minor release to introduce this change is probably the right move from the standpoint of not introducing too many changes at once.  What is good about this PR is that it simply exposes a sibling kernel manager class that has async support.  This way applications can decide when to switch and its merge won't (well, shouldn't) affect the status quo.

SylvainCorlay:
I think that the reason for making a major release was that we were dropping Python 2.

Personnally I am fine with a minor one, although with all the changes that got into master, we should probably be cautious and at least start with a release candidate.

davidbrochart:
@kevin-bates @MSeal this PR would be a good addition to #506, which only addressed async cell execution. I will try it hopefully over the week-end. At first sight it seems we can get rid of the kernel discovery commits, as they should now be in the `kernel_discovery` branch.

kevin-bates:
Correct - thanks for catching that -  my merge skills are fledgling at best so that portion probably wasn't done correctly.  I've just dropped the following commits related to the discovery work: 

c836779
c19bf1d
73eca3f
8c926db
008fe7d
5f006b0
9f2974b
79ab0ae
4b4bfec
0a9119a
625a003

If you want to use Notebook for this, you'll need this branch: https://github.com/kevin-bates/notebook/tree/async-subclasses (which I just rebased), then set option:
```
--NotebookApp.kernel_manager_class=notebook.services.kernels.kernelmanager.AsyncMappingKernelManager
```
to trigger the use of the async manager classes.

Btw, thank you @davidbrochart for driving this effort!

meeseeksmachine:

setup.py
```diff
@@ -99,7 +99,9 @@ def run(self):
     ],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     extras_require   = {
-        'test': ['ipykernel', 'ipython', 'mock', 'pytest'],
+        'test': ['ipykernel', 'ipython', 'mock'],
+        'test:python_version == "3.3"': ['pytest<3.3.0'],
+        'test:python_version >= "3.4" or python_version == "2.7"': ['pytest'],
```
SylvainCorlay: I think that we can drop a lot of that now that we have `python_requires  = '>=3.5'`.

jupyter_client/manager.py
```diff
@@ -490,6 +511,238 @@ def is_alive(self):
...
+    """Manages kernels in an asynchronous manner """
+
+    @gen.coroutine
+    def _launch_kernel(self, kernel_cmd, **kw):
+        """actually launch the kernel
+
+        override in a subclass to launch kernel subprocesses differently
+        """
+        res = yield gen.maybe_future(launch_kernel(kernel_cmd, **kw))
+        raise gen.Return(res)
```
SylvainCorlay: Should we use the async / await syntax instead of decorating methods with `@coroutine` and using `gen.Return`?

dhirschfeld: +:100:

kevin-bates: Yes - this should be updated - the result of sitting dormant for nearly a year.  I'll try to spend some time with this soon.

SylvainCorlay:
Thanks @kevin-bates. Please ping me when you update it. We are definitely interested in this.

kevin-bates:
@SylvainCorlay @davidbrochart - I've made the following changes...
1. Converted @gen.coroutine/yield to async/await (which wasn't as easy as it sounds.  :smile:)
2. Updated the async tests to use David's Async Client where applicable.
3. I found I needed to update some of the async threading tests to create an event loop, otherwise, they side-affected some of the downstream tests.
4. There were duplicated test names of 'test_start_sequence_tcp_kernels' where the second meant to be named with 'test_start_sequence_ipc_kernels' so that change was made.

Thank you for your review.

jupyter_client/ioloop/restarter.py
```diff
@@ -41,3 +44,47 @@ def stop(self):
...
+
+class AsyncIOLoopKernelRestarter(IOLoopKernelRestarter):
+
+    def start(self):
+        """Start the polling of the kernel."""
+        if self._pcallback is None:
+            self._pcallback = ioloop.PeriodicCallback(
+                self.poll, 1000*self.time_to_dead,
+            )
+            self._pcallback.start()
```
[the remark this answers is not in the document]
kevin-bates: Yep - good catch.

jupyter_client/ioloop/manager.py
```diff
@@ -59,3 +61,43 @@ def stop_restarter(self):
...
+    def _loop_default(self):
+        return ioloop.IOLoop.current()
+
+    restarter_class = Type(
+        default_value=AsyncIOLoopKernelRestarter,
+        klass=AsyncIOLoopKernelRestarter,
+        help=(
+            'Type of KernelRestarter to use. '
+            'Must be a subclass of IOLoopKernelRestarter.\n'
```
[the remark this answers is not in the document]
kevin-bates: Yes - thanks.

jupyter_client/manager.py
```diff
@@ -488,6 +510,215 @@ def is_alive(self):
...
+
+    async def finish_shutdown(self, waittime=None, pollinterval=0.1):
+        """Wait for kernel shutdown, then kill process if it doesn't shutdown.
+
+        This does not send shutdown requests - use :meth:`request_shutdown`
+        first.
+        """
+        if waittime is None:
+            waittime = max(self.shutdown_wait_time, 0)
+        for i in range(int(waittime/pollinterval)):
```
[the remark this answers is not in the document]
kevin-bates: I think we should still have the cast in place since the result could be a non-integer if `pollinterval` were to be adjusted.

jupyter_client/manager.py
```diff
@@ -488,6 +510,215 @@ def is_alive(self):
...
+                # On Unix, we may get an ESRCH error if the process has already
+                # terminated. Ignore it.
+                else:
+                    from errno import ESRCH
+                    if e.errno != ESRCH:
+                        raise
+
+            # Wait until the kernel terminates.
+            try:
+                await asyncio.wait_for(gen.maybe_future(self.kernel.wait()), timeout=5.0)
```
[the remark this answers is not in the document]
kevin-bates: Yes - good point.  I think `kernel_async_wait()` would look something like this: https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/subproc/manager.py#L56-L66

Does that look right?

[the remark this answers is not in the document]
kevin-bates: That sure does look a lot like `finish_shutdown()` so I'll look into refactoring these areas a bit.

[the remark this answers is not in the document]
davidbrochart: > Yes - good point. I think `kernel_async_wait()` would look something like this: https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/subproc/manager.py#L56-L66
> 
> Does that look right?

Indeed. They even use `asyncio.subprocess.Process`, which would be ideal, but I'm not sure we can have it in `jupyter_client`.

jupyter_client/manager.py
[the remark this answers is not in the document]
davidbrochart: Good point.

jupyter_client/manager.py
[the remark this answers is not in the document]
kevin-bates: Right - the hierarchy of things gets in the way for that.

Btw, I'm finding I need to indicate that the _async_wait() function is done...
```python
    async def _async_wait(self, pollinterval=0.1):
        # Use busy loop at 100ms intervals, polling until the process is
        # not alive.  If we find the process is no longer alive, complete
        # its cleanup via the blocking wait().  Callers are responsible for
        # issuing calls to wait() using a timeout (see _kill_kernel()).
        while self.is_alive():
            await asyncio.sleep(pollinterval)

        done = asyncio.Future()
        done.set_result(None)
        return done
```
otherwise I get failures stemming from its introduction since `asyncio.wait_for()` requires a future.  This seems correct to me, but I honestly lack experience in this area.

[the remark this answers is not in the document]
davidbrochart: `asyncio.wait_for` accepts an awaitable - a coroutine, a Task or a Future. `_async_wait` doesn't have to return a result:
```python
    async def _async_wait(self, pollinterval=0.1):
        while self.is_alive():
            await asyncio.sleep(pollinterval)
```
But you need to catch the timeout:
```python
try:
    await asyncio.wait_for(self._async_wait(), timeout=5.0)
except asyncio.TimeoutError:
    # Wait timed out, just log warning but continue - not much more we can do.
    self.log.warning("Wait for final termination of kernel timed out - continuing...")
    pass
```
That is what your code was doing already. I realize my comment above was wrong, there is no `await` in the `wait_for` call.

jupyter_client/ioloop/restarter.py
```diff
@@ -9,17 +9,20 @@
 
 import warnings
 
+from tornado import gen
```
davidbrochart: Unused.

jupyter_client/manager.py
```diff
@@ -488,6 +509,221 @@ def is_alive(self):
...
+
+    client_factory = Type(klass='jupyter_client.asynchronous.AsyncKernelClient')
+
+    async def _launch_kernel(self, kernel_cmd, **kw):
+        """actually launch the kernel
+
+        override in a subclass to launch kernel subprocesses differently
+        """
+        res = launch_kernel(kernel_cmd, **kw)
+        return res
```
davidbrochart: There is nothing `async` about this function, it doesn't `await` anything.

jupyter_client/manager.py
```diff
@@ -488,6 +509,221 @@ def is_alive(self):
...
+        `**kw` : optional
+             keyword arguments that are passed down to build the kernel_cmd
+             and launching the kernel (e.g. Popen kwargs).
+        """
+        kernel_cmd, kw = self.pre_start_kernel(**kw)
+
+        # launch the kernel subprocess
+        self.log.debug("Starting kernel (async): %s", kernel_cmd)
+        self.kernel = await self._launch_kernel(kernel_cmd, **kw)
+        self.post_start_kernel(**kw)
```
davidbrochart: Since `_launch_kernel` should not be a coroutine, neither should `start_kernel`.

jupyter_client/manager.py
```diff
@@ -488,6 +509,221 @@ def is_alive(self):
...
+            if hasattr(os, "getpgid") and hasattr(os, "killpg"):
+                try:
+                    pgid = os.getpgid(self.kernel.pid)
+                    os.killpg(pgid, signum)
+                    return
+                except OSError:
+                    pass
+            self.kernel.send_signal(signum)
+        else:
+            raise RuntimeError("Cannot signal kernel. No kernel is running!")
```
davidbrochart: `interrupt_kernel` and `signal_kernel` should not be coroutines, they are not doing anything blocking.

jupyter_client/manager.py
```diff
@@ -488,6 +509,221 @@ def is_alive(self):
...
+        """Kill the running kernel.
+
+        This is a private method, callers should use shutdown_kernel(now=True).
+        """
+        if self.has_kernel:
+            # Signal the kernel to terminate (sends SIGKILL on Unix and calls
+            # TerminateProcess() on Win32).
+            try:
+                if hasattr(signal, 'SIGKILL'):
+                    await self.signal_kernel(signal.SIGKILL)
```
davidbrochart: No need to await here, `signal_kernel` should not be a coroutine.

jupyter_client/ioloop/restarter.py
kevin-bates: Thanks.

jupyter_client/manager.py
kevin-bates: Thanks for the review David, but I don't agree with this comment (and the others below) regarding the marking of a method with `async` that doesn't `await`.  While it's true that _this implementation_ doesn't await.  Subclasses, especially those managing remote kernels, certainly will, since the times to start, interrupt, signal, etc. take an order of magnitude longer (perhaps more) to complete.  

Because callers won't know they're talking to a kernel manager that is dealing with remote kernels (nor should they), I believe the async markers need to remain so the caller is always performing `await start_kernel()`, for example.  

Similarly, I think `is_alive()` needs to be added back into AsyncKernelManager since `poll()` is an RPC in some remote kernels (and a REST call in others).  And, by extension, I believe AsyncKernelClient needs an async `is_alive()` method since it will call on the kernel manager's `is_alive()` and we'll need to check the class to conditionally apply the `await`.  I have such an incantation working right now, but I wanted to get your opinion before its push.

davidbrochart: You convinced me @kevin-bates! Thanks for putting this work into perspective, I had not thought of the remote kernel use case. It makes more sense now, so please discard my previous comments.

davidbrochart:
Just a general question, any reason we have an `ioloop` directory or should it be under `asynchronous`?

kevin-bates:
I guess I don't understand your question.  `ioloop` contains class definitions for both sync and async classes, just like `manager.py` and `multikernelmanager.py`.  Are you suggesting we move all of these async classes into the `asynchronous` sub-directory?

davidbrochart:
You're right, it doesn't make a lot of sense.
I think this is ready to be merged. Thanks a lot for all the effort you put into this PR!
Comments anybody?

kevin-bates:
@davidbrochart - thank you for your review - it's much appreciated.  I'm really stoked that this has been revived (since there are two other layers above this with dormant async management PRs :smile: ).

davidbrochart:
It may be out of this PR's scope, but how would it play with Jupyter server's kernel manager? Will we need to write a `MappingAsyncKernelManager`, or just rewrite https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/services/kernels/kernelmanager.py so that `MappingKernelManager` inherits from `AsyncMultiKernelManager`?

kevin-bates:
The plan is for juptyer_server to adopt the new [kernel management framework](https://github.com/takluyver/jupyter_kernel_mgmt) via [this PR](https://github.com/jupyter/jupyter_server/pull/112).  This transition is targetting Server's 0.4 release.  Since kernel management in JKM is only async and the library no longer provides a `MultiKernelManager`, Server's `MappingKernelManager` will be straight async.  **What also needs to happen is applying your client async changes to JKM (please!)**.

However, the dual-mode behavior is exactly what I did for Notebook and I'm hoping once jupyter_client has a carrier for this PR, then [Notebook PR 4479](https://github.com/jupyter/notebook/pull/4479) can be similarly revived.  That will then unblock the [Enterprise Gateway PR](https://github.com/jupyter/enterprise_gateway/pull/580), which will set all the right classes to make the full stack async and something gateway users desperately need!

davidbrochart:
Thanks for the feedback, I have a lot to catch up!

kevin-bates:
No worries.  If you're interested in the server efforts, there's a weekly dev meeting on Thursdays: https://jupyter.readthedocs.io/en/latest/community/content-community.html#monthly-meetings

davidbrochart:
@kevin-bates I'm not sure I did the right thing, the commit history shows I am the author of this commit, I don't see you name. Is it because I squashed it?

SylvainCorlay:
> The plan is for juptyer_server to adopt the new kernel management framework via this PR. This transition is targetting Server's 0.4 release. Since kernel management in JKM is only async and the library no longer provides a MultiKernelManager, Server's MappingKernelManager will be straight async. What also needs to happen is applying your client async changes to JKM (please!).

@kevin-bates I think we need to have a plan for jupyter-server before the move to jupyter_kernel_management goes in. Moving jupyterlab and notebook to server, and also dropping jupyter_client all at once seems *very* disruptive to me, and should probably be done in two stages.

We have no feedback yet on jupyter_kernel_management, and I would like us to be cautious.

kevin-bates:
@davidbrochart - Regarding your comment: https://github.com/jupyter/jupyter_client/pull/428#issuecomment-594897533. No worries - this just means you "touched it last"! :smile:

Just kidding. Normally I wouldn't care too much, but this particular effort was significant and has been more than a year coming, so I think this should be corrected.   I don't know _how_ this happened, but I would think that commit could be amended with something like this: 
```
git commit --amend --author="Kevin Bates <kbates4@gmail.com>"
```
cc: @MSeal, @SylvainCorlay  for their git fu.

SylvainCorlay:
@kevin-bates no worries I will fix the git history first thing tomorrow to include your commits.

MSeal:
Thanks for tackling this PR against everyone. It's nice seeing folks helping get these larger changes merged.

If any additional git fu is needed let me know, I can probably help too.

SylvainCorlay:
@kevin-bates you should now be credited for your commits in master!

PR #438, opened 2019-05-14
Backport PR #437: Avoid kernel failures with multiple processes

alexrudy:
**Note**: This might fail, since python versions 3.3 and 3.4 were dropped before this commit – if the tests pass, then backporting might be okay, but if they fail, I'll need some additional advice.


mpacer:
This seems to be reasonable to me! Merging!

PR #441, opened 2019-05-28
Feature: fork kernel

maartenbreddels:
Feature needed for referenced PR to ipykernel + test script


SylvainCorlay:
@maartenbreddels could you please rebase your PR on the top of the latest master?

We reset master to 5.x. The two branches had been diverging too much for no good reason because there was a backward incompatible branch in master that was never really completed.

PR #444, opened 2019-06-01
Update wrapper kernel docs

blink1073:
Clean up outdated Octave reference and add notes about Metakernels.

cc @dsblank


dsblank:
@blink1073 Looks good!

rgbkrk:
Thanks for the doc updates!

Issue #446, opened 2019-06-12
Supporting the Debug Adapter Protocol


JohanMabille:
At QuantStack, we are working on a prototype of debugger for the Jupyter ecosystem. This issue tracks the changes we need to make to get it done; some of them might be integrated independently from the debugger if they are considered as a global improvement.

The idea is to reuse the [Debug Adaper Protocol](https://microsoft.github.io/debug-adapter-protocol/specification) from Microsoft. These messages are wrapped in new Jupyter messages that are sent on the Control channel.

### 1] Changes in the protocol

- [ ] The `Shell` and `Control` channels should accept different messages; AFAIK this is somehow the current implementation, where most of the messages are sent on the Shell, while `shutdown_request` and `interrupt_request` are sent on the Control, however it would be clearer to have it formaly stated in the protocol itself. @jasongrout opened #388 to update the documentation along these lines.

- [ ] Three new messages are added to the protocol: `debug_request`, `debug_reply` and `debug_event`. `debug_request` wraps a request described in the Debug Adapter Protocol and is sent on the Control channel exclusively. It expects a `debug_reply` wrapping a response described in the Debug Adapter Protocol. `debug_event` is sent by the kernel on the `IOPub` channel. Having the messages from the Debug Adapter Protocol wrapped in Jupyter Messages avoids copying the full specification from Microsoft and "polluting" the Jupyter specification with a lot of new messages that might be ambiguous (`terminate`, `configurationDone`, ...)

### 2] Changes in the frontends

Frontends should be able to send messages directly on the Control channel, which is currently not exposed at all:

- [ ] JupyterLab should expose a `sendControlMessage`, as well as specific methods to send `debug_request` messages, handle `debug_reply` and `debug_event` messages in the `@jupyterlab/services` package. Interfaces for messages sent on the Control channel should be added.
- [ ] The notebook server should expose the same methods in `notebook/services/kernels/kernelmanager.py`. Messages whose `channel` attribute is `control` should be rerouted to these methods.
- [ ] The Jupyter server should be updated the same way.
- [ ] The Control channel should be exposed by `jupyter_client` classes that are used in the implementation of the Notebook Services and the Jupyter Server Services. In the current implementation, the frontend sends all the messages on the Shell channel, and some messages such as  `shutdown_request` get a special handling to be sent on the Control channel.
- [ ] Design a UI ;) (for the prototype we will quickly make a UI to test)

### 3] Changes in the backends

[...]

PR #447, opened 2019-06-12
Exposes control channel in public API

JohanMabille:
This PR exposes the control channel in the public API, as required in https://github.com/jupyter/jupyter_client/issues/446. The `KernelClient` has not been updated to connect to the Control socket, it might be interesting to do it later.

EDIT: additional changes might be required in `KernelManager`to avoid having another socket trying to connect on the Control channel, but this is not clear to me for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ __pycache__
 .coverage
 .cache
 absolute.json
+*.swp
 
 # Sphinx documentation
 _build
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ __pycache__
 .coverage
 .cache
 absolute.json
+*.swp
```
rgbkrk: This shouldn't have been added as part of this PR, but it's fine for now. I recommend adding vim swap files to your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) so this doesn't come up for you in future repositories.

JohanMabille: Thanks, to be honest I wasn't aware of this! I can make a PR to revert this change if you want.

rgbkrk: I think it's ok for the moment as it won't hurt anything.

PR #448, opened 2019-06-13
Changed zmqContext to not reuse instances

MSeal:
Followup to https://github.com/jupyter/jupyter_client/pull/438 to provide the same thread / process safety as the kernel manager used by nbconvert. This has a small memory overhead to avoid parent processes from deadlocking when operating with concurrency.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -38,7 +38,7 @@ class KernelManager(ConnectionFileMixin):
     # The PyZMQ Context to use for communication with the kernel.
     context = Instance(zmq.Context)
     def _context_default(self):
-        return zmq.Context.instance()
+        return zmq.Context()
 
     # the class to create with our `client` method
     client_class = DottedObjectName('jupyter_client.blocking.BlockingKernelClient')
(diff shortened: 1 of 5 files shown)


mpacer:
Is there a way to add tests to verify that this behaviour continues (basically a uniqueness check for the accessible zmqContext instances?

MSeal:
I'm not sure a uniqueness check is a useful test. I looked into each of the classes that got touched and found that the KernelManager already has strong test coverage on the calls that use zmq, including parallelization. The MultiKernelManager was missing test coverage however so I added some parallelism sanity tests.

rgbkrk:
This seems right, but since I'm not as well versed in pyzmq I've added @minrk to check it out in case there's anything we're missing.

rgbkrk:
Now that I've had a better chance to look over this, I'm happy to bring it in.

minrk:
Sorry for being slow, but 100% okay with this. This does mean that the zmq context is not shared across threads, which is perhaps atypical. The the only functional result should be that:

1. there will be a couple more C++ io threads running to handle messages (probably doesn't matter), and
2. `inproc` can't be used, but I think there are other reasons why inproc can't be used already, mainly that it's only in-process, so can't talk to out-of-process kernels.

The alternate fix would have been to ensure that `zmq.Context.instance()` is fork-safe, which I honestly thought it was, but it totally isn't. I should have known this, since I wrote it!

PR #454, opened 2019-07-08
Harmonize “status” field

flying-sheep:
It’s only mentioned at the top that kernel_info_reply (as all replies) has that field.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -614,6 +614,11 @@ Message type: ``complete_request``::
 Message type: ``complete_reply``::
 
     content = {
+    # status should be 'ok' unless an exception was raised during the request,
+    # in which case it should be 'error', along with the usual error message content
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks!

CC @MSeal - did the release already go out?

MSeal:
Yes it went out yesterday

flying-sheep:
You can still create a branch for this release and cherry-pick this commit into there. That way, it’d show up in the stable docs.

MSeal:
Is this something we should prioritize a patch release for? It'd be easy to ship one out real quick if it's really needed. I was going to keep an eye out for minor issues post release anyway.

flying-sheep:
This is only docs, it’s completely possible to update them without making a release

MSeal:
Ahh yes, sry I missed that. Yep merging now

PR #456, opened 2019-07-09
call _create_connected_socket to instantiate _control_socket in Kerne…

JohanMabille:
…lManager

This PR fixes a bug introduced with #447; now that `IOLoopKernelManager` exposes a decorator of `connect_control`, this method is wrongly called by the KernelManager when creating its internal control socket: https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L199.

This results in creating a `ZMQStream` instead of a regular `ZMQSocket` as it was the case before, when `connect_control` of `ConnectionFileMixin` was directly called. I'm not familiar with pyzmq but this prevents the `shutdown_request` message to reach the kernel, leading to always send a signal to kill it.

cc @SylvainCorlay @rgbkrk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ __pycache__
 .coverage
 .cache
 absolute.json
-*.swp
 
 # Sphinx documentation
 _build
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -196,7 +196,7 @@ class KernelManager(ConnectionFileMixin):
 
     def _connect_control_socket(self):
         if self._control_socket is None:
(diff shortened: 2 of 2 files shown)


SylvainCorlay:
Thanks @JohanMabille. I will merge when all is green.

SylvainCorlay:
So the repo2docker builds seem to have queued up and they take a lot of time. This may delay the merge by a bit.

SylvainCorlay:
@JohanMabille do you want to open the backport PR to 5.x?

JohanMabille:
Yes

PR #457, opened 2019-07-09
[backport #456 to 5.x] call _create_connected_socket to instantiate _control_socket in KernelManager

JohanMabille:
I had to solve conflicts so the commit is not the same as the one on master. Let me know if this is a problem, and how I can fix it.


SylvainCorlay:
@JohanMabille I would just make the main change in that PR.

PR #460, opened 2019-07-12
Reduce count of warnings

Jamim:
Hello,

Hope you'll find this PR useful.

This PR fixes warnings like these:
 - PytestUnknownMarkWarning: Unknown pytest.mark.parallel - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
 - DeprecationWarning: Session._key_changed is deprecated in traitlets 4.1: use @observe and @unobserve instead.
 - DeprecationWarning: Interpreting naive datetime as local 2013-07-03 16:34:52.249482. Please add timezone info to timestamps.
 - DeprecationWarning: zmq.eventloop.ioloop is deprecated in pyzmq 17. pyzmq now works with default tornado and asyncio eventloops.

Best regards!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ __pycache__
 .coverage
 .cache
 absolute.json
+htmlcov/
 
 # Sphinx documentation
 _build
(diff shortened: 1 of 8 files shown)


minrk:
Excellent, thanks!

Issue #443, opened 2019-05-29
execute_reply has extra fields even when status is error or abort


jasongrout:
In https://jupyter-client.readthedocs.io/en/latest/messaging.html#request-reply, the implication is that if the content status is 'error' or 'abort', the usual fields of the reply won't be available. However, in https://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-results it indicates that execution_count will be available even if the status is 'error'. If this is an exception (I think it is), it should be explicitly called out as such.

takluyver:
ipykernel sets the execution_count field of the reply regardless of the status:
https://github.com/ipython/ipykernel/blob/638bc6ff76eb5c99409427acc2cb5aaaeb1d69a0/ipykernel/ipkernel.py#L303-L321

The notebook JS uses it to set the input prompt number after execution:
https://github.com/jupyter/notebook/blob/70d74d21ac051fbeaa81d4ef4fff9fa759de96d7/notebook/static/notebook/js/codecell.js#L415

jupyter_console also expects it to be there for an error reply:
https://github.com/jupyter/jupyter_console/blob/5755ec42a95b70492813be4a6bb8342015126062/jupyter_console/ptshell.py#L628-L646

So I think we can clarify the documentation that it's meant to be there in execute_reply regardless of the status.

PR #462, opened 2019-07-17
Fix ambiguous specification around execution_reply field

### LINKED ISSUE #129: Doc discrepancy on iopub error message
There is small mistake in messaging.rst:

```
Execution errors
----------------

When an error occurs during code execution

Message type: ``error``::

    content = {
       # Similar content to the execute_reply messages for the 'error' case,
       # except the 'status' field is omitted.
    }
```

This way one would guess that `'execution_count'` is part of the message, as it is with `execute_reply`. 
Error messages miss this field, so it should say `# except the 'status' and 'execution_count' fields are omitted.`

takluyver:
This fixes #129 and #443. The resolutions are up for debate, but these are the choices that I believe will require the least code to change - see the issues for details of what I looked at.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1187,7 +1187,7 @@ Message type: ``error``::
 
     content = {
        # Similar content to the execute_reply messages for the 'error' case,
-       # except the 'status' field is omitted.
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for fixing this!

PR #464, opened 2019-07-29
Document debug_[request|reply] and debug_event

SylvainCorlay:
This is adding documentation for

 - `debug_[request/reply]`, which are new messages on the `control` channel,
 - `debug_event`, which are sent back debugging kernels to the front-end on `IOPub`.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -48,15 +48,15 @@ kernel has dedicated sockets for the following functions:
    each frontend and the kernel.
 
 2. **IOPub**: this socket is the 'broadcast channel' where the kernel publishes all
-   side effects (stdout, stderr, etc.) as well as the requests coming from any
-   client over the shell socket and its own requests on the stdin socket.  There
-   are a number of actions in Python which generate side effects: :func:`print`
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
In this PR, we don't specify yet what `debug_[request|reply]` and `debug_event` messages contain, except that the former are on the `Control` channel (to prevent being queued after execution requests) and the latter on `IOPub`.

The actual protocol is almost certainly always going to be the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/), and this is what we have implemented in the xeus-python kernel and in the (in-progress) work with in JupyterLab extensions.

All of it is functional with the latest versions of `jupyter_client`, `jupyter_server /  notebook`, `xeus-python`. However, we need the (backward-compatible) [PR](https://github.com/jupyterlab/jupyterlab/pull/6704 ) merged in JupyterLab for the front-end extension to work.

I would argue that we should be able to move forward with merging this PR and then the JupyterLab PR. We already have a working implementation that demonstrates that this change can accomodate the Debugger Adapter Protocol. However, this is still not *electing* the Debugger Adapter Protocol *per se* at this point.

PR #467, opened 2019-08-16
CoCalc is no longer affected by the unicode offset issue

williamstein:
Just a simple update to the docs, now that this is fixed.

See https://github.com/sagemathinc/cocalc/issues/1994


stevengj:
As I commented on that issue, I think you tested with the wrong characters.  You need to check the protocol messages that you are sending for non-BMP characters.

williamstein:
Thanks.  I'm going to close this, then make a new PR when this is fixed.  I will also include an actual test case (in the PR here) that a user can try to see whether or not these characters are probably handled.

PR #468, opened 2019-08-17
cocalc now properly works; also add a concrete example of an "astral plane unicode" identifier, so anybody can easily see what happens in the Jupyter client of their choice.

williamstein:


SylvainCorlay:
LGTM!

PR #470, opened 2019-08-26
Limit polling time to 500ms to avoid hanging on shutdown.

impact27:
With spyder, `self.time_to_dead` is 18 seconds. This means that if the code wants to exit, it will have to wait a maximum of 18 seconds while `self.poller.poll(1000 * until_dead)` is processing. This makes smaller polls until the time is reached. Also, it properly processes `ZMQError` on exit.

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -100,32 +100,29 @@ class HBChannel(Thread):
 
         self.poller.register(self.socket, zmq.POLLIN)
 
-    def _poll(self, start_time):
+    def _poll(self, until_dead):
         """poll for heartbeat replies until we reach self.time_to_dead.
 
         Ignores interrupts, and returns the result of poll(), which
         will be an empty list if no messages arrived before the timeout,
         or the event tuple if there is a message to receive.
         """
-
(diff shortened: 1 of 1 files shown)


jupyter_client/channels.py
```diff
@@ -147,7 +144,10 @@ def run(self):
             # either a recv or connect, which cannot be followed by EFSM
             self.socket.send(b'ping')
             request_time = time.time()
-            ready = self._poll(request_time)
+            ready = False
+            while (not ready and not self._exit.is_set()
+                   and self.time_to_dead > (time.time() - request_time)):
+                ready = self._poll(.5)
```
rgbkrk: Is this the only usage of `_poll` in this repo?

impact27: Yes

impact27: Yes it seems so.

SylvainCorlay:
@impact27 could you please rebase your PR on the top of the latest master?

We reset master to 5.x. The two branches had been diverging too much for no good reason because there was a backward incompatible branch in master that was never really completed.

impact27:
replaced by https://github.com/jupyter/jupyter_client/pull/651

PR #472, opened 2019-08-31
Add support for templated env entries

kevin-bates:
I've encountered a number of people (including myself) that would like templated environment variable entries in kernelspecs, so I thought I'd implement it add see if this is something useful to the community.

With this change, kernelspec env entries (or even those clients that set `extra_env` on the `KernelManager` instance (along with `kernel_cmd`)) can reference existing env variables using `${ENV_VAR}` within various `env` entries and expect those substitutions to be performed prior to the kernel's launch.

Here's an example kernelspec env entry:

```json
"env": { 
    "PYTHONPATH": "my_special_python_path:${PYTHONPATH}", 
    "FOO": "BAR", 
    "MY_APP_PATH": "my_bin:${PATH}"
}
```
Following the kernel's invocation, it's environment will include the following variable/values:
```
PYTHONPATH=my_special_python_path:<notebook server's PYTHONPATH>
FOO=BAR
MY_APP_PATH=my_bin:<notebook server's PATH>
```
If the templated env variable is not in the launching environment, the templated value will remain in place - just as it shows in the kernelspec's `kernel.json` file.

Kernelspec administrators should be cognizant of potential to share environments.  However, given that's case currently, this shouldn't introduce any new issues.

### DIFF
--- a/docs/kernels.rst
+++ b/docs/kernels.rst
@@ -152,7 +152,10 @@ JSON serialised dictionary containing the following keys and values:
   the client will default to ``signal`` mode.
 - **env** (optional): A dictionary of environment variables to set for the kernel.
   These will be added to the current environment variables before the kernel is
-  started.
+  started.  Existing environment variables can be referenced using ``${<ENV_VAR>}`` and
+  will be substituted with the corresponding value.  Administrators should note that use
+  of ``${<ENV_VAR>}`` can expose sensitive variables and should use only in controlled
+  circumstances.
 - **metadata** (optional): A dictionary of additional attributes about this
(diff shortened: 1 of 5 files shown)


blink1073:
👍 for this idea, thanks for proposing it!

blink1073:
I think a note and warning to admins [here](https://github.com/jupyter/jupyter_client/blob/master/docs/kernels.rst#kernel-specs) would do the trick.

kevin-bates:
Excellent idea Steven - thank you.  I've placed this PR into WIP status since it might be a couple days before I get back to this.  Doc updates slipped my mind - thanks again.

PR #483, opened 2019-09-30
Secure core import

MSeal:
Bug fixes for windows issues are going into jupyter_core for secure_write -- we can move to depending on that location to pick up fixes.

@kevin-bates We can bump the setup requirement to 4.5.1 once https://github.com/jupyter/jupyter_core/pull/166 merges and releases.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup_args = dict(
     ],
     install_requires = [
         'traitlets',
-        'jupyter_core',
+        'jupyter_core>=4.5.0',
         'pyzmq>=13',
         'python-dateutil>=2.1',
         'tornado>=4.1',
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup_args = dict(
     ],
     install_requires = [
         'traitlets',
-        'jupyter_core>=4.5.0',
+        'jupyter_core>=4.6.0',
         'pyzmq>=13',
         'python-dateutil>=2.1',
         'tornado>=4.1',
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,6 @@ python:
     - 2.7
 sudo: false
 install:
-  - |
-    # install pip 10 on python 3.3
-    # to get requires_python support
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -88,7 +88,7 @@ def run(self):
     ],
     install_requires = [
         'traitlets',
-        'jupyter_core',
+        'jupyter_core>=4.5.0',
```
kevin-bates: Since we're already building a [`4.6.0` jupyter_core release](https://github.com/jupyter/jupyter_core/pull/164) - which is waiting on the final resolution for secure_write,  I think we should just move to 4.6.0 with this change.

kevin-bates:
Now that [core is at 4.6.0](https://pypi.org/project/jupyter-core/), it seems like we should pump out a jupyter_client release that depends on 4.6.0.  Then, I believe many of the Windows issues should be resolved - although there's something fishing about the `pywin32==225` release going on.

cc: @MSeal, @minrk, @rgbkrk

MSeal:
@kevin-bates the pywin32==225 build sounds like exclusively an anaconda build issue where the recipe is doing bad things. I've opened an issue in https://github.com/conda-forge/pywin32-feedstock/issues/18 and linked to the thread in winpy32

MSeal:
Gonna merge and release to get this fixed

MSeal:
5.3.4 released!

vidartf:
Does this mean that the `"pywin32 >=1.0 ; sys_platform == 'win32'"` dependency can be removed, and the conda forge package moved back to noarch?

kevin-bates:
@MSeal - thank you!  That's actually great new about `pywin32==225`.

Since there's a [NB patch release brewin'](https://github.com/jupyter/notebook/pull/4891#issuecomment-539732075), does it make sense to bump the min versions of core and client to 4.6.0 and 5.3.4, respectively?  I'm happy to submit that PR if you agree.

What is the timeframe to get `secure_write` into client/master?

MSeal:
@vidartf Yes, assuming the dependency being arch specific doesn't force the downstream package to do so (don't think it does). If you want to make the PR I can help get that merged.

@kevin-bates +1.

PR #492, opened 2019-10-15
Only cache ports if the cache_ports flag is set to True

martinRenou:
cc. @kevin-bates @SylvainCorlay

### DIFF
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -14,7 +14,7 @@ import zmq
 from traitlets.config.configurable import LoggingConfigurable
 from ipython_genutils.importstring import import_item
 from traitlets import (
-    Instance, Dict, Unicode, Any, DottedObjectName, observe, default
+    Any, Bool, Dict, DottedObjectName, Instance, Unicode, default, observe
 )
 from ipython_genutils.py3compat import unicode_type
 
@@ -78,7 +78,7 @@ class MultiKernelManager(LoggingConfigurable):
         def create_kernel_manager(*args, **kwargs):
             km = kernel_manager_ctor(*args, **kwargs)
 
-            if km.transport == 'tcp':
+            if km.cache_ports:
                 km.shell_port = self._find_available_port(km.ip)
                 km.iopub_port = self._find_available_port(km.ip)
                 km.stdin_port = self._find_available_port(km.ip)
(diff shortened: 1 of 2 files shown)


jupyter_client/multikernelmanager.py
```diff
@@ -56,6 +56,8 @@ class MultiKernelManager(LoggingConfigurable):
         """
     )
 
+    cache_ports = Bool(True)
```
kevin-bates: The trait should be on the `KernelManager` since it's the `KernelManager` that is aware of the kernel it's managing - including its locality (and whether ports should be cached).

jupyter_client/multikernelmanager.py
```diff
@@ -78,7 +80,7 @@ def _create_kernel_manager_factory(self):
         def create_kernel_manager(*args, **kwargs):
             km = kernel_manager_ctor(*args, **kwargs)
 
-            if km.transport == 'tcp':
+            if self.cache_ports and km.transport == 'tcp':
```
kevin-bates: This would then become `if km.cache_ports:` since the default value for `cache_ports` (in the KernelManager) could be `self.transport == 'tcp'`

SylvainCorlay:
> @martinRenou - thank you for adding the flag. Since MultiKernelManager can manage different types (and localities) of kernels, the decision to cache ports for a given KernelManager should, ideally, come from the KernelManager instance.

This sounds like a convoluted design while we are not quite sure why disabling that behavior is required in the first place - and you can still disable it completely.

Would you like to have a quick chat to discuss this viva voce?

kevin-bates:
Sure - I think that would be great.  We can meet here: ibm.webex.com/meet/kevin.bates unless you have something else.

SylvainCorlay:
I am on the room.

martinRenou:
As discussed during the meeting. We'll add the flag in `KernelManager`, in case the `MultiKernelManager` manages remote **and** local kernels (this caching logic only makes sense in the case of a local kernels).

martinRenou:
Would you like to give it a look @kevin-bates ?

SylvainCorlay:
Go!

meeseeksmachine:

kevin-bates:
@martinRenou - I took JC 6.0 for a spin in my EG environment yesterday and had an issue with `cache_ports` during shutdown because the cached ports weren't in [`currently_used_ports`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/multikernelmanager.py#L179-L181) resulting in a `KeyError`.

Clearly, the remote kernels don't need (or want) to cache the ports, and I need to change `cache_ports` to false for these kinds of manager instances.  However, I figured I'd be able to do this via the command line (e.g., `--KernelManager.cache_ports = False`) but now realize that the `cache_ports` traitlet isn't configurable.  Moreover, for subclasses to set `cache_ports = False`, they must implement a constructor (or whatever `__init__()` is called in python) because the cached ports are [captured immediately following the manager's instantiation](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/multikernelmanager.py#L76-L86).  Once I created the constructor, I was able to manage the remote kernels w/o a `KeyError`.

My question is, was it intentional to not have the `cache_ports` traitlet configurable (I.e., `config=True`)?  I'm curious how someone would set it to false if they still wanted `tcp` without having to modify their code, create a new release, etc.?  I think being able to do something like `--XXXKernelManager=False` while still allowing `KernelManager` instances to cache ports would be nice.

I may not fully understand how traitlets work, so wanted to get your take on how users should change `cache_ports`.

cc: @SylvainCorlay

PR #493, opened 2019-10-20
Session.msg_id optimization

tomoboy:
I was benchmarking the ipyparallel code and discovered that the scripts where spending a lot of time acquiring msg_ids. 
The msg_id property of the Session class takes a while to run just to get a unique ID. 

To optimise this I changed the msg_id property function to use a message counter and to just return the session_id + message count. This also return a unique ID and runs much faster. 

Using timeit to compare the existing code with my implementation I got these runtimes for 1000 000 calls to session.msg_id: 

| function             | Runtime in seconds |
|----------------------|--------------------|
| Existing msg_id      | 3.2                |
| Reimplemented msg_id | 0.86               |

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -527,10 +527,12 @@ class Session(Configurable):
         new_session.digest_history.update(self.digest_history)
         return new_session
 
+    message_count = 0
     @property
     def msg_id(self):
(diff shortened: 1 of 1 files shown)


minrk:
Great!

meeseeksmachine:

PR #496, opened 2019-11-09
Warn that discovery.py is deprecated

### LINKED ISSUE #495: Kernel Providers should be deprecated in jupyter client
While looking into how [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels/blob/master/nb_conda_kernels/discovery.py) work I realized that jupyter_client has discovery.py - which exposes the KernelProvider work that is really taking place in @takluyver's repos: [takluyver/jupyter_kernel_mgmt](https://github.com/takluyver/jupyter_kernel_mgmt) and [takluyver/jupyter_protocol](https://github.com/takluyver/jupyter_protocol).  In addition, I see that there are folks using the kernel provider framework in jupyter_client - which includes the `make_manager()` method - which no longer exists in the definitive framework.

The plan is that the [jupyter_server will leverage jupyter_kernel_mgmt's KernelProvider framework](https://github.com/jupyter/jupyter_server/pull/112).  The framework will support asyncio and parameterized launches (aka parameterized kernels), etc.  There are no plans to use this framework from within jupyter_client.  Applications using jupyter_client will need to switch to using juptyer_kernel_mgmt/juptyer_protocol to take advantage of KernelProviders, async management and parameterized launches.  This does not preclude building similar support into jupyter_client's kernel management - although those efforts should not include the KernelProvider architecture.

As a result, we should deprecate jupyter_client's discovery.py, and the classes that it implements, with proper messaging to use jupyter_kernel_mgmt instead.  Fortunately, discovery.py is not currently included in any releases, nor should it be.

kevin-bates:
Fixes #495


SylvainCorlay:
@kevin-bates there was never a release with the kernel discovery, so I don't think we need to deprecate it.

We backported all the PRs that were related to kernel_discovery to 5.x and reset master to 5.x. There is a kernel_discovery branch that corresponds to the previous master.

Hence I am closing this.

Issue #511, opened 2020-02-13
Tagging a release including the async APIs


SylvainCorlay:
Hey, we are blocking on the jupyter_client async API being included in a release (pun unintended).

Is there any objection to us tagging a jupyter_client release in the next couple of days?

cc @minrk @Carreau @jasongrout

MSeal:
If it's against the 5.x branch it's probably good but the master branch is way out of sync if the change is PRd against that. I don't have the context on what's the state / intention with master but releases and code changes for the past year or two have been out of that branch. Perhaps one of the pinged people can assist or help with merging 5.x back into master?

SylvainCorlay:
This change is in the master branch.

Do you mean that there are 5.x PRs that are not backports of PRs of the master branch?

MSeal:
Yes. 5.x is deviated far from master and has critical changes in it that we should not make a release without.

SylvainCorlay:
> Yes. 5.x is deviated far from master and has critical changes in it that we should not make a release without.

This is *very* unfortunate. I think that we should not have done this in the 5.x branch.

We need to forward port these changes to master.

MSeal:
I've asked on a few threads why it is that way but the folks who made that decision / did the work haven't been active in the past year.

There are some non-trivial changes in master that I don't know the impact of, nor if they were tested against existing downstream libraries. Someone likely needs to evaluate a 6.0 release candidate for these things with a prerelease. I've been managing the nbconvert 6 and papermill 2 releases so it's been pretty low on my list of things to dig into.

SylvainCorlay:
OK, thanks. I just saw your comment in the async API PR about this.

I will look into it in the next week or so.

davidbrochart:
I could try and merge 5.x back into master.

davidbrochart:
What would be the best strategy here, rebase 5.x onto master, merge master into 5.x?

davidbrochart:
I merged `5.x` back into `master` in https://github.com/jupyter/jupyter_client/pull/513. Could this be merged, or do you think more tests are needed?

MSeal:
Thanks @davidbrochart for all the effort here. One of us is going to need to dig into what had changed in master and make sure it's compatible to the downstream clients. I can do some more thorough testing and code analysis next week to that effort. A brief scan of all the commits in the last 2 years looks like nothing unreasonable.

SylvainCorlay:
[...]

PR #513, opened 2020-02-18
Forward 5.x branch into master

davidbrochart:


SylvainCorlay:
@davidbrochart this gives a good idea of what ought to be forward ported to master.

SylvainCorlay:
@minrk do you remember why releases were made in the 5.x branch in the beginning.

davidbrochart:
I just removed the "release" commits.

MSeal:
I left a comment in #511 but I think so long as i e of us tests against nbclient, papermill, and notebook servers and does a thorough read of the changes in master but not in 5.x I think we can merge. For the  6.0release we may also want to do python 3.5+ only change like all the other jupyter libraries are doing too

davidbrochart:
I tested `nbclient` with this PR, all tests pass: https://github.com/jupyter/nbclient/pull/25

minrk:
@SylvainCorlay It's 6.x for the new kernel discovery machinery (#261), which is still in draft, I think.

davidbrochart:
I tested `papermill` with this PR, all tests pass: https://github.com/nteract/papermill/pull/475

SylvainCorlay:
> @SylvainCorlay It's 6.x for the new kernel discovery machinery (#261), which is still in draft, I think.

@minrk @mseal @davidbrochart @maartenbreddels since this this is still in draft after more than two years, I propose that we

 - rename `5.x` into `master`.
 - call the current master branch including this PR "kernel discovery"
 - backport all PRs that are only in the current master in 5.x

That will make things clearer for future releases in my opinion.

Thoughts?

MSeal:
@SylvainCorlay I'm fine with that plan, it'll be lower risk as well.

minrk:
I'm okay with that. I think the kernel discovery plan has moved toward the new jupyter-protocol and friends packages anyway, rather than trying to do it in-place in jupyter-client.

davidbrochart:
I have a PR for back-porting `master` to `5.x` (excluding the kernel discovery commits): https://github.com/jupyter/jupyter_client/pull/516

SylvainCorlay:
Closing as we are swapping master with 5.x

PR #518, opened 2020-02-24
remove unused sphinx import in test_kernelapp

bollwyvl:
In working on https://github.com/conda-forge/jupyter_client-feedstock/pull/38 uncovered a new test dependency on `sphinx` which may be spurious... doesn't appear to be used, but figured PR is better than issue, if straightforward.


bollwyvl:
Hooray!

Issue #520, opened 2020-02-25
IMessage.buffers kernel testing, protocol changelog


jamesdbrock:
I want to add support for `IMessage.buffers` to the IHaskell Jupyter kernel. https://github.com/gibiansky/IHaskell/issues/1144

~Once I've implemented this, how can I test it, without writing a special client extension specifically to test this feature?~

~__Question 1__: Is there some part of stock Jupyter or JupyterLab client which uses `IMessage.buffers`, and which will fail if the kernel is not correctly receiving or sending `IMessage.buffers`?~

What version of Jupyter client protocol added `IMessage.buffers`, and what else is supported in that version? Here is the changelog for jupyter-client. https://jupyter-client.readthedocs.io/en/stable/changelog.html

Here is the jupyter-client messaging doc, which mentions changes to the Jupyter client protocol: https://jupyter-client.readthedocs.io/en/stable/messaging.html

__Question 2__: Is there a changelog for the Jupyter client protocol?

Edit: crossed out question 1, because that's a frontend question.

jamesdbrock:
Gitter https://gitter.im/jupyterlab/jupyterlab?at=5e5487899aeef6523218189e

> Jason Grout @jasongrout 11:44
> FYI, the Image ipywidgets widget uses binary message buffers to transfer the image data

> Jason Grout @jasongrout 11:45
> That may be the simplest thing in stock Jupyter that uses binary message buffers
> especially if you already have ipywidgets support 
> If you have generic comm support, you could write a custom comm object that transfers things with binary buffers

> James Brock @jamesdbrock 11:46
> yeah... IHaskell no longer supports ipywidgets, the support bitrotted because we didn't keep up with the protocol changes.

jamesdbrock:
Edit: crossed out question 1, because that's a frontend question.

jasongrout:
The specific `IMessage.buffers` interface sounds like a haskell implementation detail, not terminology from the protocol spec. I assume you are referring to the `buffers` key in the [General Message Format](https://jupyter-client.readthedocs.io/en/stable/messaging.html#general-message-format)?

[...]

PR #525, opened 2020-02-27
add changelog for message spec

minrk:
As requested in #520 

Changelog goes back to 5.0 (inclusive), when we started tracking these things well and third-party implementations really got going.

Marked as draft because I'm not sure this is the right way to do it. The changelog info is entirely redundant with the `version changed/added/deprecated` notes (that's where I got the changes), so it means every time we document a change, we must do it twice.

Since Sphinx is fancy about roles, maybe there's a better *automated* way to display this information twice? Or maybe we should *replace* the local-to-the-message change notes with the changelog? I'm not sure. For me, the version changed notes near the messages that have changed seems preferable, but as brought up in #520 it's not efficient for a Kernel author who implements protocol 5.0 and asks "what do I need to change to support 5.4?" which is currently to search the doc for "New in" and "changed".

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1582,7 +1582,7 @@ Completion:
 - ``clear_output``: ``stdout``, ``stderr``, and ``display`` boolean keys for selective clearing are removed,
   and ``wait`` is added.
   The selective clearing keys are ignored in v4 and the default behavior remains the same,
-  so v4 clear_output messages will be safely handled by a v4.1 frontend.
+  so v4 ``clear_output`` messages will be safely handled by a v4.1 frontend.
 
 
 Notes
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1477,6 +1477,114 @@ handlers should set the parent header and publish status busy / idle,
 just like an execute request.
 
 
+Changelog
+=========
+
+5.5 (draft)
+-----------
+
+- Added ``debug_request/reply`` messages
+- Added ``debug_event`` message
+
+5.4
+---
+
+- Sending a ``shutdown_request`` message on the ``shell`` channel is deprecated.
+  It should be sent on the control channel.
+
+5.3
+---
+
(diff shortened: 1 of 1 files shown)


willingc:
@minrk Should we merge this as is?

blink1073:
I agree with @willingc, I think this can merge as-is for the upcoming 7.0 release

rgbkrk (approved):

docs/messaging.rst
```diff
@@ -1477,6 +1477,114 @@ handlers should set the parent header and publish status busy / idle,
...
+- new ``is_complete_request`` and ``is_complete_reply`` messages
+
+4.1
+---
+
+- ``comm_open/close/msg`` messages added
+- ``clear_output``: ``stdout``, ``stderr``, and ``display`` boolean keys for selective clearing are removed,
+  and ``wait`` is added.
+  The selective clearing keys are ignored in v4 and the default behavior remains the same,
+  so v4 clear_output messages will be safely handled by a v4.1 frontend.
```
davidbrochart, line 1585:
```suggestion
  so v4 ``clear_output`` messages will be safely handled by a v4.1 frontend.
```

docs/messaging.rst
```diff
@@ -1477,6 +1477,114 @@ handlers should set the parent header and publish status busy / idle,
...
+Execution:
+
+- ``user_variables`` is removed from ``execute_request/reply`` because it is redundant with ``user_expressions``
+- ``password`` key added to ``input_request``
+
+Output:
+
+- ``data`` key in stream messages renamed to ``text`` for consistency with the notebook format.
+- ``application/json`` in mimebundles should be unpacked JSON data,
+  not a double-serialized JSON strong.
```
davidbrochart, line 1560:
```suggestion
  not a double-serialized JSON string.
```

willingc (approved):

willingc:
@davidbrochart If this looks good to you after CI, please 🚢

PR #530, opened 2020-03-12
[WIP] Remove async send_signal method, defer to subclass

kevin-bates:
After validating the downstream modules that can use async kernel
management (notebook and enterprise-gateway) it was found that
the async send_signal method was not warranted since it doesn't
return an awaitable and deferring to the subclass's sync version is fine.

I'm fairly certain this is the last of the "touch-ups" for the async kernel management changes.  Thanks for your patience.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -642,9 +642,9 @@ class AsyncKernelManager(KernelManager):
             # TerminateProcess() on Win32).
             try:
                 if hasattr(signal, 'SIGKILL'):
-                    await self.signal_kernel(signal.SIGKILL)
+                    self.signal_kernel(signal.SIGKILL)
                 else:
-                    await self.kernel.kill()
+                    self.kernel.kill()
             except OSError as e:
                 # In Windows, we will get an Access Denied error if the process
                 # has already terminated. Ignore it.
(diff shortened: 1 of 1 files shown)


davidbrochart:
No problem @kevin-bates, iterating is necessary.
However when [I suggested to not implement `signal_kernel` as a coroutine](https://github.com/jupyter/jupyter_client/pull/428#discussion_r386870415), you said rightfully that other implementations (those managing remote kernels) could be async.
Note that even though `signal_kernel` doesn't await on anything in this implementation, it is still a valid coroutine and works as fine as making it a blocking function and not awaiting on it (I mean, it doesn't hurt to leave it async).

kevin-bates:
Thanks for questioning this @davidbrochart.  yes, I remember that comment and forgot to reference it in the PR description.  Having this be async was causing failures in Enterprise Gateway leading to `NoneType can't be used in 'await' expression`.  I'll keep looking and have moved this to WIP in the meantime.

davidbrochart:
Then it probably means you have your own implementation of `signal_kernel` in Enterprise Gateway, which is not async.

kevin-bates:
Damn, you're spot on - found that just prior to reading your response!  I'll likely close this PR, but want to run more tests again.

kevin-bates:
Thank you!  That was indeed the issue.  Closing now.

PR #542, opened 2020-04-09
Add AsyncKernelClient client_class to AsyncKernelManager

davidbrochart:


kevin-bates:
Good catch David!  Thanks.  Do we need to cut a 6.1.3 for this?  If so, let me know how I can help.

davidbrochart:
I'd like to do it myself, so that I don't have to bother @MSeal each time, but I think I don't have the rights to PyPI yet.

Issue #546, opened 2020-05-17
Passing kernel_id argument to MultiKernelManager.start_kernel would raise TypeError.


qzchenwl:
Because `pre_start_kernel` does not remove kernel_id from kwargs.

https://github.com/jupyter/jupyter_client/blob/c25d304313da1376925ec1624ccf31b3730038f8/jupyter_client/multikernelmanager.py#L149-L158

When called with `kernel_id`, the exception:
```
      File "/conda/envs/notebook/lib/python3.6/site-packages/mtjupyter_singleuser/kernelmanager.py", line 35, in start_kernel
        MultiKernelManager.start_kernel(self, kernel_id=kernel_id, env=env, **kwargs)
      File "/conda/envs/notebook/lib/python3.6/site-packages/jupyter_client/multikernelmanager.py", line 158, in start_kernel
        km.start_kernel(**kwargs)
      File "/conda/envs/notebook/lib/python3.6/site-packages/jupyter_client/manager.py", line 305, in start_kernel
        self.kernel = self._launch_kernel(kernel_cmd, **kw)
      File "/conda/envs/notebook/lib/python3.6/site-packages/jupyter_client/manager.py", line 212, in _launch_kernel
        return launch_kernel(kernel_cmd, **kw)
      File "/conda/envs/notebook/lib/python3.6/site-packages/jupyter_client/launcher.py", line 135, in launch_kernel
        proc = Popen(cmd, **kwargs)
    TypeError: __init__() got an unexpected keyword argument 'kernel_id'
```

Version 5.x has no problem, because `kernel_id` is removed from `kwargs`.
https://github.com/jupyter/jupyter_client/blob/c1f8ce8c8fdd8927e69ee53edc264c7cbcdadc73/jupyter_client/multikernelmanager.py#L85-L110

kevin-bates:
@qzchenwl - This looks like an issue - thank you for reporting it.  Did you want to provide a pull request for this?

qzchenwl:
@kevin-bates Yes, I opened a pull request https://github.com/jupyter/jupyter_client/pull/547.

PR #547, opened 2020-05-17
Fix start_kernel error when passing kernel_id. Issue #546.

qzchenwl:


kevin-bates:
This change looks good - thank you.  This was a rookie mistake on my part and I apologize for the inconvenience.  

I think it might be good idea to add two tests (sync,async) for bringing your own kernel_id.  If we update [`_run_lifecycle`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/tests/test_multikernelmanager.py#L34) to something like the following...
```python
    def _run_lifecycle(self, km, test_kid=None):
        if test_kid:
            kid = km.start_kernel(stdout=PIPE, stderr=PIPE, kernel_id=test_kid)
            assert kid == test_kid
        else:
            kid = km.start_kernel(stdout=PIPE, stderr=PIPE)
```
and add 
```python
    def test_tcp_lifecycle_with_kernel_id(self):
        km = self._get_tcp_km()
        self._run_lifecycle(km, test_kid=unicode_type(uuid.uuid4()))
```
then I think we'd be covered moving forward.

qzchenwl:
@kevin-bates Thanks for the advice, I've added the two tests now.

davidbrochart:
LGTM. Thanks a lot @qzchenwl!

jupyter_client/tests/test_multikernelmanager.py
```diff
@@ -2,13 +2,14 @@
...
 import threading
+import uuid
 import multiprocessing as mp
 
 from subprocess import PIPE
 from unittest import TestCase
 from tornado.testing import AsyncTestCase, gen_test
-
 from traitlets.config.loader import Config
+from ipython_genutils.py3compat import unicode_type
```
MSeal, line 12: Since we're python 3 only now I don't think you need to import this and can just use `str`

davidbrochart: Right, and it's also used in `multikernelmanager.py`. It could be good to make the change there also.

MSeal: Going to merge for now, we can do a python 3 compat cleanup later on

PR #548, opened 2020-05-24
[BugFix] [Resource Leak] Gracefully Close ZMQ Context upon kernel shutdown

jalpan-randeri:
As part of kernel manager instance creation, we create new context of ZMQ.
This kernel manager are responsible for managing lifecycle of kernel process.
At the end of kernel process lifecycle we are just simply killing kernel process and not closing ZMQ context. This leaked ZMQ contexts holds sockets and end up exhausting system resources

This PR fixes this by gracefully shutting down ZMQ context during kernel shutdown process

### Tests
- Enhanced existing tests to track zmq context checks

### Manual Tests Steps:
1. start jupyter server
2. run kernel_lifecycle.sh script 3 times
3. record socket usage

#### kernel_lifecycle.sh
```
kernel_id=$(curl --silent --show-error  -X POST 'localhost:8888/api/kernels' | jq .id | tr -d '"')
echo $kernel_id
read -n 2 -p 'Stop?'
echo 'stopping '
curl -X DELETE "localhost:8888/api/kernels/$kernel_id"
```

#### Track socket usage using following cmd
```
macos:   lsof -p <pid of jupyter> | grep unix | wc -l
linux:   lsof -p <pid of jupyter> | grep event | wc -l
```

### Local Macbook Test Runs:
After 3 executions of kernel_lifecycle script, socket usage
- without fix
```
$ lsof -p 17207 | grep unix | wc -l 
      21
```

- with fix
```
$ lsof -p 16076 | grep unix | wc -l 
       3
```

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -605,8 +605,6 @@ class AsyncKernelManager(KernelManager):
             # most 1s, checking every 0.1s.
             await self.finish_shutdown()
 
-        self.cleanup(connection_file=not restart)
-
         from . import __version__
         from distutils.version import LooseVersion
 
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -28,6 +28,7 @@ from .connect import ConnectionFileMixin
 from .managerabc import (
     KernelManagerABC
 )
+from deprecated import deprecated
 
 
 class KernelManager(ConnectionFileMixin):
@@ -348,6 +349,18 @@ class KernelManager(ConnectionFileMixin):
         self._close_control_socket()
         self.session.parent = None
 
+    @deprecated(version='6.1.3', reason="please use cleanup with restart function")
+    def cleanup(self, connection_file=True):
+        """Clean up resources when the kernel is shut down"""
+        warnings.warn("cleanup with connection_file is deprecated, "
+                      "use cleanup with restart to perform kernels cleanup.")
+        if connection_file:
+            self.cleanup_connection_file()
+
+        self.cleanup_ipc_files()
+        self._close_control_socket()
+        self.session.parent = None
+
     def shutdown_kernel(self, now=False, restart=False):
         """Attempts to stop the kernel process cleanly.
 
@@ -378,7 +391,10 @@ class KernelManager(ConnectionFileMixin):
             # most 1s, checking every 0.1s.
             self.finish_shutdown()
 
-        self.cleanup(restart=restart)
+        self.cleanup(connection_file=not restart)
+        if not restart:
+            # shutdown ZMQ context as we no longer using this kernel
+            self.context.destroy(linger=100)
 
     def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
@@ -593,7 +609,11 @@ class AsyncKernelManager(KernelManager):
             # most 1s, checking every 0.1s.
             await self.finish_shutdown()
 
-        self.cleanup(restart=restart)
+        self.cleanup(connection_file=not restart)
+
+        if not restart:
+            # shutdown ZMQ context as we no longer using this kernel
+            self.context.destroy(linger=100)
 
     async def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -36,9 +36,12 @@ class KernelManager(ConnectionFileMixin):
     This version starts kernels with Popen.
     """
 
+    _created_context = Bool(False)
+
     # The PyZMQ Context to use for communication with the kernel.
     context = Instance(zmq.Context)
     def _context_default(self):
+        self._created_context = True
         return zmq.Context()
 
     # the class to create with our `client` method
@@ -339,12 +342,14 @@ class KernelManager(ConnectionFileMixin):
 
     def cleanup_resources(self, restart=False):
         """Clean up resources when the kernel is shut down"""
+        if not restart:
+            self.cleanup_connection_file()
+
         self.cleanup_ipc_files()
         self._close_control_socket()
         self.session.parent = None
 
-        if not restart:
-            self.cleanup_connection_file()
+        if self._created_context and not restart:
             self.context.destroy(linger=100)
 
     def cleanup(self, connection_file=True):
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -341,12 +341,14 @@ class KernelManager(ConnectionFileMixin):
         """Clean up resources when the kernel is shut down"""
         if not restart:
             self.cleanup_connection_file()
-            # shutdown ZMQ context as we no longer using this kernel
-            self.context.destroy(linger=100)
 
         self.cleanup_ipc_files()
         self._close_control_socket()
         self.session.parent = None
+
+        # shutdown ZMQ context as we no longer using this kernel
+        if not restart:
+            self.context.destroy(linger=100)
 
     def cleanup(self, connection_file=True):
         """Clean up resources when the kernel is shut down"""
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -28,7 +28,6 @@ from .connect import ConnectionFileMixin
 from .managerabc import (
     KernelManagerABC
 )
-from deprecated import deprecated
 
 
 class KernelManager(ConnectionFileMixin):
@@ -338,7 +337,7 @@ class KernelManager(ConnectionFileMixin):
                 self.log.debug("Kernel is taking too long to finish, killing")
                 self._kill_kernel()
 
-    def cleanup(self, restart=False):
+    def cleanup_resources(self, restart=False):
         """Clean up resources when the kernel is shut down"""
         if not restart:
             self.cleanup_connection_file()
@@ -349,17 +348,11 @@ class KernelManager(ConnectionFileMixin):
         self._close_control_socket()
         self.session.parent = None
 
-    @deprecated(version='6.1.3', reason="please use cleanup with restart function")
     def cleanup(self, connection_file=True):
         """Clean up resources when the kernel is shut down"""
-        warnings.warn("cleanup with connection_file is deprecated, "
-                      "use cleanup with restart to perform kernels cleanup.")
-        if connection_file:
-            self.cleanup_connection_file()
-
-        self.cleanup_ipc_files()
-        self._close_control_socket()
-        self.session.parent = None
+        warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
+                      DeprecationWarning)
+        self.cleanup_resources(restart=not connection_file)
 
     def shutdown_kernel(self, now=False, restart=False):
         """Attempts to stop the kernel process cleanly.
@@ -391,10 +384,13 @@ class KernelManager(ConnectionFileMixin):
             # most 1s, checking every 0.1s.
             self.finish_shutdown()
 
-        self.cleanup(connection_file=not restart)
-        if not restart:
-            # shutdown ZMQ context as we no longer using this kernel
-            self.context.destroy(linger=100)
+        from . import __version__
+        from distutils.version import LooseVersion
+
+        if LooseVersion(__version__) < LooseVersion('6.2'):
+            self.cleanup(connection_file=not restart)
+        else:
+            self.cleanup_resources(restart=restart)
 
     def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
@@ -611,9 +607,13 @@ class AsyncKernelManager(KernelManager):
 
         self.cleanup(connection_file=not restart)
 
-        if not restart:
-            # shutdown ZMQ context as we no longer using this kernel
-            self.context.destroy(linger=100)
+        from . import __version__
+        from distutils.version import LooseVersion
+
+        if LooseVersion(__version__) < LooseVersion('6.2'):
+            self.cleanup(connection_file=not restart)
+        else:
+            self.cleanup_resources(restart=restart)
 
     async def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -339,15 +339,12 @@ class KernelManager(ConnectionFileMixin):
 
     def cleanup_resources(self, restart=False):
         """Clean up resources when the kernel is shut down"""
-        if not restart:
-            self.cleanup_connection_file()
-
         self.cleanup_ipc_files()
         self._close_control_socket()
         self.session.parent = None
 
-        # shutdown ZMQ context as we no longer using this kernel
         if not restart:
+            self.cleanup_connection_file()
             self.context.destroy(linger=100)
 
     def cleanup(self, connection_file=True):
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -355,7 +355,7 @@ class KernelManager(ConnectionFileMixin):
     def cleanup(self, connection_file=True):
         """Clean up resources when the kernel is shut down"""
         warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
-                      DeprecationWarning)
+                      FutureWarning)
         self.cleanup_resources(restart=not connection_file)
 
     def shutdown_kernel(self, now=False, restart=False):
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -377,6 +377,9 @@ class KernelManager(ConnectionFileMixin):
             self.finish_shutdown()
 
         self.cleanup(connection_file=not restart)
+        # shutdown ZMQ context as we no longer using this kernel
+        if not restart:
+            self.context.destroy(linger=100)
 
     def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
@@ -592,6 +595,9 @@ class AsyncKernelManager(KernelManager):
             await self.finish_shutdown()
 
         self.cleanup(connection_file=not restart)
+        # shutdown ZMQ context as we no longer using this kernel
+        if not restart:
+            self.context.destroy(linger=100)
 
     async def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -337,10 +337,12 @@ class KernelManager(ConnectionFileMixin):
                 self.log.debug("Kernel is taking too long to finish, killing")
                 self._kill_kernel()
 
-    def cleanup(self, connection_file=True):
+    def cleanup(self, restart=False):
         """Clean up resources when the kernel is shut down"""
-        if connection_file:
+        if not restart:
             self.cleanup_connection_file()
+            # shutdown ZMQ context as we no longer using this kernel
+            self.context.destroy(linger=100)
 
         self.cleanup_ipc_files()
         self._close_control_socket()
@@ -376,10 +378,7 @@ class KernelManager(ConnectionFileMixin):
             # most 1s, checking every 0.1s.
             self.finish_shutdown()
 
-        self.cleanup(connection_file=not restart)
-        # shutdown ZMQ context as we no longer using this kernel
-        if not restart:
-            self.context.destroy(linger=100)
+        self.cleanup(restart=restart)
 
     def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
@@ -594,10 +593,7 @@ class AsyncKernelManager(KernelManager):
             # most 1s, checking every 0.1s.
             await self.finish_shutdown()
 
-        self.cleanup(connection_file=not restart)
-        # shutdown ZMQ context as we no longer using this kernel
-        if not restart:
-            self.context.destroy(linger=100)
+        self.cleanup(restart=restart)
 
     async def restart_kernel(self, now=False, newports=False, **kw):
         """Restarts a kernel with the arguments that were used to launch it.
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -76,6 +76,11 @@ class MultiKernelManager(LoggingConfigurable):
         kernel_manager_ctor = import_item(self.kernel_manager_class)
 
         def create_kernel_manager(*args, **kwargs):
+            if self.shared_context:
+                if self.context.closed:
+                    # recreate context if closed
+                    self.context = self._context_default()
+                kwargs.setdefault("context", self.context)
             km = kernel_manager_ctor(*args, **kwargs)
 
             if km.cache_ports:
@@ -104,9 +109,27 @@ class MultiKernelManager(LoggingConfigurable):
 
                 return port
 
+    shared_context = Bool(
+        True,
+        config=True,
+        help="Share a single zmq.Context to talk to all my kernels",
+    )
+
+    _created_context = Bool(False)
+
     context = Instance('zmq.Context')
+
+    @default("context")
     def _context_default(self):
+        self._created_context = True
         return zmq.Context()
+
+    def __del__(self):
+        if self._created_context and self.context and not self.context.closed:
+            if self.log:
+                self.log.debug("Destroying zmq context for %s", self)
+            self.context.destroy()
+        super().__del__()
 
     connection_dir = Unicode('')
 
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -512,5 +512,5 @@ class AsyncMultiKernelManager(MultiKernelManager):
             self.request_shutdown(kid)
         for kid in kids:
             await self.finish_shutdown(kid)
-            self.cleanup(kid)
+            self.cleanup_resources(kid)
             self.remove_kernel(kid)
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -198,7 +198,7 @@ class MultiKernelManager(LoggingConfigurable):
         self.log.info("Kernel shutdown: %s" % kernel_id)
 
     @kernel_method
-    def cleanup(self, kernel_id, connection_file=True):
+    def cleanup(self, kernel_id, restart=False):
         """Clean up a kernel's resources"""
 
     def remove_kernel(self, kernel_id):
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -129,7 +129,12 @@ class MultiKernelManager(LoggingConfigurable):
             if self.log:
                 self.log.debug("Destroying zmq context for %s", self)
             self.context.destroy()
-        super().__del__()
+        try:
+            super_del = super().__del__
+        except AttributeError:
+            pass
+        else:
+            super_del()
 
     connection_dir = Unicode('')
 
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -198,7 +198,11 @@ class MultiKernelManager(LoggingConfigurable):
         self.log.info("Kernel shutdown: %s" % kernel_id)
 
     @kernel_method
-    def cleanup(self, kernel_id, restart=False):
+    def cleanup(self, kernel_id, connection_file=True):
+        """Clean up a kernel's resources"""
+
+    @kernel_method
+    def cleanup_resources(self, kernel_id, restart=False):
         """Clean up a kernel's resources"""
 
     def remove_kernel(self, kernel_id):
--- a/jupyter_client/kernelapp.py
+++ b/jupyter_client/kernelapp.py
@@ -77,7 +77,7 @@ class KernelApp(JupyterApp):
             self.setup_signals()
             self.loop.start()
         finally:
-            self.km.cleanup()
+            self.km.cleanup_resources()
 
 
 main = KernelApp.launch_instance
--- a/jupyter_client/tests/test_kernelmanager.py
+++ b/jupyter_client/tests/test_kernelmanager.py
@@ -197,6 +197,28 @@ class TestKernelManager(TestCase):
 
         self._env_test_body(kc)
 
+    def test_cleanup_context(self):
+        km = KernelManager()
+        self.assertIsNotNone(km.context)
+
+        km.cleanup_resources(restart=False)
+
+        self.assertTrue(km.context.closed)
+
+    def test_no_cleanup_shared_context(self):
+        """kernel manager does not terminate shared context"""
+        import zmq
+        ctx = zmq.Context()
+        km = KernelManager(context=ctx)
+        self.assertEquals(km.context, ctx)
+        self.assertIsNotNone(km.context)
+
+        km.cleanup_resources(restart=False)
+        self.assertFalse(km.context.closed)
+        self.assertFalse(ctx.closed)
+
+        ctx.term()
+
 
 class TestParallel:
 
--- a/jupyter_client/tests/test_kernelmanager.py
+++ b/jupyter_client/tests/test_kernelmanager.py
@@ -67,6 +67,7 @@ class TestKernelManager(TestCase):
         km.interrupt_kernel()
         self.assertTrue(isinstance(km, KernelManager))
         km.shutdown_kernel(now=True)
+        self.assertTrue(km.context.closed)
 
     def test_tcp_lifecycle(self):
         km = self._get_tcp_km()
@@ -135,6 +136,7 @@ class TestKernelManager(TestCase):
 
         self.assertTrue(km.is_alive())
         self.assertTrue(kc.is_alive())
+        self.assertFalse(km.context.closed)
 
     def _env_test_body(self, kc):
 
@@ -157,6 +159,7 @@ class TestKernelManager(TestCase):
 
         self.assertTrue(km.is_alive())
         self.assertTrue(kc.is_alive())
+        self.assertFalse(km.context.closed)
 
         self._env_test_body(kc)
 
@@ -190,6 +193,7 @@ class TestKernelManager(TestCase):
 
         self.assertTrue(km.is_alive())
         self.assertTrue(kc.is_alive())
+        self.assertFalse(km.context.closed)
 
         self._env_test_body(kc)
 
@@ -307,6 +311,7 @@ class TestParallel:
         execute('check')
 
         km.shutdown_kernel()
+        assert km.context.closed
 
 
 class TestAsyncKernelManager(AsyncTestCase):
@@ -351,6 +356,7 @@ class TestAsyncKernelManager(AsyncTestCase):
         self.assertTrue(isinstance(km, AsyncKernelManager))
         await km.shutdown_kernel(now=True)
         self.assertFalse(await km.is_alive())
+        self.assertTrue(km.context.closed)
 
     @gen_test
     async def test_tcp_lifecycle(self):
@@ -417,6 +423,7 @@ class TestAsyncKernelManager(AsyncTestCase):
         finally:
             await km.shutdown_kernel(now=True)
             kc.stop_channels()
+            self.assertTrue(km.context.closed)
 
     @gen_test(timeout=10.0)
     async def test_start_new_async_kernel(self):
@@ -431,3 +438,4 @@ class TestAsyncKernelManager(AsyncTestCase):
         finally:
             await km.shutdown_kernel(now=True)
             kc.stop_channels()
+            self.assertTrue(km.context.closed)
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,6 @@ setup_args = dict(
         'pyzmq>=13',
         'python-dateutil>=2.1',
         'tornado>=4.1',
-        'deprecated',
     ],
     python_requires  = '>=3.5',
     extras_require   = {
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ setup_args = dict(
         'pyzmq>=13',
         'python-dateutil>=2.1',
         'tornado>=4.1',
+        'deprecated',
     ],
     python_requires  = '>=3.5',
     extras_require   = {


jalpan-randeri:
#https://github.com/jupyter/jupyter_server/issues/234

SylvainCorlay:
Ping @JohanMabille.

blink1073:
Well done, @jalpan-randeri!  I'll buy you a beverage of choice if we meet in person.

kevin-bates:
Outstanding @jalpan-randeri - thank you!  I ditto @blink1073's offer!  This appears to resolve the leaks we're seeing in EG: https://github.com/jupyter/enterprise_gateway/issues/762#issuecomment-633245738

jupyter_client/manager.py
```diff
@@ -377,6 +377,9 @@ def shutdown_kernel(self, now=False, restart=False):
             self.finish_shutdown()
 
         self.cleanup(connection_file=not restart)
+        # shutdown ZMQ context as we no longer using this kernel
+        if not restart:
+            self.context.destroy(linger=100)
```
davidbrochart, line 382: Since it's a resource cleanup thing, maybe it should be inside `cleanup()`? Maybe we should change it to `cleanup(restart=False)` and have `connection_file = not restart` inside?

jupyter_client/manager.py
```diff
@@ -592,6 +595,9 @@ async def shutdown_kernel(self, now=False, restart=False):
             await self.finish_shutdown()
 
         self.cleanup(connection_file=not restart)
+        # shutdown ZMQ context as we no longer using this kernel
+        if not restart:
+            self.context.destroy(linger=100)
```
davidbrochart, line 600: Same here.

jupyter_client/manager.py
kevin-bates: This seems like a good idea.  If moved, I think we'd need to preserve the default value (e.g., `cleanup(shutdown=True)`). Even though it's associated with a different name, the semantics would be preserved and subclasses wouldn't require updates.

MSeal: +1

jalpan-randeri: Sure.

jupyter_client/manager.py
jalpan-randeri: Done

kevin-bates:
The recent commit makes sense for where we want to be and `restart=False` is common, but I think we need a b/c change for existing users of `KerneManager` subclasses that need this awesome fix.  Any chance we could add the original commit (or compatible) into a `6.1.4` release?  Then, I guess we'd need `6.2.0` for the parameter change.

jalpan-randeri:
@kevin-bates  Sure, will you guide me on how to achieve this?  
I found a tag for `6.1.4`, but i am not sure how to cherry-pick original commit to this release.
Alternatively, I was thinking reusing `connection_file` parameter to cleanup ZMQ context, but then it made logic convoluted

jupyter_client/manager.py
```diff
@@ -337,10 +337,12 @@ def finish_shutdown(self, waittime=None, pollinterval=0.1):
...
                 self._kill_kernel()
 
-    def cleanup(self, connection_file=True):
+    def cleanup(self, restart=False):
         """Clean up resources when the kernel is shut down"""
-        if connection_file:
+        if not restart:
             self.cleanup_connection_file()
+            # shutdown ZMQ context as we no longer using this kernel
+            self.context.destroy(linger=100)
```
minrk, line 345: It is not safe to assume that the context is owned by this KernelManager. It can (and likely should be) shared across all kernel managers in a given process, so destroying it unconditionally is probably not ideal.

minrk:
I think this issue might be resolved by a smaller change, switching the default context:

```python
    def _context_default(self):
        return zmq.Context()
```

to default to the shared global context (more idiomatic use of zmq):

```python
    def _context_default(self):
        return zmq.Context.instance()
```

(this default generator occurs in a few places)

If this change still leaks FDs, that means there are *sockets* being left open that should be closed, not just Contexts. These should be identified and closed explicitly in stop_channels. `destroy` is a way to ensure there are no sockets still open, even if we lost track of them, but in the long run it's probably better to not lose track of those sockets in the first place.

kevin-bates:
Thanks @minrk.  

>  I think the ideal fix is to share the global Context rather than requiring separate Contexts per kernel manager, which this PR does. I thought we were already doing that, but apparently not.

You're correct.  You had added a global context 9 years ago, that was switched just over a year ago.  The change to a local context first took place in #437 to address multiprocessing aspects of kernels.  It seems that reverting to a global context would jeopardize those efforts (there were 3 additional PRs related to this, although those appear to be mostly version-management related, back-ports, etc)?

FWIW, I've reverted my EG env back to using a global ZMQ context and do not see the leaks occurring (implying we haven't lost track of the sockets). :+1:

MSeal:
@minrk ! Glad to see you on more threads again :)

Yeah as @kevin-bates pointed out we definitely don't want to go back to globals as they break all concurrency in higher abstractions, causing deadlocks against the global state. So it looks like we need to dig into zmq context cleanup more to figure out why we're not cleaning the FDs up under ideal exit conditions?

MSeal:
Or did I misunderstand the latest comments? Is it that we still see FD leaks with this PR or that this PR resolves them similarly to switching back to a global context?

jupyter_client/manager.py
MSeal: I thought we had switched the contexts to be 1:1 with managers because they're not threadsafe. But it's been a while since we made changes around that.

kevin-bates:
We no longer see the leaks with this PR or if we were to switch back to the global context as Min suggested.

My primary concern is compatibility with existing installations.  After thinking further I think we need at least two releases - a backport to 5.3.x and a current release.  I'd prefer we have a compatible fix in 6.1 as well, then have the incompatible fix in, I guess, 6.2?

jalpan-randeri:
Looking at system level, without this fix we are leaking only ZMQbg/Reaper sockets, these seems to be coming from ZMQ context only. We create 5 sockets during kernel startup and we shutdown these 5 sockets at shutdown, Left behinds are only `ZMQbg/Reaper`one, shutting down zmq.context cleans them up. 

The ask was to make ZMQ.context global, which will lead to race conditions, we have exclusive 1:1 zmq context per kernel manager. 

This PR fixes it by destroying zmq context at kernel shutdown. I tested with running my script concurrently and did not observe any failures 

I am just stuck at making change backward compatible.

MSeal:
I don't think we need to backport to 5.x unless we really want to (and shouldn't block this PR if we do) since Python 2 is fully unsupported now. The 6.x release has been fairly stable relative with only some carryover of 5.x issues that weren't resolved like this one. Do you have some code @kevin-bates that still relies on Python 2 for EG that would need this fix?

What's the incompatible fix vs compatible fix referring to?

I think this would be a 6.2 release since there's some minor contract changes and a cleanup change.

Are we good with a merge for now and follow up with release strategy?

kevin-bates:
I agree with moving forward and merging now, but I believe we need at least one backport.

> Do you have some code @kevin-bates that still relies on Python 2 for EG that would need this fix?

We have users that are running EG relative to 5.3.4.  I don't know if they're running Python 2 but when taking master to check this change out, they ran into another 6.x change (port caching) that breaks them and they shouldn't have to take a new EG release to have their leaks fixed.  Enterprise users are slow-moving, so would really prefer an option for them via 5.3.5.

> What's the incompatible fix vs compatible fix referring to?

I may not be understanding python bindings correctly, but when moved into `cleanup()` it replaces the positional parameter and flips its default value from `True` to `False`.  This is incompatible with subclasses since they will be [calling the superclass relative to the old parameter value (that defaults to True)](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/kernels/remotemanager.py#L431-L441) - not to mention I think there would be a positional argument mismatch.   As a result, applications that subclass `KernelManager` and override `cleanup()` will break.

I believe the "compatible" change needs to either be the original commit (that lies outside `cleanup()`) or placed in `cleanup()`, but with the current parameter (and its default) retained.  

There may be other applications besides EG affected by this and those applications shouldn't require their own patch releases to resolve their leaks, IMO.

FWIW, I just pulled the candidate fix w/o changing EG and this manifests itself as a TypeError due to the positional argument change.
```
[E 200526 12:47:55 web:1670] Uncaught exception DELETE /api/kernels/99367f67-c6f5-4029-8bbf-c9ad1d82afa1 (9.160.104.100)
    HTTPServerRequest(protocol='http', host='yarn-eg-node-1.fyre.ibm.com:8888', method='DELETE', uri='/api/kernels/99367f67-c6f5-4029-8bbf-c9ad1d82afa1', version='HTTP/1.1', remote_ip='9.160.104.100')
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
        value = future.result()
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 326, in wrapper
        yielded = next(result)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/handlers.py", line 66, in delete
        yield maybe_future(km.shutdown_kernel(kernel_id))
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 302, in shutdown_kernel
        return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/multikernelmanager.py", line 183, in shutdown_kernel
        km.shutdown_kernel(now=now, restart=restart)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py", line 381, in shutdown_kernel
        self.cleanup(restart=restart)
    TypeError: cleanup() got an unexpected keyword argument 'restart'
```

MSeal:
Ahh I see.

> /opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/multikernelmanager.py", line 183, in shutdown_kernel
        km.shutdown_kernel(now=now, restart=restart)

That just looks like the shutdown_kernel call was missed in the PR's changes: https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/multikernelmanager.py#L183

We can try making the interface more backwards compatible before releasing. Should we just accept `connection_file` as an argument mapped to `not restart` with a deprecation warning printed that this will be removed in... 7.0? Should make it completely compatible and clear to users what to change to in the future.

MSeal:
> Enterprise users are slow-moving, so would really prefer an option for them via 5.3.5.

Alright, if you think this change is important for those users we can make a back-port for this after we get a 6.x release.

kevin-bates:
Thank you Matthew - this change is critical for EG users since EG represents a long-running server who's sole purpose in life is to create and shutdown kernels and leaks like these add up quickly.  

Regarding the change itself, `shutdown_kernel()` is fine.  It's all about the signature change on `cleanup()` that's the compatibility issue.  I agree that leaving `connection_file=True` in place with a deprecation warning is the best way forward.  I'm sorry that's the case but it's the correct thing to do.

MSeal:
@jalpan-randeri Would you mind adding the backwards comparability for `connection_file` with a warning to the PR? Or would you prefer we tackle that after merge?

jalpan-randeri:
i am confused on what changes needs to be done. 

There are two suggestion

1. make ZMQ context global 
>[downside] race conditions, global shared state in multithreaded environment

2. upgrade cleanup method contract
> [downside] backward incompatible 

We don't want to do both of these changes.


---
#### This leaves to following changes needs to be done:
1. move zmq.context cleanup to "shutdown_kernel' method
2. add backward compatibility warning

@kevin-bates  @MSeal 
I do not know how to do 2nd.  does this mean adding logging warning message?
could you guys give me some pointer here?

MSeal:
> make ZMQ context global
> [downside] race conditions, global shared state in multithreaded environment

I think this isn't a valid option for us at this point unless @minrk has more to dig in there.

Here's an example of how to add a deprecation warning: https://www.lesinskis.com/python_deprecation_tutorial.html

I would leave reset=False, but also add `connection_file=None` and if it's not None then emit the warning change set `reset` accordingly.

`def cleanup(reset=False, connection_file=None):`

jalpan-randeri:
Sweet. thank you for pointer. updating PR please wait for new commits. give me sometime

MSeal:
No rush, ty for being so helpful with the longer PR feedback. We really appreciate the contribution!

kevin-bates:
Yes @jalpan-randeri, your patience is appreciated - thank you.

@MSeal - I think you meant to say `restart` rather than `reset`.

Unfortunately, that signature, nor any change in signature will be incompatible because the override method arguments are fixed at `def cleanup(self, connection_file=True)`.

The same traceback that was provided above still occurs.

MSeal:
Yes I mean restart instead of reset, thanks. Where's the override example you're referring to @kevin-bates ?

kevin-bates:
It was provided in a previous comment, but here it is directly: https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/services/kernels/remotemanager.py#L431-L441

MSeal:
I see, sorry I just wanted to make sure I wasn't about to respond about the wrong thing.

We can swap the order of the arguments in cleanup here to be compatible `def cleanup(self, connection_file=None, restart=False)`. Likely we want this to be keyword assigned instead of arg order anyways in the future.

kevin-bates:
That still won't work.  The derived signature is unchanged, so _any_ additional argument will be unexpected.

The _call_ from `shutdown_kernel()` cannot reference the new argument.  Since it must reflect whether the shutdown is actually on behalf of restart, that must be conveyed via `connection_file` and only `connection_file`.

Here's the latest attempt...
```
[E 200526 16:50:29 web:1670] Uncaught exception DELETE /api/kernels/47c9662e-5b47-4693-ab70-142d4eaf0197 (9.160.104.100)
    HTTPServerRequest(protocol='http', host='yarn-eg-node-1.fyre.ibm.com:8888', method='DELETE', uri='/api/kernels/47c9662e-5b47-4693-ab70-142d4eaf0197', version='HTTP/1.1', remote_ip='9.160.104.100')
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
        value = future.result()
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/gen.py", line 326, in wrapper
        yielded = next(result)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/handlers.py", line 66, in delete
        yield maybe_future(km.shutdown_kernel(kernel_id))
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 302, in shutdown_kernel
        return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/multikernelmanager.py", line 183, in shutdown_kernel
        km.shutdown_kernel(now=now, restart=restart)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py", line 381, in shutdown_kernel
        self.cleanup(restart=restart)
    TypeError: cleanup() got an unexpected keyword argument 'restart'
```

MSeal:
Ok, well this gets uglier but we can catch calls to cleanup that raise a TypeError and call the old pattern with a deprecation warning from the kernel manager? I was missing how the inheritance is intertwined.

jalpan-randeri:
how about this?
---------------------

```
# new method must be used from now on
def cleanup_resources(self, restart=False):
            if not restart:
            self.cleanup_connection_file()
            self.context.destroy(linger=100)
        self.cleanup_ipc_files()
        self._close_control_socket()
        self.session.parent = None


# old method deprecated, retained for b/c
@deprecated(version='6.1.3', reason="please use cleanup with restart function")
def cleanup(self, connection_file=True):
        warnings.warn("cleanup with connection_file is deprecated, "
                      "use cleanup with restart to perform kernels cleanup.")
        if connection_file:
            self.cleanup_connection_file()

        self.cleanup_ipc_files()
        self._close_control_socket()
        self.session.parent = None


 def shutdown_kernel(self, now=False, restart=False):
        self.stop_restarter()

        if now:
            self._kill_kernel()
        else:
            self.finish_shutdown()

        from . import __version__
        from distutils.version import LooseVersion

        if LooseVersion(__version__) < LooseVersion('6.1.4'):
            self.cleanup(connection_file=not restart)
            if not restart:
                self.context.destroy(linger=100)
        else:
            self.cleanup_resources(restart=restart)
```

kevin-bates:
This seems like it should work.  I'm not sure what `LooseVersion` does, but the comparison would need to be against `6.1.5` (or `6.2`) since `6.1.4` is the _carrier_ for this fix.  I would also just have the old cleanup method just defer to `cleanup_resources(restart=not connection_file)` after issuing the deprecation warning.  Adding the new method makes a lot of sense.

I can try this and get back.  I just tested a fix trapping TypeError as well.

For the 5.3.5 release, I'm hoping we can just use the original commit (outside `cleanup()`) or add to cleanup with only `connection_file=True`.

blink1073:
We use LooseVersion in Jupyterlab, I like that solution.

jalpan-randeri:
@kevin-bates i will wait for your testing, based on for your feedback/findings, i will modify this pr

kevin-bates:
This appears to work.  Here's the code I used (taking the deferral approach).  We'll also need to settle on which version numbers are used.
```python
    # new method must be used from now on
    def cleanup_resources(self, restart=False):
        if not restart:
            self.cleanup_connection_file()
            self.context.destroy(linger=100)

        self.cleanup_ipc_files()
        self._close_control_socket()
        self.session.parent = None

    # old method deprecated, retained for b/c
    @deprecated(version='6.1.4', reason="please use cleanup with restart function")
    def cleanup(self, connection_file=True):
        warnings.warn("cleanup with connection_file is deprecated, "
                      "use cleanup_resources with restart to perform kernels cleanup.")
        self.cleanup_resources(restart=not connection_file)
```
and in each shutdown method...
```python
        from . import __version__
        from distutils.version import LooseVersion

        if LooseVersion(__version__) < LooseVersion('6.2'):
            self.cleanup(connection_file=not restart)
        else:
            self.cleanup_resources(restart=restart)
```

The warnings statement came out a bit clunky; seems like it's not happy about the broken string.
```
/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py:354: UserWarning: cleanup with connection_file is deprecated, use cleanup_resources with restart to perform kernels cleanup.
  warnings.warn("cleanup with connection_file is deprecated, "
```

Perhaps we could say...
"Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False)."

Also, this gets issued once per call usage (by default) (so once or twice since restarts should be a different usage).  I'm okay with that, just putting it out for others.

jalpan-randeri:
wow! Sweet, this version is even better.
i got versioning idea from spark codebase,
error message you suggested looks more informative, i will go with this one. 
```
Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).
```

For version selection: i am fine with 6.2 
i got 6.1.4 from _version.py file, i am not exactly sure which file we maintain constants so i can reuse them instead of it inline.

how about i define a new constant named `minimum_recommended_version` into _version.py file?
- https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/_version.py
```
minimum_recommended_version = 6.2  <-- Version to pick 
 if LooseVersion(__version__) < LooseVersion(minimum_recommended_version):
            self.cleanup(connection_file=not restart)
        else:
            self.cleanup_resources(restart=restart)
```

kevin-bates:
I don't have a strong opinion on adding a `minimum_recommended_version`.  I just wouldn't want it to require maintenance or get us into situations if, say, it's update was forgotten or was ill-advised.

jupyter_client/manager.py
minrk: contexts are threadsafe, sockets are not. There was an [issue](https://github.com/jupyter/jupyter_client/pull/437) around *multiprocessing* which I believe is also fixed in recent-enough pyzmq (if not, it's a pyzmq bug).

minrk:
Thanks for digging into the history, @kevin-bates! We should definitely still *allow* a shared context, so the termination of the context should be contingent on whether the KernelManager created it itself. If we stick with this change (I don't want to reintroduce the issues splitting contexts fixed, though there is probably another fix we should make and we should eventually move back to the global context by default as the right idiomatic pattern), recording that we created the context is probably good enough:

```python

_created_context = Bool(False)
@default('context')
def _context_default(self):
    self._created_context = True
    return zmq.Context()
...
def cleanup_resources(self, restart=False):
    ...
    if not restart and self._created_context:
        self.context.destroy(linger=100)     
```

That way, we're safe if a KernelManager is passed a shared context with `KernelManager(context=my_context)`.

jupyter_client/manager.py
minrk: I also just noticed the `_close_control_socket` below - destroying the context should occur after closing sockets.

minrk:
I also want to clarify:

> The ask was to make ZMQ.context global, which will lead to race conditions, we have exclusive 1:1 zmq context per kernel manager.

There shouldn't be race conditions on the context. `zmq.Context.instance()` is a threadsafe class method for creating a global shared context, and the standard way to create a shared Context for a given application in multithreaded scenarios. Methods on zmq Context objects are also themselves threadsafe (Sockets are not). Creating several Contexts in one process should typically only happen in fairly unusual scenarios and/or to work around pyzmq bugs (as found in https://github.com/jupyter/jupyter_client/pull/437, which related to fork-safety of pyzmq, I believe).

kevin-bates:
> We should definitely still allow a shared context, so the termination of the context should be contingent on whether the KernelManager created it itself. If we stick with this change [parenthesized comment remove], recording that we created the context is probably good enough:
[code removed]
That way, we're safe if a KernelManager is passed a shared context with KernelManager(context=my_context).

Doesn't this imply that the "context-creating" KernelManager instance must be the last to leave the party?  I.e., won't its early termination, and destruction of the zmq context, side-affect other kernel managers using that shared context?  I have no idea if that pattern is already mandated.  I suppose it wasn't applicable prior to #437.

MSeal:
> Creating several Contexts in one process should typically only happen in fairly unusual scenarios and/or to work around pyzmq bugs (as found in #437, which related to fork-safety of pyzmq, I believe).

There were significant issues therein that weren't resolved without a 1:1 relationship with fork safety in pyzmq. I did extensive testing on threaded and multi-process patterns such that I'd feel very uncomfortable reverting the behavior back to a shared context without significant testing, evaluation, and digging back into the concurrent behavior in pyzmq. @minrk could we maybe move that concern into a separate issue and track it? The code has been relatively stable for a number of concurrent applications and I think this PR would resolve the memory leak without worsening shared context concerns, since we already don't share by default for the past year.

jupyter_client/manager.py
MSeal: > I also just noticed the _close_control_socket below - destroying the context should occur after closing sockets.

That is a good catch

minrk:
@MSeal 

>  I'd feel very uncomfortable reverting the behavior back to a shared context without significant testing, evaluation, and digging back into the concurrent behavior in pyzmq

Agree 100%. I fully support scoping this PR to improving cleanup of KernelManager-created Contexts and saving the switch back to shared context as a future task. I just want to make sure that we don't make a change that *breaks* applications that pass a context to the KernelManager, which this currently does. Checking if it's using the self-created default is enough for that.

>  @minrk could we maybe move that concern into a separate issue and track it?

Opened #549 


@kevin-bates 

> Doesn't this imply that the "context-creating" KernelManager instance must be the last to leave the party?

It does. I think that's okay, though—if you aren't passing the context to the kernel manager, it makes sense to me to consider the kernel manager the "owner" of the lifecycle of that context. It's technically possible to run afoul of this, but it would require some slightly odd code that I don't think we need to support, such as:

```python
km = KernelManager()
# start using kernel
ctx = km.context

s = ctx.socket(...)
# stop using kernel
km.cleanup()

# keep using context and/or sockets that you got from the KM, but after you've destroyed it?
s.send(...???)
```

What *does* make sense, however, is to create a shared context and pass it to the KM, and then keep using it after the KM is gone:

```python
ctx = zmq.Context()
km = KernelManager(context=ctx)
...
km.cleanup()
s = ctx.socket(...) # should still work!
```

In fact, we should have a test in this PR to make sure that works.

MSeal:
Thanks @minrk, this sounds like a reasonable plan!

jupyter_client/manager.py
jalpan-randeri: Aha, nice catch. I missed this during refactoring. Let me fix this in new revision

jalpan-randeri:
@minrk @MSeal  @kevin-bates @blink1073 

Need a pointer on zmq.context creation, 

Currently on master branch, a new instance of context is always created when we create new instance of kernel manager.
```
    # The PyZMQ Context to use for communication with the kernel.
    context = Instance(zmq.Context)
    def _context_default(self):
        return zmq.Context()
```


This means we are not sharing context today, so following code segments are not possible without changing KernelManager context creation logic.

Here is the test i ran. 

```
class TestKernelManager(TestCase):

    def setUp(self):
        self.test_context = zmq.Context()

    def _get_tcp_km(self):
        c = Config()
        km = KernelManager(config=c, context=self.test_context)
        return km

    def _run_lifecycle(self, km):
        km.start_kernel(stdout=PIPE, stderr=PIPE)
        self.assertTrue(km.is_alive())
        km.restart_kernel(now=True)
        self.assertTrue(km.is_alive())
        km.interrupt_kernel()
        self.assertTrue(isinstance(km, KernelManager))
        km.shutdown_kernel(now=True)
        print(km.context)
        print(self.test_context)
        self.assertTrue(km.context.closed)
        self.assertTrue(self.test_context.closed). <<-- Failed at this line, because context is still open

```

From debugging this test, i can assert that this line
`km = KernelManager(config=c, context=self.test_context)`
has no impact. Kernel Manager always creates a new instance of zmq.Context. 

Questions:
- If we change context creation logic, how do we ensure that we do not break other parts of system?

- it seems, from above test, following code segment is invalid, as injected ctx is simply replaced with new instance so it always works with current code change. Why do we need to add test for this ?
```
ctx = zmq.Context()
km = KernelManager(context=ctx)
...
km.cleanup()
s = ctx.socket(...) # should still work!
```

jalpan-randeri:
jenkins retest please

minrk:
Here are two tests that should pass:


```python
    def test_cleanup_context(self):
		"""kernel manager cleans up its own context"""
        km = KernelManager()
        assert km.context is not None
        km.cleanup_resources(restart=False)
        assert km.context.closed

    def test_no_cleanup_shared_context(self):
        """kernel manager does not terminate shared context"""
        ctx = zmq.Context()
        km = KernelManager(context=ctx)
        assert km.context is ctx
        km.cleanup_resources(restart=False)
        assert not ctx.closed # <-- fails right now due to unconditional context.destroy()
        ctx.term()
```

And this change should get both tests passing:


```python
_created_context = Bool(False)

def _context_default(self):
    self._created_context = True
    return zmq.Context()
...
def cleanup_resources(self, restart=False):
    ...
    if self._created_context and not restart:
        self.context.destroy(linger=100)    
```

kevin-bates:
Thanks @jalpan-randeri and @MSeal.  I intend to build the branch and check that EG is handling the changes as well.  I won't be able to get to that until after 14:30 Pacific though.

jalpan-randeri:
@kevin-bates Sure, i noticed in my personal branch enterprise gateway and jupyter-server are still using older cleanup method, shall i migrate them to new cleanup_resource methods?

kevin-bates:
I agree with that relative to EG since it subclasses KernelManager and you have already submitted a [PR](https://github.com/jupyter/enterprise_gateway/pull/815) (thank you).  That said, there are existing installations that shouldn't have to update to fix their leaks - which is why this b/c route has been taken.

I'm surprised to hear about jupyter_server since its essentially a copy of the notebook server (for now) and neither subclass KernelManager or call `cleanup()` to the best of my knowledge.  Could you please provide a reference to the KM cleanup method?

kevin-bates:
I see a couple of things.
1. I never see the [deprecation warning](https://docs.python.org/3/library/warnings.html#warning-categories).  I suspect this is because the `DeprecationWarning` class is ignored by default unless it's triggered in `__main__`.
If I set the class to `FutureWarning`, I see the following:
```
/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py:358: FutureWarning: Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).
  FutureWarning)
```
so I think we should update the warning category/class.

2. I'm sorry to report that **JupyterLab** produces the following exception just after responding to the shutdown (DELETE)...
```
[I 15:19:37.899 LabApp] Kernel shutdown: bef25493-bea9-4499-9249-209adfa3c3e8
/opt/anaconda2/envs/py3/lib/python3.6/site-packages/jupyter_client/manager.py:358: FutureWarning: Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).
  FutureWarning)
[D 15:19:38.407 LabApp] 204 DELETE /api/sessions/de984a89-b0c3-4383-8616-bc70075ecc27?1590790777844 (9.211.148.27) 520.81ms
[D 15:19:38.485 LabApp] Websocket closed bef25493-bea9-4499-9249-209adfa3c3e8:07a65e3d-bfbb-4680-93b4-dbda27bd6e81
[E 15:19:38.486 LabApp] Uncaught exception, closing connection.
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/iostream.py", line 752, in wrapper
        return callback(*args)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/stack_context.py", line 300, in null_wrapper
        return fn(*args, **kwargs)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/websocket.py", line 488, in on_connection_close
        self.on_close()
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/handlers.py", line 455, in on_close
        stream.on_recv(None)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 187, in on_recv
        self._drop_io_state(zmq.POLLIN)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 532, in _drop_io_state
        self._update_handler(self._state)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 539, in _update_handler
        if state & self.socket.events:
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/sugar/attrsettr.py", line 48, in __getattr__
        return self._get_attr_opt(upper_key, opt)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/sugar/attrsettr.py", line 52, in _get_attr_opt
        return self.get(opt)
      File "zmq/backend/cython/socket.pyx", line 501, in zmq.backend.cython.socket.Socket.get
      File "zmq/backend/cython/socket.pyx", line 262, in zmq.backend.cython.socket._getsockopt
      File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
    zmq.error.ZMQError: Socket operation on non-socket
[E 15:19:38.500 LabApp] Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x7f18df7a7510>)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/ioloop.py", line 758, in _run_callback
        ret = callback()
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/stack_context.py", line 300, in null_wrapper
        return fn(*args, **kwargs)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/iostream.py", line 752, in wrapper
        return callback(*args)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/stack_context.py", line 300, in null_wrapper
        return fn(*args, **kwargs)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/tornado/websocket.py", line 488, in on_connection_close
        self.on_close()
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/notebook/services/kernels/handlers.py", line 455, in on_close
        stream.on_recv(None)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 187, in on_recv
        self._drop_io_state(zmq.POLLIN)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 532, in _drop_io_state
        self._update_handler(self._state)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 539, in _update_handler
        if state & self.socket.events:
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/sugar/attrsettr.py", line 48, in __getattr__
        return self._get_attr_opt(upper_key, opt)
      File "/opt/anaconda2/envs/py3/lib/python3.6/site-packages/zmq/sugar/attrsettr.py", line 52, in _get_attr_opt
        return self.get(opt)
      File "zmq/backend/cython/socket.pyx", line 501, in zmq.backend.cython.socket.Socket.get
      File "zmq/backend/cython/socket.pyx", line 262, in zmq.backend.cython.socket._getsockopt
      File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
    zmq.error.ZMQError: Socket operation on non-socket
```
If I change to notebook (setting the endpoint to `/tree`), I don't see this.  I think this is because Lab tends to deal with WebSockets differently than Notebook.  If I start as notebook (/tree), then switch to lab (/lab), I start seeing the issue again.

I can't recall using Lab in the earlier incantations of this PR, but I suspect it's a timing thing relative to the ZMQ Context's destruction.

I'm using Notebook 6.0.3 under the covers.  This also happens when hitting EG and running Lab remotely because, in that case, the gateway layer in the remote notebook proxies the kernel WebSocket interactions.

jalpan-randeri:
sorry. my bad. it was `jupyter_client` pkg only i was reading master branch of jupyter_client thinking it is was jupyter_server project.

Let me try out with lab, i was using jupyter_classic whole time

jalpan-randeri:
I noticed, jupyter_lab is trying to establish new connection via websocket, after sending  shutdown message. But i am unable to trace, which part of code is performing this, since I am not familiar with typescript language. from network trace via chrome seems like some sort of reconnect logic to handle transient network outage

kevin-bates:
@blink1073 - do you know why Lab's behavior is different from Notebook's following a kernel's shutdown?  

FWIW, Lab also varies from Notebook following restart requests.  In that case, it is **not** reconnecting with the WebSocket.

blink1073:
@jasongrout made some intentional changes in that regard, he has a better background on the subject.

kevin-bates:
Given the lab issue, I'm not sure how we should proceed.  Here are the sequences surrounding a notebook's shutdown from Lab (via /lab URL) and Notebook (via /tree URL) for the same notebook server instance:

**Lab:**
```
[D 16:30:15.774 LabApp] Clearing buffer for 2ace824a-0e4c-467b-b3e3-4a45355e262d
[I 16:30:15.774 LabApp] Kernel shutdown: 2ace824a-0e4c-467b-b3e3-4a45355e262d
[D 16:30:15.981 LabApp] 204 DELETE /api/sessions/cee32fc7-0d27-4c42-a5eb-374deb07f82e?1591227015770 (::1) 208.03ms
[D 16:30:15.984 LabApp] Websocket closed 2ace824a-0e4c-467b-b3e3-4a45355e262d:5faa0546-4782-4a62-9590-611db13c398e
```

**Notebook:**
```
[D 16:31:22.815 LabApp] Websocket closed 34569310-53ca-4867-b088-237cdfc04f74:0d334be6b2f2451fb2f528936d9a0626
[I 16:31:22.816 LabApp] Starting buffering for 34569310-53ca-4867-b088-237cdfc04f74:0d334be6b2f2451fb2f528936d9a0626
[D 16:31:22.816 LabApp] Clearing buffer for 34569310-53ca-4867-b088-237cdfc04f74
[D 16:31:22.819 LabApp] Clearing buffer for 34569310-53ca-4867-b088-237cdfc04f74
[I 16:31:22.820 LabApp] Kernel shutdown: 34569310-53ca-4867-b088-237cdfc04f74
[D 16:31:23.022 LabApp] 204 DELETE /api/sessions/9ec0f60c-41b1-400e-9826-14f44b79ffda (::1) 204.15ms
```
Note: juptyer_client == 6.1.2 was used for this, so the traceback from https://github.com/jupyter/jupyter_client/pull/548#issuecomment-636221159 is not produced.

MSeal:
Yeah I was waiting for @jasongrout to chime in if he could as well.

It's looking like maybe we'll have to have the object go into a 'off' mode that catches the zmq failure when it knows it's already cleaned up?

minrk:
@kevin-bates I believe that traceback is precisely what would be caused by calling `ctx.destroy()` instead of `ctx.term()`, because it's closing open sockets via a low-level method and zmqstream doesn't handle the underlying socket being closed by someone else. This should probably be considered a bug in `ZMQStream.closed()` which assumes that sockets can only be closed by calling `stream.close()`, which isn't the case here, but we should probably also be closing the streams explicitly *before* calling this method.

I'll try to dig into what's the right thing to do on that front. On the notebook side, closing websockets prior to shutdown might be the simplest.

kevin-bates:
@MSeal - I think we can forgo having to deal with a backport to 5.x.  My plan for that is to allow for a [configurable option to use a global ZMQ context](https://github.com/jupyter/enterprise_gateway/pull/820) in the older EG release.  Since it's my understanding that the multi-threaded issues with the global context come about more so in lower-level (or more direct) applications (and not so much in the general Notebook applications) and use of the global context also addresses the FD leaks we see in EG, I think it's sufficient to just override the default context method (although not a python best practice) and conditionally create a global context.

I'm not certain if this might be an approach we want to bake into jupyter_client in current versions, but it will definitely alleviate the need for a back-port while addressing critical resource issues in long-running services (like EG).  It would also address the issues we see with Lab using the WebSocket after the kernel's termination.

jasongrout:
Sorry didn't see me being mentioned above until now. I can try to look at the lab behavior next week.

MSeal:
All good @jasongrout. Thanks for offering to look into it.

MSeal:
@kevin-bates Ok, let's discuss the global context and backport options in https://github.com/jupyter/jupyter_client/issues/549 then. If there's sufficient testing done to prove it's forks and thread safe and pyzmq expects the global then I'm fine with helping move it back to that. I agree it feels like a general anti-pattern to move to global state managers, though port management is inherently global to a system and pyzmq has an opinion here so I'll go with the flow and help where I can.

minrk:
The above patch in pyzmq (https://github.com/zeromq/pyzmq/pull/1392) turns the traceback into a warning because zmq sockets are being closed before unregistering them from the event loop:

```
Websocket closed 80473386-6bd6-4c60-8512-d49ad473c802:d8c670ad-c556-4af9-865c-4919eaeb5cec
/Users/minrk/dev/zmq/pyzmq/zmq/eventloop/zmqstream.py:435: UserWarning: Unregistering FD 36 after closing socket. This could result in unregistering handlers for the wrong socket. Please use stream.close() instead of closing the socket directly.
  self.close()
/Users/minrk/dev/zmq/pyzmq/zmq/eventloop/zmqstream.py:435: UserWarning: Unregistering FD 41 after closing socket. This could result in unregistering handlers for the wrong socket. Please use stream.close() instead of closing the socket directly.
  self.close()
/Users/minrk/dev/zmq/pyzmq/zmq/eventloop/zmqstream.py:435: UserWarning: Unregistering FD 44 after closing socket. This could result in unregistering handlers for the wrong socket. Please use stream.close() instead of closing the socket directly.
  self.close()
/Users/minrk/dev/zmq/pyzmq/zmq/eventloop/zmqstream.py:435: UserWarning: Unregistering FD 47 after closing socket. This could result in unregistering handlers for the wrong socket. Please use stream.close() instead of closing the socket directly.
  self.close()
```

These warnings are informative, because under heavy connect/close load, it's possible (not sure how likely) for this circumstance to actually result in closing the wrong FDs.

My hunch is that the difference between lab and classic is that lab shuts down the kernel before closing connections, while classic closes connections first. The notebook server does not enforce that connections are closed before shutting down a kernel, and this isn't even technically invalid.

I'm starting to think that tying implicit context destruction to shutdown might not be the right thing to do. Instead, tying it to `__del__` might have closer to the right lifetime by default. It doesn't feel right to conflate process-lifetime operations (start/shutdown) with connection-lifetime operations (start_channels/context close).

Here's a valid course of action:

1. connect iopub
2. start consuming iopub
3. shutdown kernel
4. continue flushing iopub until we are sure no more messages will come
5. finally cleanup the context

which this PR would *require* passing a user-created context to the kernel manager to do.


With this PR otherwise unchanged, we could get some of both benefits in the notebook server in particular by having `MultiKernelManager`'s default behavior be using an explicit shared context across its children. Not `zmq.Context.instance()`, but the same logic we have here, with `context=self.context` in the child constructors.

minrk:
[This diff](https://github.com/jalpan-randeri/jupyter_client/compare/bugfix-zmq-leaks...minrk:shared-multikernel?expand=1) applies the following changes:

- destroy context in `__del__` instead of `cleanup_resources` (having a cleanup method makes sense to me, but doing this action in shutdown does not)
- share context across children by default in MultiKernelManager

and has no warnings with lab or classic in my tests (because contexts are never terminated)

MSeal:
> destroy context in __del__ instead of cleanup_resources

This won't be guaranteed to be called until after all references are cleaned, and even then it's not guaranteed to run if the process is exiting. So if someone has a local `kc = km.client` snippet in their code the cleanup behavior would stop executing. Typically I've seen with IO cleanup, you explicitly call it when you're done with that IO and make the __del__ method call as well just in case for this reason.

> share context across children by default in MultiKernelManager

That seems like a good idea to add.

minrk:
What's the best way forward? It does make sense to me to have explicit cleanup, not relying on del, but shutdown does not seem like the right place to close a context (it's quite reasonable to close sockets after shutdown, and we see that jupyterlab does exactly that, hence the errors reported above). We have already that 'explicit cleanup' is `manager.context.term()`, but a wrapper method that closes everything we might have opened is a good idea, too. Should we stop calling `cleanup_resources()` in shutdown and make it an explicit call for consumers of the KernelManager API?

So maybe the right thing is to keep this as-is—manager terminates the context on shutdown when the context is not shared, and then ensure for contexts like the notebook server that the context *is* shared (cherry-pick 4cdbf54fffc44c27e5c82fd5ff03ba00b5bb3142 onto this PR)? We'll get implicit cleanup on shutdown in single KernelManager use cases, one context by default in MultiKernel cases (notebook servers, most notably).

MSeal:
> So maybe the right thing is to keep this as-is—manager terminates the context on shutdown when the context is not shared, and then ensure for contexts like the notebook server that the context is shared (cherry-pick 4cdbf54 onto this PR)? We'll get implicit cleanup on shutdown in single KernelManager use cases, one context by default in MultiKernel cases (notebook servers, most notably).

I think that could be a working plan. I think we want to emphasize that client cleanup and if they don't we do a best effort to cleanup for them. That should cover current behavior for lab and classic unless I am mistaken, and we can ask that lab move to a more deliberate cleanup mode over time? Unless I missed anything?

MSeal:
Should one of us apply the commit to this PR and get it so folks can test it out to confirm behavior before merge?

jalpan-randeri:
just to confirm, we want to provide cleanup_resources as explicit call and not do it as part of shutdown method.  something like this commit https://github.com/jupyter/jupyter_client/commit/4cdbf54fffc44c27e5c82fd5ff03ba00b5bb3142

minrk:
@jalpan-randeri yes, please cherry-pick that commit onto your PR. Then I think we can proceed with testing.

@MSeal to be clear, I believe lab is doing nothing wrong - explicitly closing sockets after shutdown in general should be fine, so I don't think we need to ask them to change. If there is a change to make forcing the closure of sockets before shutdown, that would belong in the notebook server, collecting and closing all `WebSocketHandlers` associated with a kernel as part of shutdown.

jalpan-randeri:
cherry-picking commit gave me following error

Not sure if this is expected behavior

```
AttributeError: 'super' object has no attribute '__del__'
Exception ignored in: <function MultiKernelManager.__del__ at 0x10a0d9290>
  File "/jupyter_client/jupyter_client/multikernelmanager.py", line 132, in __del__
    super().__del__()
AttributeError: 'super' object has no attribute '__del__'

```

jupyter_client/multikernelmanager.py
```diff
@@ -104,10 +109,28 @@ def _find_available_port(self, ip):
...
     def _context_default(self):
+        self._created_context = True
         return zmq.Context()
 
+    def __del__(self):
+        if self._created_context and self.context and not self.context.closed:
+            if self.log:
+                self.log.debug("Destroying zmq context for %s", self)
+            self.context.destroy()
+        super().__del__()
```
minrk, line 132: Oops. Need to handle super not having defined del (I thought this should have been on all objects...):


```suggestion
        try:
            super_del = super().__del__
        except AttributeError:
            pass
        else:
            super_del()
```

jalpan-randeri: Done

jupyter_client/manager.py
```diff
@@ -337,15 +340,24 @@ def finish_shutdown(self, waittime=None, pollinterval=0.1):
...
         self._close_control_socket()
         self.session.parent = None
 
+        if self._created_context and not restart:
+            self.context.destroy(linger=100)
+
+    def cleanup(self, connection_file=True):
+        """Clean up resources when the kernel is shut down"""
+        warnings.warn("Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).",
+                      DeprecationWarning)
```
kevin-bates, line 358: As noted in this [previous comment](https://github.com/jupyter/jupyter_client/pull/548#issuecomment-636221159), I'm not seeing any warning messages produced in the server console/logs.  I think we should adjust the class such that it produces at least one message.  (Using `FutureWarning` does just that.)

jalpan-randeri: Done.

minrk:
Looks great to me, thanks everyone!

minrk:
@kevin-bates if you want to do a backport in order to make sure more versions of the notebook server get it, backporting just the shared context for MultiKernelManager should be the smallest way to achieve that.

kevin-bates:
Thanks @minrk.  I'm kinda new to back-porting, but it sounds like I'd cherry-pick 2d5ba4b and 1ce1d97 to a 5.x branch.  Since no branch for back-ports exists, would we just create 5.x from the 5.3.4 tag?  or should that branch be named 5.3.x?

If someone could set up the target branch, I'd be happy to back-port the appropriate commits and help get a 5.3.5 out.

MSeal:
I can help kick off a releases for these things. Shall I do the `6.1.4` release now?

MSeal:
`6.1.5` and `5.3.5` are both released with this change included. I messed up the `6.1.4` upload (had the wrong commits) and deleted that release right away.

kevin-bates:
Thank you Matt!

jalpan-randeri:
wow!. thank you Kevin, Matt, Min RK  😃

PR #554, opened 2020-06-15
Fix exception causes all over the codebase

cool-RR:
This is a continuation of #552 .


cool-RR:
I just made a blog post about this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

blink1073:
@cool-RR thank you for leading this effort.  It will have a huge impact on users over the years.

PR #555, opened 2020-06-25
Backport zmq leaks fix for MultiKernelManager

kevin-bates:
This is a backport of a _portion_ of PR #548.  Specifically, we use a shared ZMQ context (by default) from the _`MultiKernelManager`_ that, when enabled, uses the same context across all `KernelManager` instances.  The portion not backported is the ability to create (and destroy) a ZMQ context per `KernelManager`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ __pycache__
 .coverage
 .cache
 absolute.json
+.idea/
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

MSeal:
Back port is released with `5.3.5`

PR #561, opened 2020-07-13
Refactor kernel manager tests for pytest

kevin-bates:
This converts the kernel manager tests to the pytest framework. `TestParallel` was already using pytest, so this converts the other tests.  

The subclasses added in #560 are now used in more tests as general subclasses and a specific subclass test now exists.

Note: There was a bug uncovered in `TestParallel` which was not setting the transport correctly.  As a result, the `tcp` transport was used for tests that were meant to be tested using the `ipc` transport.  Of those, the parallel thread and process tests fail using `ipc` and have been temporarily skipped in this PR.

### DIFF
--- a/jupyter_client/tests/test_kernelmanager.py
+++ b/jupyter_client/tests/test_kernelmanager.py
@@ -66,8 +66,6 @@ def install_kernel():
         }))
 
 
-## SYNC-related
-
 @pytest.fixture
 def start_kernel():
     km, kc = start_new_kernel(kernel_name='signaltest')
@@ -246,6 +244,8 @@ class TestParallel:
 
     @pytest.mark.timeout(TIMEOUT)
     def test_start_parallel_thread_kernels(self, config, install_kernel):
+        if config.KernelManager.transport == 'ipc':  # FIXME
+            pytest.skip("IPC transport is currently not working for this test!")
         self._run_signaltest_lifecycle(config)
 
         thread = threading.Thread(target=self._run_signaltest_lifecycle, args=(config,))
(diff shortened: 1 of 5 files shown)


jupyter_client/tests/test_kernelmanager.py
```diff
@@ -7,117 +7,174 @@
...
+@async_generator  # This is only necessary while Python 3.5 is support afterwhich both it and yield_() can be removed
+async def start_async_kernel():
+    km, kc = await start_new_async_kernel(kernel_name='signaltest')
+    await yield_((km, kc))
+    kc.stop_channels()
+    await km.shutdown_kernel()
+    assert km.context.closed
+
+
+class TestKernelManager:
```
MSeal, line 129: I don't think the class is necessary btw. You can just define test functions directly.

kevin-bates: True.  I was just keeping the classes around for organization purposes, although I did find it convenient to specify `@pytest.mark.asyncio` on the `TestAsyncKernelManager` declaration for its tests.

Do you want these removed and things flattened out?

MSeal: Nah, I agree it's helping organize here. Was just commenting in case it wasn't known. Let's merge your work :)

MSeal:
Should we kick off a patch release?

kevin-bates:
Thanks for the merge @MSeal.  A patch release sounds like a grand idea - thank you!

MSeal:
Done

PR #568, opened 2020-08-04
Fix launching non python kernels from a windows store installed jupyter

### LINKED ISSUE #567: Windows store installed jupyter will not run any kernel except python kernels.
See these bugs here:
https://bugs.python.org/issue41196
https://github.com/dotnet/interactive/issues/612

I believe the root cause of the problem is that the Windows store python will redirect %APPDATA% to a different folder under the covers while returning the true APPDATA folder as the path.

This can be fixed by using os.path.realpath when passing the connection file to a kernel.

Basically change this line here:
https://github.com/jupyter/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/multikernelmanager.py#L170
To be this instead:

```
        km = self.kernel_manager_factory(connection_file=os.path.realpath(os.path.join(
                    self.connection_dir, "kernel-%s.json" % kernel_id)),
                    parent=self, log=self.log, kernel_name=kernel_name,
                    **constructor_kwargs
        )

```

rchiodo:
Fixes #567 

The root cause of this problem is the path of the connection file that's passed to an external process. 

From within the window's store python (where jupyter_client is running), the path looks like so:
```
C:\Users\rchiodo.REDMOND\AppData\Roaming\jupyter\runtime\kernel-e62fdf2f-ca02-4611-b38d-a363a8cdff10.json
```
However this path doesn't actually exist. It's a redirected path caused by the window's store version of python (see this bug here:  https://bugs.python.org/issue41196)

What the actual path is this:
```
C:\Users\rchiodo.REDMOND\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\Roaming\jupyter\runtime\kernel-e62fdf2f-ca02-4611-b38d-a363a8cdff10.json
```
This change 'fixes' the path to the real path before launching the kernel.

Sorry but not sure how I'd write a test for this other than manually testing it with a Windows store python.

### DIFF
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -11,6 +11,12 @@ from ipython_genutils.encoding import getdefaultencoding
 from ipython_genutils.py3compat import cast_bytes_py2
 from traitlets.log import get_logger
 
+def force_real_path(p):
+    """ Turns any paths into their real version """
+    if (os.path.exists(p)):
+        return os.path.realpath(p)
+    return p
+
 
 def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
                   independent=False, cwd=None, **kw):
@@ -86,6 +92,10 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
         if cwd:
             cwd = cast_bytes_py2(cwd, sys.getfilesystemencoding() or 'ascii')
             kwargs['cwd'] = cwd
+
+        # Windows store python will pass invalid paths for the connection file. 
+        # Turn the connection file into a real path
+        cmd = [ force_real_path(c) for c in cmd]
 
         from .win_interrupt import create_interrupt_event
         # Create a Win32 event for interrupting the kernel
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -11,6 +11,7 @@ from ipython_genutils.encoding import getdefaultencoding
 from ipython_genutils.py3compat import cast_bytes_py2
 from traitlets.log import get_logger
 
+
 def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
                   independent=False, cwd=None, **kw):
     """ Launches a localhost kernel, binding to the specified ports.
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -11,13 +11,6 @@ from ipython_genutils.encoding import getdefaultencoding
 from ipython_genutils.py3compat import cast_bytes_py2
 from traitlets.log import get_logger
 
-def force_real_path(p):
-    """ Turns any paths into their real version """
-    if (os.path.exists(p)):
-        return os.path.realpath(p)
-    return p
-
-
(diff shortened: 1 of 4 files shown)


jupyter_client/launcher.py
```diff
@@ -87,6 +93,10 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
             cwd = cast_bytes_py2(cwd, sys.getfilesystemencoding() or 'ascii')
             kwargs['cwd'] = cwd
 
+        # Windows store python will pass invalid paths for the connection file. 
+        # Turn the connection file into a real path
+        cmd = [ force_real_path(c) for c in cmd]
```
davidbrochart, line 98: Note that `cmd` represents Python code, not all elements of the list are paths. I don't think it is safe to assume that an element that is not a path won't exist as a path by chance.

davidbrochart:
Thanks for reporting the issue and submitting a PR @rchiodo. I agree replacing the path to the connection file with the "real path" can solve the issue, however it might be better to do it in:
https://github.com/jupyter/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/manager.py#L175-L208

rchiodo:
> Thanks for reporting the issue and submitting a PR @rchiodo. I agree replacing the path to the connection file with the "real path" can solve the issue, however it might be better to do it in:
> https://github.com/jupyter/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/manager.py#L175-L208

rchiodo:
> Thanks for reporting the issue and submitting a PR @rchiodo. I agree replacing the path to the connection file with the "real path" can solve the issue, however it might be better to do it in:
> https://github.com/jupyter/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/manager.py#L175-L208

I'm not sure that would work (I can try) but when I did something similar in multikernelmanager.py it broke python kernels (as I believe other spots must have the connection_file with the old path).

I think it would be better to just change the code in the launcher.py to instead verify it's the kernel-<guid>.json file. Then there wouldn't be a problem with possibly changing paths on things that aren't files.

rchiodo:
No your suggested change did work. Thanks. I'll update to that.

rchiodo:
@MSeal I'd be happy to add a test but it would have to install the window's store python (or perhaps mock realpath to force it to something different for a unit test).

rchiodo:
Do you have any suggestions on where to start with say a unit test? That seems like the easier way to verify this.

MSeal:
I would mock realpath to set the path to a fake location and test that the command returned by `format_kernel_cmd` has an expected realpath mocked value present. Should be a fairly small test to cover behavior though you will need to set kernel_cmd, connection_file on the tested object to ensure the function can complete.

There's no existing tests on `format_kernel_cmd` or `pre_start_kernel` which calls it. But the https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/tests/test_kernelmanager.py file has some other examples testing adjacent or parent functions.

PR #571, opened 2020-08-17
Use py3 metaclass syntax

rahulporuri:
This PR replaces the use of `with_metaclass` with the `metaclass=` kwarg - which is the standard way to set metaclass when defining classes on Python 3. The `with_metaclass` imports have been removed from the modules.

Note that this PR doesn't fix any specific issue that I am aware of. This is just a drive-by cleanup.

### DIFF
--- a/jupyter_client/channelsabc.py
+++ b/jupyter_client/channelsabc.py
@@ -5,10 +5,8 @@
 
 import abc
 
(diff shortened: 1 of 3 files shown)


davidbrochart:
LGTM, thanks!

PR #572, opened 2020-08-17
Use python 3 super syntax

rahulporuri:
This PR replaces the use of `super(_class_name_, self)` with `super()` - which is the standard way to access the inheritance tree on Python 3.


davidbrochart:
Thanks a lot!

rahulporuri:
@davidbrochart thanks for the quick review. I don't think I can merge the PR myself because I see "Only those with write access to this repository can merge pull requests.". Not sure what development workflow is jupyter_client/jupyter org follows.

davidbrochart:
Let's wait a bit for other people to review.

PR #573, opened 2020-08-17
Remove more py2/3 compat code

rahulporuri:
- replace use of compat iteritems(dict) with dict.items()
- remove use of no-op cast_bytes_py2 on python 3
- remove use of unicode_type and string_types

FTR, here's the compat module from `ipython_genutils`

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -11,7 +11,6 @@ import warnings
 from dateutil.parser import parse as _dateutil_parse
 from dateutil.tz import tzlocal
 
-from ipython_genutils import py3compat
 next_attr_name = '__next__' # Not sure what downstream library uses this, but left it to be safe
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


jupyter_client/jsonutil.py
```diff
@@ -11,7 +11,6 @@
 from dateutil.parser import parse as _dateutil_parse
 from dateutil.tz import tzlocal
 
-from ipython_genutils import py3compat
```
rahulporuri, line 14: this is drive-by cleanup - unused import

jupyter_client/session.py
```diff
@@ -199,14 +198,14 @@ class Message(object):
 
     def __init__(self, msg_dict):
         dct = self.__dict__
-        for k, v in iteritems(dict(msg_dict)):
+        for k, v in dict(msg_dict).items():
```
rahulporuri, line 201: i'm not sure why `ipython_genutils.py3compat` refers to `iteritems` as `iter(dict.items())` when on python 3 `dict.items()` already returns an iterator!

PR #576, opened 2020-08-25
Silence kill_kernel when no process is present

### LINKED ISSUE #575: Exception in callback Error when closing the kernel
OS: Ubuntu 18.04 

Latest Jupyter notebook

WHen I close a kernel from command prompt using ctrl+c, it closes the kernel and throws and exception. It was not doing it when I had `tensorflow 2.2`, but i switched to `tensorflow 2.0` and it is giving this jupyter error.

```error
Shutdown this notebook server (y/[n])? y
[C 18:45:49.608 NotebookApp] Shutdown confirmed
[I 18:45:49.610 NotebookApp] Shutting down 1 kernel
[I 18:45:50.313 NotebookApp] Kernel shutdown: b62da79b-ca59-43f3-ba21-a021e6330235
[I 18:45:50.314 NotebookApp] Shutting down 0 terminals
[I 18:45:50.319 NotebookApp] KernelRestarter: restarting kernel (1/5), keep random ports
WARNING:root:kernel b62da79b-ca59-43f3-ba21-a021e6330235 restarted
[E 18:45:50.321 NotebookApp] Exception in callback <bound method KernelRestarter.poll of <jupyter_client.ioloop.restarter.IOLoopKernelRestarter object at 0x7fc102298510>>
      File "/home/ahmad/anaconda3/envs/deep/lib/python3.7/site-packages/tornado/ioloop.py", line 907, in _run
        return self.callback()
      File "/home/ahmad/anaconda3/envs/deep/lib/python3.7/site-packages/jupyter_client/restarter.py", line 113, in poll
        self.kernel_manager.restart_kernel(now=True, newports=newports)
      File "/home/ahmad/anaconda3/envs/deep/lib/python3.7/site-packages/jupyter_client/manager.py", line 442, in restart_kernel
        self.shutdown_kernel(now=now, restart=True)
      File "/home/ahmad/anaconda3/envs/deep/lib/python3.7/site-packages/jupyter_client/manager.py", line 383, in shutdown_kernel
        self._kill_kernel()
      File "/home/ahmad/anaconda3/envs/deep/lib/python3.7/site-packages/jupyter_client/manager.py", line 487, in _kill_kernel
        raise RuntimeError("Cannot kill kernel. No kernel is running!")
    RuntimeError: Cannot kill kernel. No kernel is running!
(deep) ahmad@ahmad-HP-Laptop-15-da1xxx:~$ 
```
How to solve this error. Thanks

kevin-bates:
During a kernel's shutdown, a race condition can occur between the kernel manager and the kernel restarter such that the restarter sees the kernel process has terminated, yet it's view of the kernel manager still thinks its managing an instance.  In such cases, the restarter initiates its restart sequence and, when attempting to kill the kernel, finds that there's now no kernel process being managed.  In such conditions, the kernel manager was raising an exception indicating there's no process to kill.  Instead, it should perform the same action as when there _is_ a process to kill, only to have the signaling fail because there is no longer any process - [which is to ignore that condition](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L476-L486).  This change essentially ignores the case when a _final_ kill request (i.e., `SIGKILL` vs. `SIGTERM`) finds that a process is no longer being managed (i.e., `self.has_kernel is None`).

Fixes #575

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -464,7 +464,6 @@ class KernelManager(ConnectionFileMixin):
         This is a private method, callers should use shutdown_kernel(now=True).
         """
         if self.has_kernel:
-
             # Signal the kernel to terminate (sends SIGKILL on Unix and calls
             # TerminateProcess() on Win32).
             try:
@@ -488,8 +487,6 @@ class KernelManager(ConnectionFileMixin):
             # Block until the kernel terminates.
             self.kernel.wait()
             self.kernel = None
-        else:
-            raise RuntimeError("Cannot kill kernel. No kernel is running!")
 
     def interrupt_kernel(self):
         """Interrupts the kernel by sending it a signal.
@@ -714,8 +711,6 @@ class AsyncKernelManager(KernelManager):
                 if self.kernel is not None:
                     self.kernel.wait()
             self.kernel = None
-        else:
-            raise RuntimeError("Cannot kill kernel. No kernel is running!")
 
     async def interrupt_kernel(self):
         """Interrupts the kernel by sending it a signal.


MSeal:
Thanks. Do we need a patch release real quick for this change to get the improvement out?

kevin-bates:
Thank you David and Matt.

> Do we need a patch release real quick for this change to get the improvement out?

Thanks for asking.  Unless there are other changes already in the queue, I don't consider this a "release-triggering" PR.  Looking at the set of unreleased PRs, I think #568 might be the only other "fix", but it strikes me as a bit more urgent than this PR.

Or perhaps @ahmadmustafaanis views the priority differently.  Ahmad, do you feel this warrants an immediate patch release?

MSeal:
That's a good point on on #568. I should release for that change given it's been sitting fixed for a bit -- I can kick off a patch release this afternoon with both.

PR #582, opened 2020-09-22
Add documentation dependencies to setup.py

hrnciar:
Hello,

I have noticed that documentation dependencies are not listed between extras. I am proposing to add them there.


MSeal (approved):
LGTM, thanks!

PR #589, opened 2020-11-13
Allow use ~/ in the kernel's command or it's arguments

svetlyak40wt:
After this patch you'll be able to do this
in a kernel spec:

    {
        "argv": ["~/the-project/jupyter/kernel/run", "-f", "{connection_file}"],
        "display_name": "The Project (dev)",
        "language": "python"
    }

This can be useful when you are installing kernel with `--user` option
into `/home/my-name/.local/share/jupyter/kernels/` and having your jupyter
notebooks in different subdirectories of `~/the-project/` folder.

In this case kernel for each notebook is started in the notebook's
subdirectory and without `~/` in the spec Jupyter is unable to find a
file for running a custom kernel.

### DIFF
--- a/jupyter_client/launcher.py
+++ b/jupyter_client/launcher.py
@@ -128,6 +128,9 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
             env['JPY_PARENT_PID'] = str(os.getpid())
 
     try:
+        # Allow to use ~/ in the command or its arguments
+        cmd = list(map(os.path.expanduser, cmd))
+
         proc = Popen(cmd, **kwargs)
     except Exception as exc:
         msg = (


davidbrochart:
Thanks for the PR!
Does it mean every part of the command is treated as a potential path? Isn't there a chance we change something that was not a path?

svetlyak40wt:
`os.path.expanduser` expands `~` only at the beginning of the string:

```pythonn
In [1]: import os.path

In [2]: os.path.expanduser('~/bin')
Out[2]: '/Users/art/bin'

In [3]: os.path.expanduser('Some~/bin')
Out[3]: 'Some~/bin'

In [4]: os.path.expanduser('Some~/bin~')
Out[4]: 'Some~/bin~'
```

There is only a very small probability that somebody will create a command-line program which wants arguments like `~some` not being path names.

davidbrochart:
Thanks!

svetlyak40wt:
👍

PR #590, opened 2020-11-17
perf: speedup date parsing using ciso8601

maartenbreddels:
This is a cutout of running the profiler on voila, where it handles a single comm msg:
image

As can be seen, the green 'parse_date' is a significant part of the cpu usage.

With this PR, is shrinks this quite a bit:
image

This uses ciso8601 to do the datetime parsing, which is [significantly faster](https://github.com/closeio/ciso8601#benchmark)

Note that is not a requirements, it will only be used when installed. However, 1 test regarding timezones fails (it should fail to convert and it does not, I'm not sure how important that is)

# Running it in a benchmark
## Before:
```
--------------------------------------------------------- benchmark: 1 tests --------------------------------------------------------
Name (time in us)                     Min       Max      Mean   StdDev    Median      IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------
test_deserialize_performance     226.2670  556.7280  272.2735  58.3634  249.3170  19.2865   291;378        3.6728    2251           1
-------------------------------------------------------------------------------------------------------------------------------------
```

## After:
```
-------------------------------------------------------- benchmark: 1 tests -------------------------------------------------------
Name (time in us)                    Min         Max     Mean   StdDev   Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------
test_deserialize_performance     59.9780  4,237.0830  76.7926  60.7056  67.9120  4.8000   190;974       13.0221    5717           1
-----------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,9 @@ jobs:
         pip install --upgrade --upgrade-strategy eager --pre -e .[test] pytest-cov codecov 'coverage<5'
         pip freeze
     - name: Run the tests
-      run: py.test --cov jupyter_client jupyter_client
+      run:  |
+        py.test --cov jupyter_client jupyter_client
+        pip install ciso8601
+        py.test --cov jupyter_client jupyter_client
     - name: Code coverage
       run: codecov
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -6,6 +6,10 @@
 from datetime import datetime
 import re
 import warnings
+try:
+    import ciso8601
+except ImportError:
+    ciso8601 = None
 
 from dateutil.parser import parse as _dateutil_parse
 from dateutil.tz import tzlocal
@@ -50,6 +54,11 @@ def parse_date(s):
     """
     if s is None:
         return s
+    if ciso8601 is not None:
+        try:
+            return _ensure_tzinfo(ciso8601.parse_datetime(s))
+        except ValueError:
+            return s
     m = ISO8601_PAT.match(s)
     if m:
         dt = _dateutil_parse(s)
--- a/jupyter_client/tests/test_session.py
+++ b/jupyter_client/tests/test_session.py
@@ -344,3 +344,21 @@ class TestSession(SessionTestCase):
         s._add_digest(digest)
         assert digest in s.digest_history
         assert digest not in s2.digest_history
+
+
+@pytest.mark.usefixtures('no_copy_threshold', 'benchmark')
+class TestPerformance(SessionTestCase):
+    @pytest.fixture(autouse=True)
+    def _request_benchmark(self, benchmark):
+        self.benchmark = benchmark
+
+    def test_deserialize_performance(self):
+        def run(data):
+            self.session.digest_history = []
+            self.session.deserialize(self.session.feed_identities(data)[1])
+        content = dict(t=ss.utcnow())
+        metadata = dict(t=ss.utcnow())
+        p = self.session.msg('msg')
+        msg = self.session.msg('msg', content=content, metadata=metadata, parent=p['header'])
+        data = self.session.serialize(msg)
+        self.benchmark(run, data)
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup_args = dict(
     ],
     python_requires  = '>=3.5',
     extras_require   = {
-        'test': ['ipykernel', 'ipython', 'mock', 'pytest', 'pytest-asyncio', 'async_generator', 'pytest-timeout'],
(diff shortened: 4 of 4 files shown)


maartenbreddels:
Relying on the regex gives about the same performance, but should let the tests pass:
```
------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------
Name (time in us)                    Min       Max     Mean   StdDev   Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------
test_deserialize_performance     57.6940  597.6290  71.3625  32.5174  61.1040  3.7240   413;661       14.0130    3586           1
---------------------------------------------------------------------------------------------------------------------------------
```

maartenbreddels:
In fact, I think we can assume we know where the date fields are right?

That gives us the following performance:

```
------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------
Name (time in us)                    Min       Max     Mean   StdDev   Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------
test_deserialize_performance     27.2420  201.6740  32.8866  11.3219  29.4945  1.3950   520;798       30.4076    5266           1
---------------------------------------------------------------------------------------------------------------------------------
```


And when using `orjson` instead of `json`, we can get:
```
-------------------------------------------------------- benchmark: 1 tests -------------------------------------------------------
Name (time in us)                    Min         Max     Mean   StdDev   Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------
test_deserialize_performance     21.9820  3,148.4180  35.3785  76.8630  23.5860  5.6985    50;395       28.2658    1939           1
-----------------------------------------------------------------------------------------------------------------------------------
```

This is 10x faster.

martinRenou:
Will you include your changes on `orjson` in this PR? Or a separate one?

Thanks for working on this!

maartenbreddels:
Well, jupyter_client users `from zmq.utils import jsonapi` which seems like a smart idea, but orjson does not support seperator, which is used in that module. And I'm also not sure if orjson supports `ensure_ascii=False, allow_nan=False,`

.github/workflows/main.yml
```diff
@@ -37,6 +37,9 @@ jobs:
         pip install --upgrade --upgrade-strategy eager --pre -e .[test] pytest-cov codecov 'coverage<5'
         pip freeze
     - name: Run the tests
-      run: py.test --cov jupyter_client jupyter_client
+      run:  |
+        py.test --cov jupyter_client jupyter_client
+        pip install ciso8601
```
minrk, line 42: Can this be added to the existing `pip install` , one step above, instead of being part of the test stage?

minrk (commented):
This is great! We did something similar in https://github.com/ipython/ipyparallel/pull/424 where we discovered date parsing is so expensive when there's lots of messages, but we disabled it entirely, since that skips the regex match as well.

What I think would be even better is to remove date parsing entirely, but that's a backward-incompatible change. That said, I think very few things (ipyparallel may even be the only one) that really rely on dates being parsed as part of deserialization, so it should be a major change with relatively minor disruption.

What do you think about skipping `extract_dates` entirely as a major revision change? I can't think of a great mechanism for gracefully deprecating this functionality. Maybe a config option on Session with a deprecation warning?

minrk:
> Well, jupyter_client users from zmq.utils import jsonapi

This is a legacy from when we supported Pythons that didn't even have json in the standard library and there were fiddly issues with str vs bytes. There's no need to keep that, as long as we use something that's producing valid utf8 JSON bytes.

jupyter_client/session.py
```diff
@@ -928,10 +928,14 @@ def deserialize(self, msg_list, content=True, copy=True):
...
         header = self.unpack(msg_list[1])
-        message['header'] = extract_dates(header)
+
+        message['header'] = header
+        message['header']['date'] = parse_date(message['header']['date'])
         message['msg_id'] = header['msg_id']
         message['msg_type'] = header['msg_type']
-        message['parent_header'] = extract_dates(self.unpack(msg_list[2]))
+        message['parent_header'] = self.unpack(msg_list[2])
+        if 'date' in message['parent_header']:
```
minrk, line 937: This will break current stable ipyparallel, which currently relies on the behavior of parsing any date-like strong, but I can deal with that. I never should have made it find and parse any valid date objects!

I suspect that parsing just the date will break exactly as many things as disabling parsing entirely, though. Which is to say: ipyparallel and probably nothing else.

blink1073:
Hi @maartenbreddels!  We're pushing for a 7.0 release, do you want to pick this back up?

PR #592, opened 2020-12-04
retry kernel_info_requests in wait_for_ready

SylvainCorlay:
This changes the logic of `wait_for_ready` to continue sending info request until there is some IOPub message as a response.

### DIFF
--- a/jupyter_client/asynchronous/client.py
+++ b/jupyter_client/asynchronous/client.py
@@ -126,14 +126,21 @@ class AsyncKernelClient(KernelClient):
 
         # Wait for kernel info reply on shell channel
         while True:
+            self.kernel_info()
             try:
                 msg = await self.shell_channel.get_msg(timeout=1)
             except Empty:
                 pass
             else:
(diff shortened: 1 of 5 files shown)


jasongrout:
Can you fill out some reasoning in the description? Thanks!

minrk:
Related to #593. While this doesn't truly solve the issue, it slightly increases the likelihood of iopub being fully connected and subscribed for that first message.

SylvainCorlay:
I was going to describe this a bit more, but I just opened this so that this change is somewhere for reference.

Untitled.ipynb
```diff
@@ -0,0 +1,111 @@
+{
```
jasongrout, line 1: Oops, can you rebase to remove this file from the history?

SylvainCorlay:
This is ready to go @minrk.

I had to change the test to check for parent id of responses because there may be older kernel info replies coming late to the party.

Now looking into `jupyter_server` and `notebook`.

PR #594, opened 2020-12-22
Fix test_session with msgpack-1

mgorny:
Replace the encoding parameter with raw=False to fix test_session
with msgpack-1.0.0+.  The encoding parameter was already deprecated
in msgpack-0.6.2, and raw=False is compatible with that version too.

### DIFF
--- a/jupyter_client/tests/test_session.py
+++ b/jupyter_client/tests/test_session.py
@@ -304,7 +304,7 @@ class TestSession(SessionTestCase):
 
         session = ss.Session(
             pack=msgpack.packb,
-            unpack=lambda buf: msgpack.unpackb(buf, encoding='utf8'),
+            unpack=lambda buf: msgpack.unpackb(buf, raw=False),
         )
         self._datetime_test(session)
 


davidbrochart:
Thanks @mgorny. In the [breaking changes in msgpack 1.0](https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10) I can see:
```
encoding option is removed. You can use raw=True to support old format.
```
Should we use `raw=True`?

mgorny:
That's weird. The old version said:

```
jupyter_client/tests/test_session.py::TestSession::test_datetimes_msgpack
  /var/tmp/jupyter_client/jupyter_client/tests/test_session.py:307: DeprecationWarning: encoding is deprecated, Use raw=False instead.
    unpack=lambda buf: msgpack.unpackb(buf, encoding='utf8'),
```

... and this sounds logical to me -- encoded = non-raw.

mgorny:
Also in the (old) code:

```
        if typ == TYPE_RAW:
            if self._encoding is not None:
                obj = obj.decode(self._encoding, self._unicode_errors)
            elif self._raw:
                obj = bytes(obj)
            else:
                obj = obj.decode('utf_8')
```

so `raw = False` is equivalent to utf-8 encoding.

davidbrochart:
Yes, they also say:
```
`raw=False` by default. It assumes str types are valid UTF-8 string and decode them to Python str (unicode) object.
```
which seems to be what we were expecting previously.
Thanks!

PR #596, opened 2021-01-08
Pin jedi<=0.17.2

davidbrochart:


bollwyvl:
This seems like kind of a bad play... what was the reasoning here?

This _added_ a dependency which seems like not was is desired, but needs to be fixed at the actual import site of a _downstream_, e.g. https://github.com/ipython/ipython/pull/12751.

At any rate, I don't think that version should stay... but this project _could_ add it as a pin in its test requirements.

davidbrochart:
> This seems like kind of a bad play... what was the reasoning here?

See https://github.com/jupyter/jupyter_client/runs/1667921500

> This _added_ a dependency which seems like not was is desired, but needs to be fixed at the actual import site of a _downstream_, e.g. [ipython/ipython#12751](https://github.com/ipython/ipython/pull/12751).
> 
> At any rate, I don't think that version should stay... but this project _could_ add it as a pin in its test requirements.

I agree.

kaxil:
This has caused ripple effects in Apache Airflow as well. Are you planning to release a new version after removing the jedi from requirements? @davidbrochart

davidbrochart:
Sorry for that, I just released `6.1.11` which should fix the issue.

kaxil:
> Sorry for that, I just released `6.1.11` which should fix the issue.

Thanks, appreciate the quick release 🙏

PR #605, opened 2021-01-12
Some removal of IPython genutils.

Carreau:
There was some request on ipython_genutils to drop nose-1 (for
packaging on openSUSE). In many cases I believe ipython_genutils can be
dropped as it was mostly helping with py2/3 compatibility.

This remove the trivial replacement with their equivalent when possible,
including one case where the assigned variable was unused, and ence was
removed.

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -23,9 +23,7 @@ import zmq
 from traitlets.config import LoggingConfigurable
 from .localinterfaces import localhost
 from ipython_genutils.path import filefind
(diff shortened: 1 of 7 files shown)


davidbrochart:
Thanks!

MSeal:
Thanks you -- those little backlog tasks get too little recognition

Carreau:
Yeah, IIRC nose will not be compatible with 3.10 so it's gonna be chaotic
at next release.

PR #612, opened 2021-01-26
Kernel Provisioning - initial implementation

### LINKED ISSUE #608: [PROPOSAL] Kernel Provisioning
This issue introduces a proposal named _Kernel Provisioning_.  Its intent is to enable the ability for third-parties to _provision_ the kernel's runtime environment _within the current framework_ of jupyter_client's kernel discovery and lifecycle management.

## Problem
The jupyter_client package currently provides a kernel manager class ([`KernelManager`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L33)) to control the lifecycle of the kernel process.  Lifecycle-action methods supported from a kernel manager include _start_kernel_, _shutdown_kernel_, _interrupt_kernel_, _restart_kernel_ , and _is_alive_.  All of these methods interact with the kernel process - which is a [`Popen`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) subprocess - to monitor and control its lifecycle.  For example, 
- [`start_kernel`](https://github.com/jupyter/jupyter_client/blob/6.1.11/jupyter_client/manager.py#L313) creates the `Popen` instance and stores that instance in the kernel manager's `kernel` attribute.  
- [`shutdown_kernel`](https://github.com/jupyter/jupyter_client/blob/6.1.11/jupyter_client/manager.py#L387) is implemented to leverage `Popen`'s `kill()` and `terminate()` methods (depending on urgency).  
- [`interrupt_kernel`](https://github.com/jupyter/jupyter_client/blob/6.1.11/jupyter_client/manager.py#L499) calls `Popen`'s `send_signal()` method (or sends a message if message-based interrupts are configured).   
- While [`is_alive`](https://github.com/jupyter/jupyter_client/blob/6.1.11/jupyter_client/manager.py#L536) is based on `Popen`'s `poll()` method.  
- For completeness, [`restart_kernel`](https://github.com/jupyter/jupyter_client/blob/6.1.11/jupyter_client/manager.py#L445) is a combination of `shtudown_kernel` and `start_kernel`.

Today, applications that wish to launch kernels beyond those of a local `Popen` process (for example, into resource-managed clusters or leverage container-based environments) must instead implement their own `KernelManager` _subclass_.  This introduces a number of issues:
1. `KernelManager` is an _application-level_ class.  That is, functionality related to the application - across all kernels - are implemented via the kernel manager.  Applications such as Notebook _extend_ this class to allow for activity monitoring functionality, for example.
2. Applications (e.g., Notebook, NBClient, etc) enable the ability to "bring your own" kernel manager.  Because `KernelManager` is an application-level class, such kernel manager implementations must be a subclass of `KernelManager` **and** are kernel-_specification_ agnostic.  That is, the same kernel manager instance must manage the lifecycles of Python, R, C++ kernels, as well as kernels launched into resource-managed clusters - which is not possible via a `Popen` subprocess instance.  However, support for the latter types of kernels requires interactions with more than just the kernel process. For example, kernel locations must be _discovered_ within the resource-managed cluster using the resources manager's API and terminated in a similar manner - allowing the resource manager to release resources, update scheduling, etc (examples of such resource managers are Hadoop Yarn or Kubernetes).  As a result, a _single_ kernel manager cannot address the needs of the various configurations in which users want their kernels to operate.
3. Support for highly demanded features such as _parameterized kernels_ cannot be sustainably implemented because 
    a) a given kernel manager instance cannot know about what parameters apply to all kernels and 
    b) a majority of kernel _parameters_ affect the kernel's runtime _environment_ and, therefore, must be applied _prior_ to the kernel's actual launch.

In essence, what is needed is the ability to **associate a kernel's lifecycle management to the kernel's _specification_, where its environment and parameters are defined, while leaving kernel manager implementations to be the responsibility of the _application_.**

## Proposed Enhancement
This proposal abstracts the kernel process layer _within the existing `KernelManager` implementation_ thereby providing the ability to create custom kernel environments across **all** Jupyter applications that use `jupyter_client` today. 

In today's implementation, the `Popen` instance is returned by the `KernelManager`'s `_launch_kernel()` method.  Upon return, the method sets the manager's `kernel` attribute to the `Popen` instance, after which all lifecycle-related methods will call through to interact with the kernel process.

Instead, this proposal will introduce a layer or wrapper around the Popen instantiation such that this class instance (let's call it `PopenProvisioner` for now) will _contain_ the `Popen` instance and return _itself_ from the `_launch_kernel()` method.  Because the method signatures of the `PopenProvisioner` will be identical to those of `Popen`, the kernel's process management will operate just like today.  (Note that Jupyter Enterprise Gateway takes this approach with its [_process proxies_](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/system-architecture.html#process-proxy), but this solution is limited to the EG application not generally available to the ecosystem.)

Of course, `PopenProvisioner` will derive from a base class that defines the various methods.  These methods will look similar to the following:

```python
class KernelProvisionerBase(LoggingConfigurable):
    """Base class defining methods for Kernel Provisioner classes.

       Theses methods model those of the Subprocess Popen class:
       https://docs.python.org/3/library/subprocess.html#popen-objects
    """
    def poll(self) -> [int, None]:
        """Checks if kernel process is still running.

         If running, None is returned, otherwise the process's integer-valued exit code is returned.
         """
        pass

    def wait(self, timeout: Optional[float] = None) -> [int, None]:
        """Waits for kernel process to terminate.  As a result, this method should be called with
        a value for timeout.

        If the kernel process does not terminate following timeout seconds, a TimeoutException will
        be raised - that can be caught and retried.  If the kernel process has terminated, its
        integer-valued exit code will be returned.

        """
        pass

    def send_signal(self, signum: int) -> None:
        """Sends signal identified by signum to the kernel process."""
        pass

    def kill(self) -> None:
        """Kills the kernel process.  This is typically accomplished via a SIGKILL signal, which
        cannot be caught.
        """
        pass

    def terminate(self) -> None:
        """Terminates the kernel process.  This is typically accomplished via a SIGTERM signal, which
        can be caught, allowing the kernel process to perform possible cleanup of resources.
        """
        pass
```
The class will also define other methods for its initialization, launch, cleanup, etc.  In addition, these methods will be created with planned support for [_parameterized kernel launches_](https://github.com/jupyter/enhancement-proposals/pull/46) - since, realistically speaking, a majority of parameters affect the kernel process's environment.

We can decide whether the base class should be abstract (probably) or not along with which methods are abstract themselves as we near implementation.

`jupyter_client` will provide _the_ default `KernelProvisioner` implementation (e.g., `PopenProvisioner`) such that all existing kernels that do not specify a kernel provisioner will utilize an instance of the default class.  In addition, this default will be configurable in case a given installation wishes to use a different provisioner for all kernels in which one is not currently specified.

### Discovery
As noted in the problem statement, we need the ability to associate a kernel's lifecycle management (i.e., its process abstraction instance) to the kernel's _specification_.  It is not sufficient to rely on a single abstraction instance across all configured specifications.  However, because this proposal _should not affect existing installations using standard kernel specifications_, this only becomes an issue when _explicit_ abstractions (i.e., those not based on the default) are necessary.

To explicitly indicate a kernel environment provisioner, one would configure the corresponding kernel specification to include an `environment_provisioner` stanza within the `metadata` stanza, similar to the following...
```JSON
  "metadata": {
    "environment_provisioner": {
      "class_name": "my.provisioner.SlurmProvisioner",
      "config": {
      }
    }
  },
```
The KernelManager instance, with access to the `KernelSpecManager`, will check for the existence of such a stanza and instantiate the class associated with that's stanza's `class_name` entry.  Should the stanza not exist, the default provisioner will be instantiated and used.  Should the configured class name not be available, an exception will be raised, thereby failing the startup of the kernel.  (I view this as better than deferring to the configured default provisioner since the specification's configuration stanza probably won't apply to _that_ provisioner, etc.)

The `config` stanza will be passed to the provisioner's initializer and consist of  configuration settings pertaining to the provisioner and its subclasses.  We should also leverage whatever config-related functionality traitlets provide (assuming provisioners are subclasses of `LoggingConfigurable`).

### Provisioner Responsibilites
Once launched, the kernel _process's_ lifecycle-management will then be the responsibility of the instantiated provisioner. The provisioner will also be responsible for:
- Definition and consumption of provisioner-specific parameters that apply to the kernel process's environment. This includes a chance to apply substitutions into the startup command string.
- Provisioning of the kernel's connection.  The provisioned connection information will be accessible to the KernelManager at which time it can be persisted for use in collaboration, etc.

### Impact on existing implementations
If no environment provisioners are configured, there is no impact on existing implementations.  They will continue to work, just like today.  The difference will be that when the appropriate version of `jupyter_client` is installed, interaction with the kernel's process will go through an additional (nearly pass-thru) layer.

In addition, existing implementations will be able to leverage parameterized kernel launches, once available and, if kernel provisioners are configured, be able to leverage their offerings immediately.

When environment provisioners are configured, any kernel specifications they provide will be immediately available to applications.

No additional packages will be necessary - all functionality is baked into `jupyter_client` - and the previously installed KEP provisioning package.

#### Existing KernelManager subclasses
By embracing `jupyter_client` and its `KernelManager` class, this proposal doesn't introduce any migration issues and most subclasses of `KernelManager` should continue to work.  Note that some `KernelManager` subclasses that completely override lifecycle-action methods will not be able to leverage this functionality - but that's their intent in the first place.  

What applications subclass `KernelManager` today?  I know that Enterprise Gateway already provides its own process abstraction via a subclass of `KernelManager`, and will need to coordinate with appropriate `jupyter_client` releases once implemented (but I have an inside scoop on that repo :smile: ).

Should I post this question to the Jupyter Google Group, Discourse, anywhere else?  I know that [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels) subclasses `KernelSpecManager` - as well as others - but they still leverage jupyter_client's `KernelManager` directly - so they should not be an issue.

### Naming
Here are a few naming suggestions, some of which are more appropriate as a _topic_ (e.g., _provisioning_) than an _implementation_ (e.g., _provider_ or _provisioner_).
- Kernel Process Provider
- Kernel Environment Provider
- Kernel Provisioning/Provisioner
- Kernel Environment Provisioning/Provisioner
- Kernel Process Proxy (adopt Enterprise Gateway's terminology)
- ???

Because this abstraction is contained within the existing `KernelManager` implementation, the _Kernel_ in the name could be dropped as it's inferred.

I prefer _Environment Provisioning_ as a topic and _Environment Provisioner_ as an implementation name but really have no strong affinity to either and am open to suggestions. The acronym `KEP` could be used for abbreviations where necessary (where the 'K' for Kernel makes the inference explicit).

Alternate names for `PopenProvisioner` could be: `JupyterClientProvisioner` or `GenericProvisioner`.  I suspect many custom provisioners will derive from this implementation.
##
I've gone ahead and cc'd folks with which I've shared these ideas.  Please feel free to add anyone else you think might be interested.

cc: @blink1073, @echarles, @lresende, @Zsailer

kevin-bates:
Here are initial changes to introduce [Kernel Provisioning](https://github.com/jupyter/jupyter_client/issues/608) (aka, _Environment Provisioning_).  By default, all kernel launches will use the `LocalProvisioner` unless a different `provisioner_name` is specified in the kernel specification's `metadata.kernel_provisioner` stanza.  The `LocalProvisioner` should yield behavior equivalent to today's kernel lifecycle management.  As a result, we should see zero changes to the existing tests (other than changes needed for continuing migration to `pytest`).  Any changes to existing tests must be justified so as to preserve compatibility - few exceptions are expected.

Provisioners are registered via the [entrypoints](https://pypi.org/project/entrypoints/) framework and discovered via the group name `jupyter_client.kernel_provisioners`.  When kernelspecs are located, each is checked for a `provisioner_name` entry.  If found, the provisioner is loaded using `entrypoints`.  If the load fails, a warning message is logged indicating the provisioner package is not available and _that_ kernel specification is dropped from the results (if using `get_all_specs()`) or results in a `NoSuchKernel` exception (if using `get_kernel_spec(name)`).  If no `provisioner_name` is found (which will constitute nearly 100% of the cases initially), the `LocalProvisioner` is instantiated as if it had been configured (per the first paragraph).

A singleton is used as a kernel provisioner factory and cache.  A map of provisioner name to `EntryPoint` instance is built at startup and maintained during the process's lifetime whenever the kernelspec manager encounters a provisioner name that is not in the map (and that entry point load succeeds).

This pull request will remain in DRAFT mode until the following has been addressed:
- [x] Determine what to do about the connection file and zmq ports.  Clearly, the port attributes must remain on `KernelManager`, but the point at which they are set will need to change.  Since the provisioner should return the connection file information - which is also consistent with the behavior described in [_Kernel handshaking pattern proposal_](https://github.com/jupyter/enhancement-proposals/pull/66).  Also, `write_connection_file()` is called before launch (which makes sense for local-only and non-handshaking) but this should be moved _closer_ to the actual launch. (Utilized `self.parent` (ie., kernel_manager) to write local connection file in local-provisioner for now.)
- [x] Should the `kernel_cmd` trait be removed as its been deprecated since 4.0 (April 2014)?  Its removal will resolve one potential (and minor) backwards compatibility issue for any applications using that trait. (See `TODO: Potential B/C issue...` in comment in `EnvironmentProvisionerBase.pre_launch()`.) (Removed in #643)
- [x] Need to convert any of the Enterprise Gateway process-proxy implementations to provisioners as a POC to be sure there aren't any glaring items missing.  If others are doing similar things, it would be good to make sure this proposal addresses those instances. (POC instances of Hadoop YARN, Distributed(ssh), K8s and Docker provisioners have been implemented [here](https://github.com/kevin-bates/remote_provisioners).)
- [x] Add test to include custom provisioner. 
- [x] Add documentation
- [x] Look at simplifying interface/base class

Implements/Resolves #608


jupyter_client/launcher.py
```diff
@@ -40,9 +41,103 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
...
+        proc = Popen(cmd, **kwargs)
+    except Exception as exc:
+        msg = (
+            "Failed to run command:\n{}\n"
+            "    PATH={!r}\n"
+            "    with kwargs:\n{!r}\n"
+        )
+        # exclude environment variables,
+        # which may contain access tokens and the like.
+        without_env = {key:value for key, value in kwargs.items() if key != 'env'}
```
MSeal, line 63: Might be good to do a try / finally block around these so if there's message formatting or logging errors it still raises the original exception

jupyter_client/launcher.py
```diff
@@ -40,9 +41,103 @@ def launch_kernel(cmd, stdin=None, stdout=None, stderr=None, env=None,
...
+            proc = await asyncio.create_subprocess_exec(*cmd, **kwargs)
+    except Exception as exc:
+        msg = (
+            "Failed to run command:\n{}\n"
+            "    PATH={!r}\n"
+            "    with kwargs:\n{!r}\n"
+        )
+        # exclude environment variables,
+        # which may contain access tokens and the like.
+        without_env = {key:value for key, value in kwargs.items() if key != 'env'}
```
MSeal, line 128: Since this is repeated it might be good to capture it inside a helper function

jupyter_client/manager.py
```diff
@@ -616,22 +667,15 @@ async def shutdown_kernel(self, now=False, restart=False):
...
         else:
             self.request_shutdown(restart=restart)
             # Don't send any additional kernel kill messages immediately, to give
             # the kernel a chance to properly execute shutdown actions. Wait for at
             # most 1s, checking every 0.1s.
-            await self.finish_shutdown()
+            await self.finish_shutdown(restart=restart)
 
-        # See comment in KernelManager.shutdown_kernel().
-        overrides_cleanup = type(self).cleanup is not AsyncKernelManager.cleanup
```
MSeal, line 628: Seems like this mechanism was removed?

jupyter_client/provisioning.py
```diff
@@ -0,0 +1,407 @@
...
+        restart is True if this operation precedes a start launch_kernel request.
+        """
+        pass
+
+    @abstractmethod
+    async def launch_kernel(self, cmd: List[str], **kwargs: Any) -> Tuple['KernelProvisionerBase', Dict]:
+        """Launch the kernel process returning the class instance and connection info."""
+        pass
+
+    async def cleanup(self, restart=False) -> None:
```
MSeal, line 82: Should we force this to be abstract as well?

jupyter_client/provisioning.py
```diff
@@ -0,0 +1,407 @@
...
+        """Perform any steps in preparation for kernel process launch.
+
+        This includes applying additional substitutions to the kernel launch command and env.
+        It also includes preparation of launch parameters.
+
+        Returns potentially updated kwargs.
+        """
+
+        env = kwargs.pop('env', os.environ).copy()
+        # Don't allow PYTHONEXECUTABLE to be passed to kernel process.
```
MSeal, line 99: Probably should generalize this in the future for other kernels (not a priority)

MSeal (commented):
A few minor comments, but it's looking clean @kevin-bates .

jupyter_client/manager.py
kevin-bates: Correct.  Since the plan is to release this via a major release, my hope was that previous deprecations would be removed.

jupyter_client/provisioning.py
kevin-bates: I'm okay with that.  Since the default implementation didn't need this (although the connection file management is a concern that could change that), I didn't want to force its (potentially unnecessary) implementation.  That said, `cleanup()` is something generally expected - with the act of nothing to do being the exception.

jupyter_client/provisioning.py
kevin-bates: Are you referring to _an_ ability for provisioners to scrub the environment passed to the kernel?  That is a good idea, although the provisioner won't necessarily know what kernel it's provisioning.  I suspect the provisioner could use the kernel_spec's `language` value (if present) to make some inference though.   

Perhaps a method to the effect of `_finalize_env()` might be helpful to subclassed provisioners?

jupyter_client/provisioning.py
MSeal: I'm fine either way. I think forcing implementers to consider what cleanup they need isn't the worst overhead even if it's a `pass` implementation but my opinion there isn't a strong one.

jupyter_client/provisioning.py
MSeal: Yeah if we're doing a major version bump it might be nice to have the abstraction expanded so it's at least partially flexible to non ipython kernel contexts here. To avoid scope creep I'd keep it in a separate PR though if we do tackle it

MSeal:
Let me know if I am missing other issues that need attention. I have less time available so I am missing some threads here and there

kevin-bates:
Thanks Matt - I really appreciate your reviews thus far.

jupyter_client/launcher.py
```diff
@@ -50,7 +55,122 @@ def launch_kernel(
...
+        proc = Popen(cmd, **kwargs)
+    except Exception as exc:
+        _handle_subprocess_exception(exc, cmd, env, **kwargs)
+
+    _finish_process_launch(proc, stdin, interrupt_event)
+
+    return proc
+
+
+async def async_launch_kernel(
```
davidbrochart, line 79: Why do we need `launch_kernel` and `async_launch_kernel`?

kevin-bates: I wasn't sure if some provisioner authors might want sync Popen.  However, this was done prior to your refactor and I suspect we could/should just go with `async_launch_kernel` only.  It should be noted though that Windows `asyncio.create_subprocess_exec` has issues (noted in the comment) and we'd still use `Popen` under the covers for that platform.

I also noticed yesterday that [all objects from `launcher.py` are exposed](https://github.com/jupyter/jupyter_client/blob/98faaf91d00b4fbf17b2fbf0b08b1fd23adc27bf/jupyter_client/__init__.py#L10) - which implies `launch_kernel()` can be used directly - although not going through `KernelManager._launch_kernel()` doesn't make a whole lot of sense to me.  I wonder if we should just assume that's the supported use-case here and no direct use is permitted.  If so, we should probably retract `launcher.py`'s presence in `__init__.py` and can move `launcher.py` into `provisioning`.

davidbrochart: I think the only thing async subprocesses offer in our case is async `wait` (for the process to terminate), which I'm not sure we need since we send `SIGTERM` then `SIGKILL`, IIRC, but we never `await` forever.

kevin-bates: Thanks David, you are correct.  I happen to find `Popen` easier to deal with anyway and would be fine only having the original (and not offering an async version).  Using `Popen` also simplifies the WIndows case as well.  The important aspect is that the corresponding methods on `KernelProvisionerBase` remain `async`.

davidbrochart: But then don't we need both to be of the same type (sync or async)? It doesn't hurt to have an async `launch_kernel()` even if there is nothing async in it (but it will have to be awaited for sure).

kevin-bates: No.  When used, `launch_kernel()` is always embedded in a provisioner, and provisioner methods are only async.  It is no longer called directly.

jupyter_client/launcher.py
kevin-bates: Added additional try/except block in https://github.com/jupyter/jupyter_client/pull/612/commits/e6e381ecea109f1564a27c15ac06a2427917f9cc.

kevin-bates:
I believe this is ready for review - thank you for your patience.

blink1073 (approved):
Thanks for the mountain of work you did on this @kevin-bates!

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+single ``KernelManager`` class at any time. While applications could
+introduce a ``KernelManager`` subclass of their own, that
+``KernelManager`` was then tied directly to *that* application and
+thereby not usable as a ``KernelManager`` in another application. As a
+result, we consider the ``KernelManager`` class to be an
+*application-owned entity* upon which application-specific behaviors can
+be implemented.
+
+Kernel provisioners, on the other hand, are contained within the
+``Kernel Manager`` (i.e., a *has-a* relationship) and applications are
```
davidbrochart, line 37: Not sure if this is a typo:
```suggestion
``KernelManager`` (i.e., a *has-a* relationship) and applications are
```

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+
+.. tip::
+    Use ``kernel_id`` as a discovery mechanism from your provisioner!
+
+Here's a prototyped implementation of a couple of the abstract methods
+of :class:`KernelProvisionerBase` for use in an Hadoop YARN cluster to
+help illustrate a provisioner's implementation.  Note that the built-in
+implementation of :class:`LocalProvisioner` can also be used as a reference.
+
+Notice the internal method `_get_application_id()`.  This method is
```
davidbrochart, line 197:
```suggestion
Notice the internal method ``_get_application_id()``.  This method is
```

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+what the provisioner uses to determine the YARN application (i.e.,
```
davidbrochart, line 198:
```suggestion
what the provisioner uses to determine if the YARN application (i.e.,
```

jupyter_client/tests/test_provisioning.py
```diff
@@ -0,0 +1,345 @@
+"""Test Provisionering"""
```
davidbrochart, line 1:
```suggestion
"""Test Provisioning"""
```

davidbrochart (approved):

kevin-bates:
Thank you @blink1073 and @davidbrochart for your reviews.

davidbrochart:
I just merged https://github.com/jupyter/jupyter_client/pull/645, looks like you have conflicts now, but that should not be a big deal. Sorry Kevin!

kevin-bates:
No worries at all David.   I think we should get at least another approval before merging this anyway.  I'll try to resolve the conflict later today.

blink1073:
Let's plan to merge this by next Thursday (22 July) if we don't get any more feedback.   @MSeal you had some early feedback if you'd like to take a final look.

jupyter_client/kernelspecapp.py
```diff
@@ -270,6 +271,22 @@ def start(self):
             self.exit(e)
 
 
+class ListProvisioners(JupyterApp):
+    version = __version__
+    description = """List available provisioners for use in kernel specifications."""
+
+    def start(self):
+        kfp = KernelProvisionerFactory(parent=self).instance()
```
Zsailer, line 279: I believe this should be:
```
KernelProvisionerFactory.instance(parent=self)
```
Instantiating this class using the __init__ method doesn’t set a global instance of this singleton. When instance is called, it creates a new instance of this class and overrides the original one created. As a consequence, parent is not actually set.

Zsailer (approved):
This looks great @kevin-bates! 

I added one minor comment in line. The PR is going to need a rebase before we can merge. Otherwise, I think it's ready to go!

jupyter_client/kernelspecapp.py
kevin-bates: Good catch Zach - thank you.  I've updated this occurrence (the other two were correct).

kevin-bates:
It looks like this PR encountered an updated version of `mypy` that complains about missing packages.  In this case, it was producing an error related to the (optional) `paramiko` package used in `ssh/tunnel.py`.  I've updated the GH workflow to add `--install-types --non-iteractive` to resolve this, but this [particular commit](https://github.com/jupyter/jupyter_client/pull/612/commits/d3e01ea1b5cc872c42533fb3bd7f8b90506e3773) may be necessary in master prior to this PR's merge (and assuming that's the correct way to address this).

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+be implemented.
+
+Kernel provisioners, on the other hand, are contained within the
+``KernelManager`` (i.e., a *has-a* relationship) and applications are
+agnostic as to what *kind* of provisioner is in use other than what is
+conveyed via the kernel's specification (kernelspec). All kernel
+interactions still occur via the ``KernelManager`` and ``KernelClient``
+classes within ``jupyter_client`` and potentially subclassed by the
+application.
+
```
echarles, line 43: Does the KernelManager or KernelClient API changes (additional (optional) methods, method signature change...) ? Is there any impact for current consumers of those APIs?

Maybe worth stating this in the documentation?

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+      "metadata": {
+        "kernel_provisioner": {
+          "provisioner_name": "k8s-provisioner",
+          "config": {
+              "image_name": "my_docker_org/kernel:2.1.5",
+              "max_cpus": 4
+          }
+        }
+      },
+
```
echarles, line 82: Would it be possible to have an additional config_schemastanza that would contain a json schema (https://json-schema.org/) of the config?

That schema could be used by consumer application (eg jupyterlab to show a form in a UI).

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+in the role meant for this kernel. If not, an exception will be thrown.
+
+.. code:: python
+
+    class RBACProvisioner(LocalProvisioner):
+
+        role: str = Unicode(config=True)
+
+        async def pre_launch(self, **kwargs: Any) -> Dict[str, Any]:
+
```
echarles, line 151: Looks like there is a well-defined lifecycle, cfr `pre_launch`. Is it possible to list those lifecycle methods here (or link them to the API docs, or to the source)?

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+If you'd like to launch your kernel in an environment other than the
+local server, then you will need to consider subclassing :class:`KernelProvisionerBase`
+directly.  This will allow you to implement the various kernel process
+controls relative to your target environment.  For instance, if you
+wanted to have your kernel hosted in a Hadoop YARN cluster, you will
+need to implement process-control methods like :meth:`poll` and :meth:`wait`
+to use the YARN REST API.  Or, similarly, a Kubernetes-based provisioner
+would need to implement the process-control methods using the Kubernetes client
+API, etc.
+
```
echarles, line 175: Are thos `poll`, `wait`... lifecycle methods enforced by the API? Here also, would be good to have a list (or link) to the lifecycle methods (if the number is not too high and does not break the reading flow).

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+    class RBACProvisioner(LocalProvisioner):
+
+        role: str = Unicode(config=True)
+
+        async def pre_launch(self, **kwargs: Any) -> Dict[str, Any]:
+
+            if not self.user_in_role(self.role):
+                raise PermissionError(f"User is not in role {self.role} and "
+                                      f"cannot launch this kernel.")
+
```
echarles, line 155: Maybe clearly state that the `user_in_role` method has to be implemented by the Provisioner and is not part of the Provisioner API.

docs/provisioning.rst
```diff
@@ -0,0 +1,355 @@
...
+The final step in getting your custom provisioner deployed is to add a
+``kernel_provisioner`` stanza to the appropriate ``kernel.json`` files.
+This can be accomplished manually or programmatically (in which some
+tooling is implemented to create the appropriate ``kernel.json`` file).
+In either case, the end result is the same - a ``kernel.json`` file with
+the appropriate stanza within ``metadata``. The *vision* is that kernel
+provisioner packages will include an application that creates kernel
+specifications (i.e., ``kernel.json`` et. al.) pertaining to that
+provisioner.
+
```
echarles, line 318: Well, I can see a vision where a single `kernel.json` and some `metadata` (the default values), a `config_schema`. The consumer application would be free to use the schema and pass some `user_config` that  would override the default ones. How those `user_config` would be passed is prolly still to be discussed.

In that sense, I wonder if the additional stanza in the `kernel.json`should be:

- `config`-> `default_config`
- `config_schema`

The `user_config` should be not persisted in the kernel.json but could perfectly find its place in e.g. the metadata of the ipynb notebook.

echarles:
Left a few comments/questions with the future idea to embed a schema for the config as to have the ability for the user to override the config.

I see 3 approvers already. I don't want to block the merge and my questions/ideas can for sure be discussed in subsequent PRs.

Feel free to merge, but really interested to get feedback on your comments.

docs/provisioning.rst
kevin-bates: Nothing changed on `KernelClient`.  As for `KernelManager` there were some changes to "internal" methods and I agree those kinds of changes should be documented (primarily calls to the Popen instance are replaced with calls to the provisioner instance).  As a result, only `KernelManager` subclass authors should be affected (and hopefully minimally).  However, I would suggest those be part of the 7.0 release docs (or whatever major release this falls in) rather than embedded in functional/dev docs pertaining to Kernel Provisioning itself.

docs/provisioning.rst
kevin-bates: Yes, I agree, but feel this falls into the _kernel paramertization_ discussion, which should occur once this is merged.  I believe there will need to be two forms of schemas - one schema produced by the kernel itself describing kernel-specific parameters, the other schema produced by the provisioner describing provisioner-specific parameters.

docs/provisioning.rst
kevin-bates: I'm not sure what you mean by this - nearly every method pertains to the lifecycle of the kernel processing since that is exactly what `KernelManager` (which drives the calls on the provisioner) is responsible for.  Are you suggesting that we document in the provisioner's method where/when they are called from the `KernelManager`?

Could you please provide an example?  I'm happy to add that kind of thing provided it adds value.

I do add the call relationship in the [base provisioner class](https://github.com/jupyter/jupyter_client/pull/612/files#diff-b370bb301ffe7b9de4edb6f77835f50f930c12552fdacd203839f9710daeedb5R144-R161) - which I view as the canonical doc text for all provisioners.

docs/provisioning.rst
kevin-bates: Again, it's not clear to me what you mean - an example will help me.  I also don't understand what you mean by _enforced by the API_.  Sorry.

docs/provisioning.rst
kevin-bates: Again, this falls into the parameterization discussion.  However, I believe we'll need to state that any _required properties_ specified in the schema must have a default value since we can't assume that every consumer of a kernelspec is going to be able to provide parameters.

I agree that persisting user-provided values in the notebook itself is reasonable.  We'll likely want some association/namespacing of those attributes to the associated schema as well.  Then, when that notebook is executed with the same provisioner/schema, _those_ values could be pre-filled/used.  Still lots to work out on that front.

docs/provisioning.rst
kevin-bates: That makes sense - update pushed.

docs/provisioning.rst
kevin-bates: Hmm - in thinking about this just now, I think we also need to separate _configuration_ from _parameters_.  With configuration, I view this as more static in nature.  It can be specified in the `kernel.json`, a config file, or something else - but that aspect of things is completely provisioner-specific and the provisioner is responsible for distinguishing configuration from parameters.  

Examples of configuration might be the resource manager hostname for _this_ installation.  In that case, you don't want to expose that to the end-user, but instead, it's a value that is required by every invocation of that particular provisioned kernel.

kevin-bates:
I really appreciate your review Eric - thank you.  I've left you with a couple of questions to clarify my understanding regarding your comments.

docs/provisioning.rst
echarles: > nearly every method pertains to the lifecycle of the kernel processing since that is exactly what KernelManager

Then it is fine to not embed more details in the doc.

> Are you suggesting that we document in the provisioner's method where/when they are called from the KernelManager?

I did some time ago a sequence diagram for the kernel providers (not merged). We can add that kind of details later on if it makes sense.

docs/provisioning.rst
echarles: I should look at the code but did not. My question is "Do all remote provisioners need to implement poll, wait...?". Another way to ask is "Are pool, wait... abstract methods that need to be implemented by all remote provisioners?

docs/provisioning.rst
echarles: Interested discussion on nomenclature. Let's say

- `resource_manager_hostname` has to be given to the provisioner, but is enforced by the kernelspec author and can not changed by the user
- `memory` has to be given to provisioner, has a default value of `1G` defined by the kernel author, but can be changed by the user with value `1G`, `5G` or `10G` as stated in a json schema defined by the kernel author.

I see them all as the same thing, by some can not be changed by the user.

Thinking loud, I could see name like `system parameter`/ `user parameter`  or `system config` / `user config`.

Alternative to system vs user could be static vs dynamic.

docs/provisioning.rst
kevin-bates: Yes, that was helpful to the providers' discussion and a provisioner-based version would be helpful as well.

docs/provisioning.rst
kevin-bates: I agree.  All parameters, _system_ or _user_, _static_ or _dynamic_, should be available to everyone.  Where that presentation takes place is, I believe, where the differences lie.  

I view system parameters as static values where you would not want them changed by the user and those would be oriented to the provisioner's underlying configuration system (which may or may not be trailtets-based). In some (typically single-user) cases, _everyone_ can be split into different roles, namely _admin_ and _user_, where a given _user_ is not exposed to a _system_ parameter.  In other cases, you might want various system parameters exposed to the user for modification.  Since any user-exposed parameters would need to be reflected in the schema and since it could be the case that _system parameters_ should also be available to users for modification, perhaps a _meta-propery_ like `is_hidden` is warranted.  This all implies, then, that the schema itself would need to be configurable (which I suppose is already the case in order to set appropriate enumerated types, etc.). 

Note, there's nothing preventing a provisioner from implementing its user-modifiable parameters based on an underlying configuration system (e.g., configurable traits) so long as the _runtime_ values override the configured values.

These are all the kinds of things that will require hashing out during _parameterized kernels_ discussions.

echarles: > These are all the kinds of things that will require hashing out during parameterized kernels discussions.

Agree with the previous comment. fyi I will start experimenting ui in jlab and push more stanza just to make sense of the parametrisation options. I hope to come with a demo at the next server meeting.

echarles:
Feeling good with that warning 👍 

`[W 2021-07-26 09:53:42.722 ServerApp] Kernel 'datalayer_env' is referencing a kernel provisioner ('loca-provisioner') that is not available.  Ensure the appropriate package has been installed and retry.`

echarles:
As further experiments, I have poc on top of this PR what it would be for JupyterLab to get a json schema baked in the kernelspec, display a form pre-populated with the default values shipped in the schema, send the user values to the kernel service that replaces the placholders defined in the kernelspec with the user values to launch the kernel with parameters. It is doable, we just need to review the changes (not so much to be honest).

I hope we can discuss this at the next server meeting.

Untitled

cc/ @kevin-bates @Zsailer @blink1073 @mlucool @goanpeca @ ericdatakelly

docs/provisioning.rst
kevin-bates: Ah - yes, abstract methods are used where "enforcement" is required.  These methods are documented on [`KernelProvisionerBase`](https://github.com/kevin-bates/jupyter_client/blob/fc3227ee71ea7db8f54614d5eb58cec99bce98b5/jupyter_client/provisioning/provisioner_base.py#L24) and in the API portion of the provisioner docs.

echarles (approved):

blink1073:
Huge thanks to @kevin-bates for all the work you did here!

blink1073:
Ready for preview: https://pypi.org/project/jupyter-client/7.0.0a1/

meeseeksmachine:

https://discourse.jupyter.org/t/errors-with-asyncio-io-loop-in-custom-ipython-kernel/10908/2

PR #586, opened 2020-11-06
Drop Travis CI, add GitHub Actions

davidbrochart:

PR #615, opened 2021-02-02
Unpin jedi

bollwyvl:
Now that IPython 7.20 is out, the jedi pin should no longer be needed :tada:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,16 @@ setup_args = dict(
     ],
     python_requires  = '>=3.5',
     extras_require   = {
-        'test': ['ipykernel', 'ipython', 'mock', 'pytest', 'pytest-asyncio', 'async_generator', 'pytest-timeout'],
+        'test': [
+            'async_generator',
+            'ipykernel',
+            'ipython',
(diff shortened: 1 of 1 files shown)


bollwyvl:
Ah, of course, this wouldn't fix 3.6, as 7.20 is 3.7+...

bollwyvl:
So... what's the story behind no windows testing, if i might ask?

kevin-bates (approved):
LGTM. I guess I would have expected ipython to enforce the conditional management of jedi versions, but I suppose this will ensure that happens.

kevin-bates:
> So... what's the story behind no windows testing, if i might ask?

Looks like these were stopped via #586, but are being revived by #613.  The problem is that they all appear to hang _after_ the tests have completed.

bollwyvl:
>  ipython to enforce the conditional management of jedi versions

IPython no longer supports <py3.7, and the last released version does not have a top pin. but let's not drag _that_ beast out any longer.


> hang after the tests have completed.

GAAAAH. that's rough... hooray for progress, though. will take a gander.

PR #619, opened 2021-02-12
Enable the macos tests for py3.8 and 3.9

kevin-bates:
While troubleshooting CI issues on #617, I found that the excluded `macos-latest` builds were successful on python 3.8 and 3.9.  This update adds those platform/version combos back into the mix.


blink1073 (approved):
Thanks!

Issue #618, opened 2021-02-05
Kernel Shutdown Proposal


mlucool:
[Shutdown kernel](https://github.com/jupyter/jupyter_client/blob/4be21b89f0b6745f86d3d719b61557d2d9d55534/jupyter_client/manager.py#L372) states the following in it's doc:
> This attempts to shutdown the kernels cleanly by:
> 1. Sending it a shutdown message over the control channel.
> 2. If that fails, the kernel is shutdown forcibly by sending it a signal.

While this seems like the right thing to do, it does not work well for kernels like IPykernel (but may be fine for kernels like Xeus-Python). The problem is the signal sent via the control channel may be blocked by a running process, so there is no real chance to cleanly shut it down.

I propose that we add a step between 1 and 2 which tries SIGINT and then tries SIGTERM before finally using SIGKILL (step 2 today). What do you think?

kevin-bates:
hi @mlucool, this makes sense to me - thanks for raising it.  

I have some questions, comments...

- I'm hoping we could just invoke `interrupt_kernel()` for the SIGINT step.  Chances are that a kernel that supports message-based interrupts have probably responded to the message-based termination anyway, and this would be cleaner.
- I'm assuming interrupt would occur after the 5-second period to determine if the message-based termination succeeded, correct?  
	- Is the SIGTERM even necessary at that point?  I see that ipykernel and IPython don't handle SIGTERM anyway, so, relative to that kernel, there's really no need, although SIGTERM prior to SIGKILL is more correct and should probably be honored.  
	- I'd prefer we not add another 5-second period following the SIGTERM and prior to SIGKILL.  Perhaps we could split the poll loops in 60/40 or something - so that the overall time for these edge cases isn't necessarily increased.

Another approach would be to unconditionally issue an `interrupt_kernel()` call prior to `request_shutdown()` and be done.  Although cleaner, this would lengthen the shutdown sequence by a message.

((I'm hoping we could drop the `if hasattr(signal, 'SIGKILL'):` code and just use `self.kernel.kill()` and `self.kernel.terminate()`.  Sending the signal and calling the method is the same thing anyway.))

mlucool:
> Another approach would be to unconditionally issue an interrupt_kernel() call prior to request_shutdown() and be done. Although cleaner, this would lengthen the shutdown sequence by a message.

I like this idea. Still, it seems like SIGTERM is needed for things inside the kernel to be able to cleanly shutdown.

New proposal:
[...]

PR #620, opened 2021-02-22
Interrupt and sigterm before sigkill kernel.

Carreau:
This implements the discussion in #618, to try to more progressively stop
the kernels.

1) this always sends and interrupt before any shutdown requests;
  goal being to stop any processing happening that may block any event
  loop.

2) sends the shutdown_requests (no changes here).

3) wait 50% of the wait time, and sends a "terminate" in quote as this
depends on your platform/os, and the type of kernel you have.
    - if subprocess.Popen calls `.terminate()` which is SIGTERM on unix; the same as
    `.kill()` on windows; if not a Popen instance send SIGTERM.

4) wait the other 50% and kills, (same as before).

This does this both on the sync client and async client.

TBD: write tests and docs; and test more properly;

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -350,8 +350,15 @@ class KernelManager(ConnectionFileMixin):
             waittime = max(self.shutdown_wait_time, 0)
         self._shutdown_status = _ShutdownStatus.ShutdownRequest
 
-        # wait 50% of the shutdown timeout...
-        for i in range(int(waittime / 2 / pollinterval)):
+        def poll_or_sleep_to_kernel_gone():
+            """
+            Poll until the kernel is not responding,
+            then wait (the subprocess), until process gone.
+
+            After this function the kernel is either:
+                - still responding; or
+                - subprocess has been culled.
+            """
             if self.is_alive():
                 time.sleep(pollinterval)
             else:
@@ -359,6 +366,11 @@ class KernelManager(ConnectionFileMixin):
                 if self.has_kernel:
                     self.kernel.wait()
                     self.kernel = None
+                return True
+
+        # wait 50% of the shutdown timeout...
+        for i in range(int(waittime / 2 / pollinterval)):
+            if poll_or_sleep_to_kernel_gone():
                 break
         else:
             # if we've exited the loop normally (no break)
@@ -367,13 +379,7 @@ class KernelManager(ConnectionFileMixin):
             self._shutdown_status = _ShutdownStatus.SigtermRequest
             self._send_kernel_sigterm()
             for i in range(int(waittime / 2 / pollinterval)):
-                if self.is_alive():
-                    time.sleep(pollinterval)
-                else:
-                    # If there's still a proc, wait and clear
-                    if self.has_kernel:
-                        self.kernel.wait()
-                        self.kernel = None
+                if poll_or_sleep_to_kernel_gone():
                     break
             else:
                 # OK, we've waited long enough.
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -32,6 +32,13 @@ from .managerabc import (
 )
 
 class _ShutdownStatus(Enum):
+    """
+
+    This is so far used only for testing in order to track the internal state of
+    the shutdown logic, and verifying which path is taken for which
+    missbehavior.
+
+    """
     Unset = None
     ShutdownRequest = "ShutdownRequest"
     SigtermRequest = "SigtermRequest"
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -35,7 +35,7 @@ class _ShutdownStatus(Enum):
     Unset = None
     ShutdownRequest = "ShutdownRequest"
     SigtermRequest = "SigtermRequest"
-    SigKillRequest = "SigKillRequest"
+    SigkillRequest = "SigkillRequest"
 
 
 class KernelManager(ConnectionFileMixin):
@@ -378,7 +378,7 @@ class KernelManager(ConnectionFileMixin):
                 # OK, we've waited long enough.
                 if self.has_kernel:
                     self.log.debug("Kernel is taking too long to finish, killing")
-                    self._shutdown_status = _ShutdownStatus.SigKillRequest
+                    self._shutdown_status = _ShutdownStatus.SigkillRequest
                     self._kill_kernel()
 
     def cleanup_resources(self, restart=False):
@@ -663,7 +663,7 @@ class AsyncKernelManager(KernelManager):
 
         except asyncio.TimeoutError:
             self.log.debug("Kernel is taking too long to finish, killing")
-            self._shutdown_status = _ShutdownStatus.SigKillRequest
+            self._shutdown_status = _ShutdownStatus.SigkillRequest
             await self._kill_kernel()
         else:
             # Process is no longer alive, wait and clear
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -45,6 +45,7 @@ class KernelManager(ConnectionFileMixin):
     """
 
     def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         self._shutdown_status = _ShutdownStatus.Unset
 
     _created_context = Bool(False)
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -12,6 +12,8 @@ import sys
 import time
 import warnings
 
+from enum import Enum
+
 import zmq
 
 from ipython_genutils.importstring import import_item
@@ -29,12 +31,21 @@ from .managerabc import (
     KernelManagerABC
 )
 
+class _ShutdownStatus(Enum):
+    Unset = None
+    ShutdownRequest = "ShutdownRequest"
+    SigtermRequest = "SigtermRequest"
+    SigKillRequest = "SigKillRequest"
+
 
 class KernelManager(ConnectionFileMixin):
     """Manages a single kernel in a subprocess on this host.
 
     This version starts kernels with Popen.
     """
+
+    def __init__(self, *args, **kwargs):
+        self._shutdown_status = _ShutdownStatus.Unset
 
     _created_context = Bool(False)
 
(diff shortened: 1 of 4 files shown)


Carreau:
I'm thinking about changing the `shutdown_kernel` to return an object with how far it went to stop the kernel; something that we can grow later; which basically tell us if the kernel replied to the shutdown_request; or had to go to the sigterm, or the sigkill. That might make it easier to test (as it is hard to get info from a dead process), and thus implement multiple kernels that will forcibly not shutdown on request, and not shut down on sigterm.

MSeal (approved):
Overall looks fine beyond some minor comments. Might be worth making a manual test plan for this one to make sure we have someone try it out in a few event loops patterns / os's? Just because I could see some odd edge case in windows for M1 osx that our automation might not catch.

jupyter_client/manager.py
```diff
@@ -29,13 +31,23 @@
     KernelManagerABC
 )
 
+class _ShutdownStatus(Enum):
+    Unset = None
+    ShutdownRequest = "ShutdownRequest"
+    SigtermRequest = "SigtermRequest"
+    SigKillRequest = "SigKillRequest"
```
kevin-bates, line 38: nit: Could you please either lowercase the 'K' in `SigKillRequest` or raise the 'T' in `SigtermRequest`?  I'd prefer the former.

jupyter_client/manager.py
```diff
@@ -584,9 +647,23 @@ async def finish_shutdown(self, waittime=None, pollinterval=0.1):
...
+                    self._async_wait(pollinterval=pollinterval), timeout=waittime / 2
+                )
+            except asyncio.TimeoutError:
+                self.log.debug("Kernel is taking too long to finish, terminating")
+                self._shutdown_status = _ShutdownStatus.SigtermRequest
+                await self._send_kernel_sigterm()
+
+            await asyncio.wait_for(
+                self._async_wait(pollinterval=pollinterval), timeout=waittime / 2
+            )
```
kevin-bates, line 662: I think this spin-wait needs to be associated with the `except` block - otherwise, it will occur again for the successful shutdown-request case (which will immediately exit, but probably not the intent).

jupyter_client/manager.py
```diff
@@ -584,9 +647,23 @@ async def finish_shutdown(self, waittime=None, pollinterval=0.1):
...
             waittime = max(self.shutdown_wait_time, 0)
         try:
-            await asyncio.wait_for(self._async_wait(pollinterval=pollinterval), timeout=waittime)
+            try:
+                self._shutdown_status = _ShutdownStatus.ShutdownRequest
+                await asyncio.wait_for(
+                    self._async_wait(pollinterval=pollinterval), timeout=waittime / 2
+                )
+            except asyncio.TimeoutError:
+                self.log.debug("Kernel is taking too long to finish, terminating")
```
kevin-bates, line 656: We should add this debug statement to the synchronous version.

kevin-bates (commented):
(Sorry, I fat-fingered the 'single-comment' button, and its like it turned previous comments into single comments as well, so I continued in that mode.)

This is looking good and I believe introducing interrupt and terminate to the shutdown mix is great.

I don't think this is anything to change, but just noting that the shutdown-request status is set after the request, while the signal-based statuses are set prior to their respective requests.  Although not consistent, I think it's probably the right approach across the three and I'm not advocating for additional (pre/post) statuses, but with Kernel Provisioning, the `terminate()` and `kill()` are not necessarily "signal" requests and we shouldn't assume that a shtudown_status of `SigtermRequest` has been completely delivered.  (Assuming there were a means of checking the status - which I could see some admin-level apps wanting to do.)

Hmm - on that note, should we consider name changes to `TerminateRequest` and `KillRequest`.  These sound like kernel messages.  I suppose we could use `Request` as a _prefix_ on all three: `RequestShutdown`, `RequestTerminate`, and `RequestKill`.

jupyter_client/manager.py
```diff
@@ -340,10 +361,25 @@ def finish_shutdown(self, waittime=None, pollinterval=0.1):
...
+            # send sigterm and wait the other 50%.
+            self._shutdown_status = _ShutdownStatus.SigtermRequest
+            self._send_kernel_sigterm()
+            for i in range(int(waittime / 2 / pollinterval)):
+                if self.is_alive():
+                    time.sleep(pollinterval)
+                else:
+                    # If there's still a proc, wait and clear
+                    if self.has_kernel:
+                        self.kernel.wait()
```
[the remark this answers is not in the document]
Carreau: I don't think so, that's the same logic as  line 360; I could make it a local function and call it in both location to be clearer.

jupyter_client/manager.py
Carreau: Yeah, it's with the one just below, let me try to change the layout to be clearer.

jupyter_client/manager.py
Carreau: Done.

jupyter_client/manager.py
```diff
@@ -458,6 +496,31 @@ def has_kernel(self):
...
+        """similar to _kill_kernel, but with sigterm (not sigkill), but do not block"""
+        if self.has_kernel:
+            # Signal the kernel to terminate (sends SIGTERM on Unix and
+            # if the kernel is a subprocess and we are on windows; this is
+            # equivalent to kill
+            try:
+                if hasattr(self.kernel, "terminate"):
+                    self.kernel.terminate()
+                elif hasattr(signal, "SIGTERM"):
+                    self.signal_kernel(signal.SIGTERM)
```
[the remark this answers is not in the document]
Carreau: added an else with a log statement; I'm not too sure of cases where we can't send sigterm; 
That would likely be windows if the kernel is not a subprocess; in which case we won't send anything, but will send sigkill equivalent anyway after the full wait time.

jupyter_client/manager.py
Carreau: I've split the two try/except to make the intention clearer, is that better ?

jupyter_client/manager.py
```diff
@@ -673,6 +752,31 @@ async def restart_kernel(self, now=False, newports=False, **kw):
...
+        """similar to _kill_kernel, but with sigterm (not sigkill), but do not block"""
+        if self.has_kernel:
+            # Signal the kernel to terminate (sends SIGTERM on Unix and
+            # if the kernel is a subprocess and we are on windows; this is
+            # equivalent to kill
+            try:
+                if hasattr(self.kernel, "terminate"):
+                    self.kernel.terminate()
+                elif hasattr(signal, "SIGTERM"):
+                    await self.signal_kernel(signal.SIGTERM)
```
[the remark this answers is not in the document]
Carreau: done

jupyter_client/tests/signalkernel.py
```diff
@@ -25,6 +29,14 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.children = []
 
+        if os.environ.get("NO_SIGTERM_REPLY", None) == "1":
+            signal.signal(signal.SIGTERM, signal.SIG_IGN)
+
+    @gen.coroutine
+    def shutdown_request(self, stream, ident, parent):
+        if os.environ.get("NO_SHUTDOWN_REPLY") != "1":
+            yield gen.maybe_future(super().shutdown_request(stream, ident, parent))
```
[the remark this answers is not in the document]
Carreau: I don't believe so,  that's the exact way ipykernel uses coroutines:

```
 @gen.coroutine
    def shutdown_request(self, stream, ident, parent):
        content = yield gen.maybe_future(self.do_shutdown(parent['content']['restart']))
        ...
```

So if this breaks; IPykernel would be broken, and if it works with ipykernel, I see no reason this wouldn't. 

I would prefer `async def` and `await`, but that unfortunately does not mix.

jupyter_client/manager.py
```diff
@@ -507,6 +514,11 @@ def _send_kernel_sigterm(self):
                     self.kernel.terminate()
                 elif hasattr(signal, "SIGTERM"):
                     self.signal_kernel(signal.SIGTERM)
+                else:
+                    self.log.debug(
+                        "Cannot set term signal to kernel, no"
```
kevin-bates, line 519: I see we prefer `SIGKILL` over `kill()` and assume that `kill()` exists.  Should we do the same with `SIGTERM` and `terminate()`? (I suspect this is due to Windows not having `SIGTERM`?) 

Actually, given that `Popen` is well-documented to use `SIGKILL` and `SIGTERM` for `kill()` and `terminate()`, respectively, and there don't appear to be python version compatibility issues, I think we'd be fine just calling the desired methods directly.  This would then forgo the need to log anything and remove the conditional attribute checks.

kevin-bates (commented):
Thanks for the update Matthias - they look good.

I just had the one comment regarding possibly cutting to the chase and using the subprocess methods unconditionally.  There might be some history here that I'm not seeing.  Thanks.

Carreau:
> I don't think this is anything to change, but just noting that the shutdown-request status is set after the request, while the signal-based statuses are set prior to their respective requests. Although not consistent, I think it's probably the right approach across the three and I'm not advocating for additional (pre/post) statuses, but with Kernel Provisioning, the `terminate()` and `kill()` are not necessarily "signal" requests and we shouldn't assume that a shtudown_status of `SigtermRequest` has been completely delivered. (Assuming there were a means of checking the status - which I could see some admin-level apps wanting to do.)
> 
> Hmm - on that note, should we consider name changes to `TerminateRequest` and `KillRequest`. These sound like kernel messages. I suppose we could use `Request` as a _prefix_ on all three: `RequestShutdown`, `RequestTerminate`, and `RequestKill`.

I do not have any preference, this is mostly for testing in order to understand the internal state and why I kept them private. ; I initially thought of returning the values but I'm not feeling confident with making these public.

jupyter_client/manager.py
Carreau: Diggning in:

I don't think that's compatibility issue, signal_kernel signal the process group, not the kernel alone.
Sigterm can be handled by a subprocess, not sigkill.

Signal_kernel send the signal to the process group (when possible) instead of only the single process, so if the kernel has children you don't want to send sigterm to the (grand)children, put leave the child process a chance to clean up it children, in case there is some cleanup logic.

Thus `terminate()` is prefered, and if you can't you sigterm  the process group. 

Kill is the opposite, as the kernel has no chance to terminate children by catching sigkill, you want to tell them to die  as well. See #314 

I guess in a perfect world you would
 - 1 shutdown request
 - 2 wait
 - 3 term child
 - 4 wait
    -  4b if don't respond term grand children
 - 5 wait
     - 5b kill everybody. 

we just skip 4,4b, and replace 3 by "term child, if you can't term everybody".

kevin-bates: Good point on the process group portion of things.  I had conflated `signal_kernel()` with `send_signal()`.

Carreau: Yeah, i I had to dig deep in the code and history to figure that one :-)

kevin-bates (approved):
Thank you Matthias (and @mlucool for proposing these changes) - this makes a kernel's management more robust.

mlucool:
Thanks @Carreau - I tested this and it does fix https://github.com/jupyter/jupyter_client/issues/618

Carreau:
there has been no objecting in a week; merging. Will do a release soon.

Issue #533, opened 2020-03-19
Refactor BlockingKernelClient


davidbrochart:
There is a lot of code duplication between `AsyncKernelClient` and `BlockingKernelClient`. `BlockingKernelClient`'s methods could be wrappers around `AsyncKernelClient`'s methods, which would run the later until complete.
What do you think?

MSeal:
I think that'd be a good idea, but maybe let's hold off making this low level infrastructure do that until 3.5 gets dropped? I'm a little worried with the hangs for 3.5 we saw in nbconvert after async change -- that one is doing a major version bump so I'm more ok with it but here we might want to wait for a major version change / 3.5 being gone before taking that risk on?

davidbrochart:
Sure, that sounds reasonable.

blink1073:
@davidbrochart is there any more work to be done on this front for 7.0?

davidbrochart:
No, it has been done in #623. Closing!

PR #623, opened 2021-03-08
Add type annotations, refactor sync/async

### LINKED ISSUE #553: Type Annotations
Now there we're in Python 3 only, and given that we use classes in this repo as API contracts, I think it's time we look into adding type annotations to our code. This would help with clarifying the contract expectations and with avoiding subtle argument consumption changes causing inherited classes outside of the repo to fail.

### LINKED ISSUE #621: AsyncKernelClient.is_alive() returns non-awaited coroutine?
[async is_alive()](https://github.com/jupyter/jupyter_client/blob/987f3e1de21e34d7f6f8b3225a6b044e1bdb0152/jupyter_client/asynchronous/client.py#L244) can return:
1) [an awaited coroutine](https://github.com/jupyter/jupyter_client/blob/987f3e1de21e34d7f6f8b3225a6b044e1bdb0152/jupyter_client/asynchronous/client.py#L251) if it was created by an `AsyncKernelManager`.
2) or [its parent's is_alive()](https://github.com/jupyter/jupyter_client/blob/987f3e1de21e34d7f6f8b3225a6b044e1bdb0152/jupyter_client/asynchronous/client.py#L252) if it was created by a `KernelManager` which is not `AsyncKernelManager`.
3) or [the heartbeat's is_beating()](https://github.com/jupyter/jupyter_client/blob/987f3e1de21e34d7f6f8b3225a6b044e1bdb0152/jupyter_client/asynchronous/client.py#L256) if we don't have access to the kernel manager that created it but we have a heartbeat.
4) or [True](https://github.com/jupyter/jupyter_client/blob/987f3e1de21e34d7f6f8b3225a6b044e1bdb0152/jupyter_client/asynchronous/client.py#L260) if we don't even have a heartbeat.

Cases 1. and 4. are fine, but 2. and 3. could potentially return a non-awaited coroutine. This is causing problems e.g. in https://github.com/jupyter/nbclient/pull/138, but I think the main issue is that we don't know the objects that we are manipulating any more, because we are trying to have a single code base that works for async and non-async. I also think that we are allowing for too much freedom, for instance mixing a non async kernel manager and an async kernel client. This is becoming too complicated.
I think adding type annotations, as @MSeal suggested in #553, will help us understand and improve the situation.

davidbrochart:
Fixes #553 
Fixes #621


jupyter_client/manager.py
```diff
@@ -55,39 +58,45 @@ def __init__(self, *args, **kwargs):
...
         self._shutdown_status = _ShutdownStatus.Unset
 
-    _created_context = Bool(False)
+    _created_context: Bool = Bool(False)
 
     # The PyZMQ Context to use for communication with the kernel.
-    context = Instance(zmq.Context)
-    def _context_default(self):
+    context: Instance = Instance(zmq.Context)
+    def _context_default(self) -> zmq.Context:
```
davidbrochart, line 65: I was surprised not to see the `defaults` decorator here:
```python
    @defaults('context')
```
Am I missing something?

jupyter_client/manager.py
```diff
@@ -55,39 +58,45 @@ def __init__(self, *args, **kwargs):
...
         self._created_context = True
         return zmq.Context()
 
     # the class to create with our `client` method
-    client_class = DottedObjectName('jupyter_client.blocking.BlockingKernelClient')
-    client_factory = Type(klass='jupyter_client.KernelClient')
-    def _client_factory_default(self):
+    client_class: DottedObjectName = DottedObjectName('jupyter_client.blocking.BlockingKernelClient')
+    client_factory: Type = Type(klass='jupyter_client.KernelClient')
+    def _client_factory_default(self) -> Type:
```
davidbrochart, line 72: No `@defaults('client_factory')` here?

jupyter_client/manager.py
```diff
@@ -55,39 +58,45 @@ def __init__(self, *args, **kwargs):
...
     # The kernel process with which the KernelManager is communicating.
     # generally a Popen instance
-    kernel = Any()
+    kernel: Any = Any()
 
-    kernel_spec_manager = Instance(kernelspec.KernelSpecManager)
+    kernel_spec_manager: Instance = Instance(kernelspec.KernelSpecManager)
 
-    def _kernel_spec_manager_default(self):
+    def _kernel_spec_manager_default(self) -> kernelspec.KernelSpecManager:
```
davidbrochart, line 88: No `@defaults('kernel_spec_manager')` here?

davidbrochart:
I don't know how to add type hints to traits, right now this doesn't have any effect (see https://github.com/ipython/traitlets/issues/647).

davidbrochart:
The fact that we are trying to support sync/async in a single code base is actually not good for type hints. For instance, a `KernelManager` has a [`_kill_kernel()` method](https://github.com/jupyter/jupyter_client/blob/b4d35ee5217789ca9b25ed6b35e23897d483df02/jupyter_client/manager.py#L602) that should return a bool. But then, an `AsyncKernelManager`, which derives from `KernelManager`, redefines [`_kill_kernel()` as an async method](https://github.com/jupyter/jupyter_client/blob/b4d35ee5217789ca9b25ed6b35e23897d483df02/jupyter_client/manager.py#L866). If we try to type annotate the return value of the latter, we get the following mypy error:
```
Return type "Coroutine[Any, Any, None]" of "_kill_kernel" incompatible with return type "None" in supertype "KernelManager"
```
I posted [a question on Stack Overflow](https://stackoverflow.com/questions/66532718/python-type-hints-and-redefining-method-as-async) about that, it seems that this is a legit error. I'm not sure what to do about that.

davidbrochart:
In #533 we suggested to have a single, async-native code base for jupyter-client, as we did in nbclient. The non-async methods would be wrappers around the async methods, and run them until complete. This would probably solve the issue I mentioned above, and it will also remove some duplicated code.
Any thoughts, should we do that?

davidbrochart:
Thanks a lot @dhirschfeld, I didn't know this project.
It seems to be based on code generation, which is actually quite nice since it allows to have a single (async) code base, without the problems of translating it to blocking code at run-time, since it's done ahead of time.
I'll look at it in greater details, thanks again.

dhirschfeld:
The folks on the `trio` gitter channel are a very helpful bunch if you have questions about adapting it to your requirements.
https://gitter.im/python-trio/general

davidbrochart:
Thanks, I was worried it would be only for `trio`-specific async code, but it seems to apply to general async code, since it basically just removes `async`, `await`, and changes `__aenter__` to `__enter__` and stuff like that, so that's great.

davidbrochart:
This last commit refactors `KernelManager`.
Previously `KernelManager` was the blocking KM, and `AsyncKernelManager` derived from it, redefining some methods as `async`. Apart from duplicating a lot of code, this prevented to type-annotate these methods.
Now, `KernelManager` is a base class from which `BlockingKernelManager` and `AsyncKernelManager` derive from. `KernelManager` only has the async version of the methods (single source of truth), and `BlockingKernelManager` wraps and run them until complete in an event loop. This event loop runs in a separate thread to avoid issues with an already running event loop.

kevin-bates:
Isn't this going to break direct subclasses of `KernelManager` (besides `AsyncKernelManager`)?  

Since we want to eventually have one implementation (async), what is the downside of a _split-and-deprecate_ approach while using inheritance purely as the means to support configuration in a backwards-compatible manner?  Yes, it means some duplication of code, but it doesn't stifle innovation, is much simpler to support and is temporary.

davidbrochart:
> Isn't this going to break direct subclasses of `KernelManager` (besides `AsyncKernelManager`)?

Yes, previous subclasses of `KernelManager` will have to subclass `BlockingKernelManager` instead.

> Since we want to eventually have one implementation (async), what is the downside of a _split-and-deprecate_ approach while using inheritance purely as the means to support configuration in a backwards-compatible manner? Yes, it means some duplication of code, but it doesn't stifle innovation, is much simpler to support and is temporary.

I'm not sure to follow what you mean, could you explain further?

kevin-bates:
> Yes, previous subclasses of `KernelManager` will have to subclass `BlockingKernelManager `instead.

OK. I guess we're talking about a major-release boundary here (e.g., 7.0) where any subclasses not updated or who haven't capped `jupyter_client < 7.0` will have their consuming applications break.

>> Since we want to eventually have one implementation (async), what is the downside of a _split-and-deprecate_ approach while using inheritance purely as the means to support configuration in a backwards-compatible manner? Yes, it means some duplication of code, but it doesn't stifle innovation, is much simpler to support, and is temporary.

> I'm not sure to follow what you mean, could you explain further?

By _split-and-deprecate_ I mean we do not have `AsyncKernelManager` invoke `super().method-name` on _any_ methods - thereby 'splitting' the functional offering of each of the two classes.  We also deprecate all `KernelManager` _methods_ (not the class) with a timeframe for their removal.

Because the implementations are split and because we want to continue to innovate, new functionality - like [Kernel Provisioning](https://github.com/jupyter/jupyter_client/issues/608) - can move forward on the async class while leaving the synchronous methods alone and not requiring proliferation of both behaviors - one of which we plan to abandon.  Once we determine the appropriate deprecation cycle, we remove the `KernelManager` methods.  But, until then, those applications that have not adopted the more modern technology will continue to work as they do today - no changes required.

We also wouldn't be introducing a new paradigm that runs every method of `KernelManager` on a separate thread.  (There don't appear to be tests using the `IOLoopKernelManager` and `IOLoopKernelRestarter` subclasses - will they be affected - other than updating the former's superclass? )

I think we could use a one-time (or semi-regularly) _maintainer's sync-up_ of sorts - open to all - where we discuss these issues and a roadmap, in general.  Would that be something worthwhile to others?

davidbrochart:
Regarding breaking the API by renaming the current `KernelManager` to `BlockingKernelManager`, I think I can avoid that. I'm going to push a commit.
I'm still not sure we want to eventually get rid of the sync behavior. Async has advantages indeed, but it forces users to enter a whole new paradigm. To me, if we can offer both APIs without much effort, then we should.

jupyter_client/manager.py
```diff
@@ -438,16 +485,16 @@ def shutdown_kernel(self, now=False, restart=False):
         # Stop monitoring for restarting while we shutdown.
         self.stop_restarter()
 
-        self.interrupt_kernel()
+        await self._async_interrupt_kernel()
```
kevin-bates, line 488: This breaks subclasses that override interrupt_kernel() as they won't have their override called.

jupyter_client/manager.py
```diff
@@ -438,16 +485,16 @@ def shutdown_kernel(self, now=False, restart=False):
...
 
         if now:
-            self._kill_kernel()
+            await self._async__kill_kernel()
```
kevin-bates, line 491: This breaks subclasses that override _kill_kernel() as they won't have their override called.

jupyter_client/manager.py
```diff
@@ -438,16 +485,16 @@ def shutdown_kernel(self, now=False, restart=False):
...
         else:
             self.request_shutdown(restart=restart)
             # Don't send any additional kernel kill messages immediately, to give
             # the kernel a chance to properly execute shutdown actions. Wait for at
             # most 1s, checking every 0.1s.
-            self.finish_shutdown()
+            await self._async_finish_shutdown()
```
kevin-bates, line 497: This breaks subclasses that override finish_shutdown() as they won't have their override called.

jupyter_client/manager.py
```diff
@@ -500,21 +554,23 @@ def restart_kernel(self, now=False, newports=False, **kw):
                                "No previous call to 'start_kernel'.")
         else:
             # Stop currently running kernel.
-            self.shutdown_kernel(now=now, restart=True)
+            await self._async_shutdown_kernel(now=now, restart=True)
```
kevin-bates, line 557: This breaks subclasses that override shutdown_kernel() as they won't have their override called during restarts.

kevin-bates (changes requested):
I think the same issues will apply for KernelClient as well.  Generally speaking, any methods called from within another method will not have their overridden methods called on subclasses.  The aliasing trick breaks subclasses.

davidbrochart:
> I think the same issues will apply for KernelClient as well. Generally speaking, any methods called from within another method will not have their overridden methods called on subclasses. The aliasing trick breaks subclasses.

There should be a test for that.

davidbrochart:
I think we should be more explicit as to what is async and what is sync. If a subclass of `KernelManager` is async, then it should override the methods prefixed with `_async_`, e.g. `_async__launch_kernel`. If is sync, then it should override e.g. `_launch_kernel`.
Maybe even the API of a kernel manager should be different, depending on whether it is async or not. I'm afraid a lot of people will forget to `await` the async API, if it has the same name as the sync API.

kevin-bates:
> I'm still not sure we want to eventually get rid of the sync behavior. Async has advantages indeed, but it forces users to enter a whole new paradigm. To me, if we can offer both APIs without much effort, then we should.

Deprecation cycles can take years ([`kernel_cmd`](https://github.com/jupyter/jupyter_client/blame/f169423a430b4f6caf99f19920f04ef60ceca89d/jupyter_client/manager.py#L132) was deprecated 7 years ago).  I'm just saying, to reasonably move forward, we should consider splitting the class implementations.  As you're finding out, once subclasses exist in nature, it is extremely difficult to do anything.  I'd rather we innovate on the async classes, allow folks to adopt the new innovations (or remain where they are if they desire), and move forward with async while the sync classes gather dust (but work).

kevin-bates:
> I think we should be more explicit as to what is async and what is sync. If a subclass of KernelManager is async, then it should override the methods prefixed with _async_, e.g. _async__launch_kernel. If it is sync, then it should override e.g. _launch_kernel.
Maybe even the API of a kernel manager should be different, depending on whether it is async or not. I'm afraid a lot of people will forget to await the async API, if it has the same name as the sync API.

I agree - in a world where we don't have to continue supporting clients, this is all true. The reality is we have a contract to uphold and introducing this level of change breaks that contract - and, yes, it makes things extremely difficult to change.  

If we're saying that we're free to completely rework our public contract whenever (and subclass authors be damned), then that's a major violation of trust.

davidbrochart:
Regarding the sync API deprecation, I meant that I'm not sure it should happen at all. Having an async-only API is a pretty big constraint for users, as once they use it async/await propagates to their entire code base. I think a sync API is very useful and should stay forever.
I wouldn't say we are breaking a contract with clients, it's just the life of software: APIs change and sometimes there are backward-incompatible changes. BTW, overriding a private method in a class doesn't look like a good practice, and this may be a good time to improve this API, and document it.

kevin-bates:
> Regarding the sync API deprecation, I meant that I'm not sure it should happen at all. 

Yes, I know.

> Having an async-only API is a pretty big constraint for users, as once they use it async/await propagates to their entire code base. 

Agreed.
> I think a sync API is very useful and should stay forever.

I think it would be good to discuss this.  I think we're going to find it hard to adequately support both, but that depends on the kinds of changes we'll be making.  I guess with this approach, new functionality can be added _within_ the dual-class structures - so long as it is async.  Is that a correct understanding?  If so, that is great.

> I wouldn't say we are breaking a contract with clients, it's just the life of software: APIs change and sometimes there are backward-incompatible changes. 

Because we have documented, promoted, and allowed folks to subclass KernelManager and KernelClient - we _are_ breaking a contract but that's what major release boundaries are for.  I had proposed in the Kernel Provisioning proposal that it be delivered at a major boundary due to a _hypothetical_ contract breakage (that has been extremely difficult to avoid - **because it's a contract**). I'm just not sure the addition of type hints warrants a major release.

> BTW, overriding a private method in a class doesn't look like a good practice, and this may be a good time to improve this API, and document it.

Actually, subclass overrides of _private-callable_ methods is _exactly_ how you allow subclasses to influence their behavior on the class.  It's important to distinguish between _publically-callable_ methods - which is the _contract_ for client _applications_ and _privately-callable_ methods - which amends the contract promised to clients _extending_ your functionality.

From what I've read in Python, a class that is intended to be subclassed would use double-underscore prefixes on method names to indicate those methods are truly private (and triggers the name-mangling).  Single-underscored methods are more like `protected` methods in Java and can be overridden.  I added a number of these kinds of methods in #612 to allow third-parties to hook into things like parameter validation, setting up env, etc.  It's actually a _best practice_ because you're saying these methods are not callable from outside the package.  The base framework calls these methods to drive the customization.

We even document that [subclasses override `_launch_kernel()`](https://github.com/jupyter/jupyter_client/blob/f169423a430b4f6caf99f19920f04ef60ceca89d/jupyter_client/manager.py#L247) to "launch kernel subprocesses differently".

I'm sorry for being so persistent about this and I do appreciate the discussion (and it's helping me see things). Thanks for spending time on this (and patience with me :smile: ).

Would you mind adding some description to the PR's description of the approach used (i.e., all sync methods are essentially placed onto a separate thread/event loop) so others don't need to glean this from the changes?  You should also mention the API breakage so we can discuss what our release cycle would be if this were to be merged.  If we're going to be introducing a major release - there are others we'll want to do like finally remove the 7-year-deprecated `kernel_cmd`, the old `cleanup()` method, etc. I'd also like to see Kernel Provisioning and [Kernel Handshaking](https://github.com/jupyter/enhancement-proposals/pull/66) make it into the next major release update as well.

davidbrochart:
Thanks Kevin for the detailed explanation, and relevant background (as usual, and this is very appreciated as I lack those).
This PR is very much a work in progress, so don't worry if you see me breaking everything. I probably need to do so for my own understanding, but will probably have to put some things back in place for backward compatibility. Thanks anyway for the guidance in this process.
I will reply more in details later.

davidbrochart:
I think this PR is ready for review. I'll continue polishing it but I believe it represents a solid ground for where we want to take jupyter-client to. I'm going to give a summary of the changes:
- I added type annotations, especially in the kernel managers and the kernel clients.
- During this process, I realized that the way managers and clients were designed didn't allow to type annotate them. Indeed we had async versions derive from sync versions and override sync functions with async ones. I also noted that sync and async supposedly equivalent code had started to diverge. So I refactored managers and clients so that there is only one source of truth. Now, the code is inherently async, and used in sync and async versions of managers and clients. The sync API runs the async code until complete using nest-asyncio.
- When there was a sync and an async version of a function, I kept the async one. This might have broken e.g. #620 and #627. We can discuss these, as I think the sync and async behaviors were not in phase.

Whatever we decide for the future (deprecate the sync API or not), I believe having an async-first code base from which the sync code derives is the best option, as it allows us to innovate on the asynchronous programming side (which will spread all over the Jupyter ecosystem in the future), while not wasting efforts maintaining the sync API (since it comes for free).

cc @kevin-bates @MSeal @minrk @SylvainCorlay @Carreau @vidartf @blink1073 @jasongrout @JohanMabille

jupyter_client/tests/test_multikernelmanager.py
```diff
@@ -206,10 +200,10 @@ def test_subclass_callables(self):
         km.get_kernel(kid).reset_counts()
         km.reset_counts()
         km.shutdown_all(now=True)
-        assert km.call_count('shutdown_kernel') == 0
+        #assert km.call_count('shutdown_kernel') == 0
```
kevin-bates, line 203: Since we're now _async-first_, I think going with the expected sequence in the async test is the correct approach.  This does mean that the framework no longer calls `request_shutdown` and `finish_shutdown` on {A}MKM classes, so we might want to discuss their deprecation.

jupyter_client/tests/test_multikernelmanager.py
```diff
@@ -256,7 +250,7 @@ async def _run_lifecycle(km, test_kid=None):
         assert kid in km.list_kernel_ids()
         await km.interrupt_kernel(kid)
         k = km.get_kernel(kid)
-        assert isinstance(k, KernelManager)
+        assert isinstance(k, AsyncKernelManager)
```
kevin-bates, line 253: Good catch!

jupyter_client/tests/test_multikernelmanager.py
```diff
@@ -463,10 +449,10 @@ async def test_subclass_callables(self):
         mkm.get_kernel(kid).reset_counts()
         mkm.reset_counts()
         await mkm.shutdown_all(now=True)
-        assert mkm.call_count('shutdown_kernel') == 1
+        #assert mkm.call_count('shutdown_kernel') == 1
```
kevin-bates, line 452: Per the previous comment, I think enabling these commented-out assertions is the way to go.

kevin-bates (approved):
@davidbrochart - this is excellent!  For you to be able to preserve the integrity of the subclass implementations is outstanding!  Because the older sync method "comes for free", I'm good with us dragging it along vs. its eventual deprecation.  This is an ideal outcome.

One problem with allowing subclassing is that we really don't know who has done what wrt to method overrides.  I've run these changes on an EG server where we override portions of both MultiKernelManager and KernelManager and see no issues.  I think that given the recently added subclass tests are passing without changes[*] and the practical example of EG appears unaffected, indicates "all systems go" relative to the manager classes! :rocket:

I'm not as familiar with the KernelClient classes, but it looks like the same approach was applied there as well - correct?

[*]: I also noticed the disparity between the `MultiKernelManager` and `AsyncMultiKernelManager` `shutdown_all()` methods while working on #627 and believe erring on the side of `AsyncMultiKernelManager`'s approach is correct.  As a result, I'm not sure how useful the MKM `request_shutdown()` and `finish_shutdown()` methods are since they are no longer invoked by the framework.  Some applications may be using them at the MKM level, but those applications would also bypassing the logic to [clean up the cached ports](https://github.com/jupyter/jupyter_client/pull/623/files#diff-3016997220ab587d3300d7d5fa8172db3797ea81e05726097e97494da5d68432R263-L216) since the `shutdown_kernel()` method is skipped.

Carreau:
Thanks for writing this; 
If my understanding is correct, `run_sync` should never reenter; right ? The async methods always call the `_async_xxx` versions, and the sync class just expose public `run_sync(...)` ? 

If that's the case; when I write this kind of things I tend to add a check in `run_sync` that it has not re-entered; and emit a warning when this is the case with "did you meant to call the _async_xxx instead of xxx". This is useful for people subclassing.

Also nice use of async_io and nesteed_asyncio; I tend to use a dummy coroutine runner but I guess the code here requires asyncio because of sockets/tornado.

davidbrochart:
> If my understanding is correct, `run_sync` should never reenter; right ? The async methods always call the `_async_xxx` versions, and the sync class just expose public `run_sync(...)` ?

`run_sync` can actually reenter. I wanted to avoid that, and have the async methods always call the `_async_xxx` versions as you mentioned, but that breaks the subclass implementations. Indeed the "official" API for subclasses is to override methods, some of them can be private.

> Also nice use of async_io and nesteed_asyncio; I tend to use a dummy coroutine runner but I guess the code here requires asyncio because of sockets/tornado.

We cannot use a dummy coroutine runner because we actually use some of asyncio's features, to gather tasks or to run a task with a timeout.

davidbrochart:
@kevin-bates 

> I've run these changes on an EG server where we override portions of both MultiKernelManager and KernelManager and see no issues.

I've also tested with nbclient and everything was fine. I intended to do more real-world testing before merging.

> I'm not as familiar with the KernelClient classes, but it looks like the same approach was applied there as well - correct?

Yes, the same approach was applied for the kernel managers and kernel clients.

> [*]: I also noticed the disparity between the `MultiKernelManager` and `AsyncMultiKernelManager` `shutdown_all()` methods while working on #627 and believe erring on the side of `AsyncMultiKernelManager`'s approach is correct. As a result, I'm not sure how useful the MKM `request_shutdown()` and `finish_shutdown()` methods are since they are no longer invoked by the framework. Some applications may be using them at the MKM level, but those applications would also bypassing the logic to [clean up the cached ports](https://github.com/jupyter/jupyter_client/pull/623/files#diff-3016997220ab587d3300d7d5fa8172db3797ea81e05726097e97494da5d68432R263-L216) since the `shutdown_kernel()` method is skipped.

Yes, I need to investigate in more details how different both (sync and async) implementations are, and come up with a mental model. But see, it was time to unify the code again, as it has started to diverge quite a lot and it was like having two branches to maintain in the same repository :smile:

Carreau:
> `run_sync` can actually reenter. I wanted to avoid that, and have the async methods always call the `_async_xxx` versions as you mentioned, but that breaks the subclass implementations. Indeed the "official" API for subclasses is to override methods, some of them can be private.

Make sens ! Thanks, good work.

davidbrochart:
I should mention that I skipped the following tests on OSX for Python 3.6 and 3.7:
- `test_start_parallel_process_kernels` in `test_kernelmanager.py`
- `test_start_sequence_process_kernels` in `test_kernelmanager.py`.
- `test_start_parallel_process_kernels` in `test_multikernelmanager.py`.

Because of a `Bad file descriptor` error. It seems to be [a known bug](https://groups.google.com/g/python-tornado/c/DkXjSNPCzsI).

davidbrochart:
I will squash and merge today if there is no feedback.

kevin-bates:
> I will squash and merge today if there is no feedback.

I'd like to test another scenario using Elyra/EG in which both kernel manager and client are subclassed.  I'll post back when completed (hopefully within a couple of hours).

kevin-bates:
@davidbrochart - I'm sorry this took so long - it was one of those days - and this took several hours itself.

I'm seeing an issue when running with this branch but not with `6.1.12`.  I've narrowed things down to [this line in papermill](https://github.com/nteract/papermill/blob/main/papermill/clientwrap.py#L46) - although there's plenty going on behind that statement.

I'm running an [Elyra](https://github.com/elyra-ai/elyra) pipeline - where each node is a notebook that is [executed by papermill](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_local.py#L173-L180). This all works fine against `jupyter_client <= 6.1.12`.  However, when running against this branch, everything is also fine until the second kernel-info request immediately following the cell executions of the first notebook.  When I set a breakpoint in [`KernelClient._async_kernel_info`](https://github.com/jupyter/jupyter_client/blob/5774be3d2557dc0caa8a7afcc8a3ad9b60fc3e34/jupyter_client/client.py#L750), the first kernel-info request (during [`async_start_new_kernel_client`](https://github.com/jupyter/nbclient/blob/master/nbclient/client.py#L384)) is hit, but [the second, when called from papermill](https://github.com/nteract/papermill/blob/main/papermill/clientwrap.py#L46) is not.  I never get to the next line (L47).

I'm using nbclient==0.5.3 and papermill==2.3.3.

((FWIW, I had spent hours today with a much more complicated configuration in which I was using KM and KC subclasses that redirect to a gateway server (also from Elyra).  Also in that scenario, the first notebook doesn't complete with the same issue.  So I (just now) decided to try things with simple, local only, notebooks and servers, and its the same behavior - which implies something more fundamental and less about the class hierarchy changes - if that helps at all.))

davidbrochart:
Thanks Kevin for taking the time to test and investigate. I could reproduce the bug in papermill's tests.
The issue comes from the fact that papermill, which uses nbclient under the hood to execute a notebook, uses an `AsyncKernelClient`. Some requests, such as `kernel_info()`, can wait for a reply when they are passed `reply=True`, or not (`reply=False`). Because the `reply` argument is passed at run-time, `kernel_info()` must be able to return either a message id (`reply=False`) or the reply message (`reply=True`), and thus must be an async function. Because the kernel client is async, the reply message has to be awaited (if the kernel client was blocking, we wait for the reply synchronously for you).
So [this line in papermill](https://github.com/nteract/papermill/blob/main/papermill/clientwrap.py#L46) has to be replaced with:
```python
info_msg = self.wait_for_reply(await self.kc.kernel_info())
```
This is not without consequences, since now the whole function call stack leading to this line has to be async. If this is unacceptable, then `kernel_info()` can be called synchronously:
```python
info_msg = self.wait_for_reply(run_sync(self.kc.kernel_info()))
```

davidbrochart:
I managed to fix the issue by having requests return synchronously when `reply=False` and asynchronously when `reply=True`, with an `AsyncKernelClient`. For a `BlockingKernelClient`, they always return synchronously.
Papermill's test suite pass with this change, could you give it a try @kevin-bates ?

kevin-bates:
Thanks David.  I don't see a relevant issue or pull request in Papermill for this - do you know if something is in the works?

I guess this raises some concerns regarding breaking existing clients.  However, I would imagine the only clients it would break would be those using `AsyncKernelClient` that do not call `kc.kernel_info()` synchronously - which are probably few.  I'm assuming Voila is not experiencing this.

I thought I tried the synchronous call approach yesterday (using `run_sync` from `nbclient`) and it didn't work, but if that could be used by a client of AKC, then I suppose we could go forward here.  (I just tried again using `jupyter_client`'s `run_sync` and see the same (hang) issue.  Are you able to get that to work?)

The alternative of forcing clients to plumb their call stack to be async may be too much to ask.  (Not saying that Papermill wouldn't do that, just talking about clients in general.)

I see you've just posted an update while I was writing this response.  I'll give this a shot.  Thanks!

kevin-bates:
Works like a charm @davidbrochart - very nice!  I also tried the more complicated scenario I referenced in which AKM and AKC subclasses come into play - ditto - like a charm!

davidbrochart:
Awesome, thanks a lot for your help Kevin.
Are we ready to squash and merge?

kevin-bates:
> Are we ready to squash and merge?

I am - although I would like to see at least another maintainer review this due to its significant changes.  

Thank you for all the work (and patience) on this - this is really outstanding (IMHO).  You've essentially flipped the script by making async the basis on which we can move forward!

davidbrochart:
I will merge today, if no objection.

blink1073 (approved):
Nice work!

blink1073:
:tada:

JohanMabille:
:rocket:

ccordoba12:
Hey guys, this seems like a huge refactoring to be included in a patch release (6.1.**13**). Furthermore, it included breaking changes (see https://github.com/jupyter/qtconsole/issues/476 and https://github.com/spyder-ide/spyder/issues/15161), so a heads up to upstream projects would be nice for next time.

davidbrochart:
Hi Carlos, although there has been big internal changes, I was not expecting any change for the user. But it looks like I was wrong, sorry for the inconvenience.
I'm waiting for someone with PyPI admin rights to yank the release (it's already marked as broken on conda-forge), and I will release 6.2.

jupyter_client/channels.py
```diff
@@ -34,24 +38,27 @@ class HBChannel(Thread):
...
-    poller = None
+    time_to_dead: float = 1.
     _running = None
     _pause = None
     _beating = None
 
-    def __init__(self, context=None, session=None, address=None, loop=None):
+    def __init__(
+        self,
+        context: zmq.asyncio.Context,
```
kevin-bates, line 53: It looks like the removal of the `None` default is what is breaking `qtconsole`, although this does imply they must not be using the HB channel since it unconditionally references through `self.context` in its `_create_socket()` method.

ccordoba12:
> I'm waiting for someone with PyPI admin rights to yank the release (it's already marked as broken on conda-forge), and I will release 6.2.

Thanks for your help @davidbrochart!

jupyter_client/channels.py
ccordoba12: > It looks like the removal of the None default is what is breaking qtconsole

Thanks for your help @kevin-bates. I found that too but now we have another problem (see below).

> although this does imply they must not be using the HB channel since it unconditionally references through self.context in its _create_socket() method.

It has something to do with the way the channel inherits from HBChannel and QObject and a re-implementation of `super` to make that work (I really don't understand that logic, @minrk implemented it).

ccordoba12: This is the error I'm seeing now while starting a kernel:

```python
ERROR:tornado.general:Uncaught exception in ZMQStream callback
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 434, in _run_callback
    callback(*args, **kwargs)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/threaded.py", line 101, in _handle_recv
    ident, smsg = self.session.feed_identities(msg)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/session.py", line 954, in feed_identities
    idx = msg_list.index(DELIM)
AttributeError: '_asyncio.Future' object has no attribute 'index'
ERROR:tornado.general:Uncaught exception in zmqstream callback
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 451, in _handle_events
    self._handle_recv()
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 480, in _handle_recv
    self._run_callback(callback, msg)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 434, in _run_callback
    callback(*args, **kwargs)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/threaded.py", line 101, in _handle_recv
    ident, smsg = self.session.feed_identities(msg)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/session.py", line 954, in feed_identities
    idx = msg_list.index(DELIM)
AttributeError: '_asyncio.Future' object has no attribute 'index'
ERROR:asyncio:Exception in callback BaseAsyncIOLoop._handle_events(30, 1)
handle: <Handle BaseAsyncIOLoop._handle_events(30, 1)>
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/tornado/platform/asyncio.py", line 189, in _handle_events
    handler_func(fileobj, events)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 451, in _handle_events
    self._handle_recv()
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 480, in _handle_recv
    self._run_callback(callback, msg)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 434, in _run_callback
    callback(*args, **kwargs)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/threaded.py", line 101, in _handle_recv
    ident, smsg = self.session.feed_identities(msg)
  File "/home/carlos/.virtualenvs/test-jupyter-client/lib/python3.8/site-packages/jupyter_client/session.py", line 954, in feed_identities
    idx = msg_list.index(DELIM)
AttributeError: '_asyncio.Future' object has no attribute 'index'
```

Any ideas?

davidbrochart: I also saw this @ccordoba12, it has to do with the threaded channels. Don't worry, I will clear all that out before the next release.

ccordoba12: Great! Please give me a ping in case you want me to test your changes.

MSeal:
6.1.13 is yanked on PyPI now as well

vidartf:
Just as an FYI, this change also breaks nbconvert 5.x branch, as it has no async wrapping around its [`get_msg` call](https://github.com/jupyter/nbconvert/blob/3210bf7b8f55b5174c1367e3a539665799d0acb5/nbconvert/preprocessors/execute.py#L478) etc:

davidbrochart:
You mean that it was reported in #637? And #638 should fix it.

PR #626, opened 2021-03-13
Add tests to ensure KernelManager subclass methods are called

kevin-bates:
So that we can better ensure that pending changes don't side-effect `KernelManager` subclass implementations, this change adds tests that assert particular methods are called on subclasses that override those methods.  Failures in these tests will likely indicate a breakage in the API contract.


davidbrochart (approved):
That's great, thanks Kevin.

kevin-bates:
Thanks for the review David.  I'm going ahead with the merge since a) its _just test code_ and b) it will help us assess the impact of changes - intentional or not.

PR #627, opened 2021-03-17
Add tests to ensure MultiKernelManager subclass methods are called

kevin-bates:
This is a follow-on to #626 to check subclass call sequences relative to the MultiKernelManager base classes.

Also found that CI was skipping the async tests of the recently added `TestKernelManagerShutDownGracefully` resulting in these warnings:
```
jupyter_client/tests/test_kernelmanager.py::TestKernelManagerShutDownGracefully::test_async_signal_kernel_subprocesses[signaltest-_install_kernel-_ShutdownStatus.ShutdownRequest]
jupyter_client/tests/test_kernelmanager.py::TestKernelManagerShutDownGracefully::test_async_signal_kernel_subprocesses[signaltest-no-shutdown-install_kernel_dont_shutdown-_ShutdownStatus.SigtermRequest]
jupyter_client/tests/test_kernelmanager.py::TestKernelManagerShutDownGracefully::test_async_signal_kernel_subprocesses[signaltest-no-terminate-install_kernel_dont_terminate-_ShutdownStatus.SigkillRequest]
  /Users/runner/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/_pytest/python.py:172: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))
``` 
which was addressed by adding the `@pytest.mark.asyncio` decorator.

### DIFF
--- a/jupyter_client/tests/utils.py
+++ b/jupyter_client/tests/utils.py
@@ -42,6 +42,7 @@ class test_env(object):
     def __exit__(self, *exc_info):
         self.stop()
 
+
 def execute(code='', kc=None, **kwargs):
     """wrapper for doing common steps for validating an execution request"""
     from .test_message_spec import validate_message
@@ -62,6 +63,17 @@ def execute(code='', kc=None, **kwargs):
     return msg_id, reply['content']
 
 
+def subclass_recorder(f):
+    def wrapped(self, *args, **kwargs):
+        # record this call
+        self.record(f.__name__)
+        method = getattr(super(self.__class__, self), f.__name__)
+        # call the superclass method
+        r = method(*args, **kwargs)
+        return r
+    return wrapped
+
+
 class RecordCallMixin:
     method_calls: Dict[str, int]
 
@@ -80,50 +92,50 @@ class RecordCallMixin:
         return self.method_calls[method_name]
 
     def reset_counts(self) -> None:
-        for record in self.method_calls.keys():
+        for record in self.method_calls:
             self.method_calls[record] = 0
 
 
-class SyncKernelManagerSubclass(RecordCallMixin, KernelManager):
+class SyncKMSubclass(RecordCallMixin, KernelManager):
 
+    @subclass_recorder
     def start_kernel(self, **kw):
-        self.record('start_kernel')
-        return super().start_kernel(**kw)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def shutdown_kernel(self, now=False, restart=False):
-        self.record('shutdown_kernel')
-        return super().shutdown_kernel(now=now, restart=restart)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def restart_kernel(self, now=False, **kw):
-        self.record('restart_kernel')
-        return super().restart_kernel(now=now, **kw)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def interrupt_kernel(self):
-        self.record('interrupt_kernel')
-        return super().interrupt_kernel()
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def request_shutdown(self, restart=False):
-        self.record('request_shutdown')
-        return super().request_shutdown(restart=restart)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def finish_shutdown(self, waittime=None, pollinterval=0.1):
-        self.record('finish_shutdown')
-        return super().finish_shutdown(waittime=waittime, pollinterval=pollinterval)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def _launch_kernel(self, kernel_cmd, **kw):
-        self.record('_launch_kernel')
-        return super()._launch_kernel(kernel_cmd, **kw)
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def _kill_kernel(self):
-        self.record('_kill_kernel')
-        return super()._kill_kernel()
+        """ Record call and defer to superclass """
 
+    @subclass_recorder
     def cleanup_resources(self, restart=False):
-        self.record('cleanup_resources')
-        super().cleanup_resources(restart=restart)
+        """ Record call and defer to superclass """
 
 
-class AsyncKernelManagerSubclass(RecordCallMixin, AsyncKernelManager):
+class AsyncKMSubclass(RecordCallMixin, AsyncKernelManager):
     """Used to test subclass hierarchies to ensure methods are called when expected.
 
        This class is also used to test deprecation "routes" that are determined by superclass'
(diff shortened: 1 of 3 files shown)


jupyter_client/tests/utils.py
```diff
@@ -181,3 +186,101 @@ class AsyncKernelManagerWithCleanup(AsyncKernelManager):
...
+        self.record('get_kernel')
+        return super().get_kernel(kernel_id)
+
+    def remove_kernel(self, kernel_id):
+        self.record('remove_kernel')
+        return super().remove_kernel(kernel_id)
+
+    async def start_kernel(self, kernel_name=None, **kwargs):
+        self.record('start_kernel')
+        return await super().start_kernel(kernel_name=kernel_name, **kwargs)
```
davidbrochart, line 258: It looks like we could have a common base class for `SyncMKMSubclass` and `AsyncMKMSubclass`, rather than duplicating all the methods. All the async methods can be regular functions, they will just return a result (sync) or an awaitable (async).
You could also use a `mkm_method` decorator as in:
https://github.com/jupyter/jupyter_client/blob/26a16c0c91e245f7403aa27a812fee5e905d2964/jupyter_client/multikernelmanager.py#L27-L40

kevin-bates: Thanks @davidbrochart.  I'm having trouble visualizing the common base class approach that applies to both sync/async subclasses, could you please give an example?

Regarding the decorator, I'm not very familiar with this kind of thing, would it resemble something like this:
```python

def mkm_recorder(f):
    def wrapped(self, *args, **kwargs): 
         # record this call
         self.record(f.__name__) 
         method = getattr(super(), f.__name__) 
         # call the superclass method 
         r = method(*args, **kwargs) 
         return r 
     return wrapped 

class AsyncMKMSubclass(RecordCallMixin, AsyncMultiKernelManager):
    @mkm_recorder
    async def start_kernel(self, kernel_name=None, **kwargs):
    """ Record call and defer to superclass """

class SyncMKMSubclass(RecordCallMixin, MultiKernelManager):
    @mkm_recorder
    def start_kernel(self, kernel_name=None, **kwargs):
    """ Record call and defer to superclass """
```

Thanks.

davidbrochart: I think something like that will do:
```python
class MKMSubclass(RecordCallMixin):

    @mkm_recorder
    def get_kernel(self, kernel_id):
        """ Record call and defer to superclass """

    # add all the other common methods here

class SyncMKMSubclass(MKMSubclass, MultiKernelManager):

    def _kernel_manager_class_default(self):
        return 'jupyter_client.tests.utils.SyncKernelManagerSubclass'

class AsyncMKMSubclass(MKMSubclass, AsyncMultiKernelManager):

    def _kernel_manager_class_default(self):
        return 'jupyter_client.tests.utils.AsyncKernelManagerSubclass'
```

Regarding the decorator, this looks fine.

kevin-bates:
Thanks @davidbrochart - I've pushed an update that implements a decorator.  I'd be interested in further refactoring so long as it doesn't affect the goal of ensuring a subclass's methods are called.  Hmm, now that I think about it, would this be something accomplished with something like a subclass _mixin_ class?
```python
class SyncMKMSubclass(RecordCallMixin, MKMSubclassMixin, MultiKernelManager):
    def _kernel_manager_class_default(self):
        return 'jupyter_client.tests.utils.SyncKMSubclass'


class AsyncMKMSubclass(RecordCallMixin, MKMSubclassMixin, AsyncMultiKernelManager):
    def _kernel_manager_class_default(self):
        return 'jupyter_client.tests.utils.AsyncKMSubclass'
```

I'll experiment a bit on that note.  Thanks for the ideas!

davidbrochart:
You can have a look at #628.

kevin-bates:
Thank @davidbrochart.  Ah - the trick is to use the private attribute `_km_class`. I had all of those changes in place with the exception of the attribute trick.  I'll merge #628 and update this PR.  Your changes will have some naming changes applied when merged with this PR.

davidbrochart:
> the trick is to use the private attribute `_km_class`.

Yes it was easier to have the super class explicitly named instead of trying to fit into an implicit mechanism, so I took a shortcut here :smile:

kevin-bates:
Looks like the CI failures are the result of ipykernel 5.5.1.  They literally started occurring as soon as that release was out since my private fork was successful with these changes but happened to pull ipykernel 5.5.0.

`jupyter_client/tests/test_client.py::TestKernelClient::test_execute_interactive` is now failing on all 3.6 platforms with the following traceback:
```
jupyter_client/blocking/client.py:145: TimeoutError
----------------------------- Captured stderr call -----------------------------
[IPKernelApp] ERROR | Exception in message handler:
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 261, in dispatch_shell
    yield gen.maybe_future(handler(stream, idents, msg))
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/tornado/gen.py", line 769, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 538, in execute_request
    user_expressions, allow_stdin,
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/tornado/gen.py", line 234, in wrapper
    yielded = ctx_run(next, result)
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/tornado/gen.py", line 162, in _fake_ctx_run
    return f(*args, **kw)
  File "/Users/runner/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ipykernel/ipkernel.py", line 290, in do_execute
    and should_run_async(code, transformed_cell=transformed_cell, preprocessing_exc_tuple=preprocessing_exc_tuple)
TypeError: should_run_async() got an unexpected keyword argument 'transformed_cell'
```

Since these failures are unrelated to this PR, I believe we can proceed with this PR's review.

blink1073:
I deleted ipykernel 5.5.1, cf https://github.com/ipython/ipykernel/issues/607

davidbrochart:
I don't want to be nitpicking, but `TestKernelManager` and `TestAsyncKernelManager` are very similar, and that's a lot of duplicated code. I think it is possible to have only one source of (async) tests, and have sync/async tests use this common base. For instance, you can `await` a non-async function by wrapping it with `ensure_async`:
https://github.com/jupyter/jupyter_client/blob/92cb460e1f95093b50631f2ceced6c64e77dc60b/jupyter_client/util.py#L22-L25

kevin-bates:
@blink1073 - thank you.  I've restarted the jobs.

@davidbrochart - no worries on the 'nitpicking' comment - I agree (with the duplication and don't believe you're being nitpicky :smile: ).  However, I'd prefer that refactor be handled outside the scope of this pull request since it isn't necessarily related and has existed for some time (besides the fact that I don't have the bandwidth to deal with that at the moment).

Perhaps you could open an issue and someone can contribute those changes?

davidbrochart:
No worries Kevin. You're right, we can refactor in a separate PR. Thanks!

PR #628, opened 2021-03-18
Make KernelManager subclass tests DRY

davidbrochart:


kevin-bates:
Thanks David, the use of the attribute `_km_class` was the kicker.  I'm going to merge these, then update #627 - with some renaming that I had already in place.  I think I'm also going to rename the attribute to `_km_superclass` at that time as well.

kevin-bates (approved):
Thanks for you help David - these changes look good (some renames will be applied in the updates on #627).

Issue #634, opened 2021-04-07
Release 6.1.13 breaks compatibility with Python 3.5


Lothiraldan:
Hi, the latest release of jupyter_ciient breaks compatibility with Python 3.5. Here is the traceback we are seeing on our CI:

```
/usr/local/lib/python3.5/site-packages/ipykernel/__init__.py:2: in <module>
    from .connect import *
/usr/local/lib/python3.5/site-packages/ipykernel/connect.py:16: in <module>
    import jupyter_client
/usr/local/lib/python3.5/site-packages/jupyter_client/__init__.py:6: in <module>
    from .asynchronous import AsyncKernelClient  # noqa
/usr/local/lib/python3.5/site-packages/jupyter_client/asynchronous/__init__.py:1: in <module>
    from .client import AsyncKernelClient  # noqa
/usr/local/lib/python3.5/site-packages/jupyter_client/asynchronous/client.py:6: in <module>
    from jupyter_client.channels import HBChannel
E     File "/usr/local/lib/python3.5/site-packages/jupyter_client/channels.py", line 47
E       time_to_dead: float = 1.0
```

It's not a big issue on our side, we can work-around it, but the packaging metadata needs to be updated if the breakage was intentional:
- https://github.com/jupyter/jupyter_client/blob/master/setup.py#L79
- https://github.com/jupyter/jupyter_client/blob/master/setup.py#L66

I would also recommend yanking the release 6.1.13 and doing a new release 6.1.14 with the correct packaging metadata to minimize the risk of people getting an invalid version on Python 3.5.

The conda-forge recipe likely needs to be updated as well https://github.com/conda-forge/jupyter_client-feedstock/blob/master/recipe/meta.yaml

carlosperate:
Yes, we are seeing the same issue with an app using qtconsole, as the Python 3.5 CI tests are failing with the same issue : https://github.com/mu-editor/mu/issues/1444

Personally I'd prefer if 3.5 was still supported, as is the default Python3 in Debian Stretch (eol is in July), but as this Python version is no longer supported it is still understandable if you'd drop it. Pip should still be able to automatically pick 6.1.12 with Python 3.5 once 6.1.13 is fixed or yanked.

davidbrochart:
Thanks for reporting, the issue seems to be related to Python 3.5 not supporting type annotations, so we will drop it and support Python >= 3.6 from now on.
[...]

PR #636, opened 2021-04-07
Pin python>=3.6.1

davidbrochart:
See #634

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,10 +63,10 @@ setup_args = dict(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
     ],
     install_requires=[
         'traitlets',
@@ -76,7 +76,7 @@ setup_args = dict(
         'tornado>=4.1',
         'nest-asyncio>=1.5',
     ],
-    python_requires='>=3.5',
+    python_requires='>=3.6.1',
     extras_require={
         'test': [
             'async_generator',


setup.py
```diff
@@ -63,10 +63,10 @@ def run(self):
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
```
mriedem, line 69: Unrelated to this py3.5 change but just confirming that py3.9 is being tested:

https://github.com/jupyter/jupyter_client/blob/018d8da0b947865d8217b459197bcc4f13e04666/.github/workflows/main.yml#L41

👍

setup.py
```diff
@@ -76,7 +76,7 @@ def run(self):
         'tornado>=4.1',
         'nest-asyncio>=1.5',
     ],
-    python_requires='>=3.5',
+    python_requires='>=3.6.1',
```
mriedem, line 79: So anyone using 3.5 with 6.1.13 will still have issues. What's the plan there? Does 6.1.13 get revoked/blacklisted somehow and then _this_ change goes into a new 6.2 release?

mriedem (commented):
Looks OK but a question inline.

setup.py
davidbrochart: Yes, v6.1.13 has been marked as broken on conda-forge, and I'm waiting for someone with management rights to yank it on PyPI.
When we merge this PR, I will release v6.2.

mriedem (approved):
LGTM

setup.py
MSeal: @davidbrochart I can yank 6.1.13. Do you want me to wait for the 6.2 release before yanking? Usually you want a roll-forward option before removing a dep, especially when it's not a security issue.

MSeal (approved):

setup.py
MSeal: @davidbrochart 6.1.13 is yanked. You should have permission on PyPI for this as well (let me know if you don't see the option)

davidbrochart: Thanks Matthew. Unfortunately I still don't see admin permission on PyPI.

MSeal: Aww I think I see the issue. Should be fixed now.

davidbrochart: Yes it is, thanks!

Issue #640, opened 2021-04-12
Next release


davidbrochart:
I had to pull `6.2.0` again, because:
1. a channel's `get_msg()` previously had a `block=True` kwarg, that was removed in `6.2.0`.
2. `get_msg()` is now always async, regardless of the client type (blocking or async). Previously, it was async for an async client and sync for a blocking client.

For the next release, I will fix 1. and add `block=True` back. For 2., we offer higher-level methods, e.g. `kc.get_iopub_msg()` that should be prefered over `kc.iopub_channel.get_msg()`, since it does the right thing (it is async for an async client and sync for a blocking client). I made the relevant changes in https://github.com/ipython/ipykernel/pull/634 for instance, but I don't know if `get_msg()` is used in other projects.

If we consider that `get_msg()` now being always async is an API breaking change, then we should make a major release (`7.0.0a0`). Otherwise, a `6.2.1a0` would be enough. How do people feel about this?

MSeal:
Thanks for keeping ontop of that. Given the scope of the change and impact it's having with the two yanked releases, along with potential breaking change I think 7.0.0a0 is the way to go.

kevin-bates:
Yes, thanks for working through this stuff David.   I also agree with Matt that a 7.0 alpha.  There are some deprecations and other changes we can also take advantage of with this.

davidbrochart:
Thanks, I just released 7.0.0a0.

PR #641, opened 2021-04-12
Add block=True back to get_msg()

davidbrochart:
See https://github.com/jupyter/jupyter_client/issues/640#issue-856223866

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -254,12 +254,17 @@ class ZMQSocketChannel(object):
         ident, smsg = self.session.feed_identities(msg)
         return self.session.deserialize(smsg)
 
-    async def get_msg(self, timeout: t.Optional[float] = None) -> t.Dict[str, t.Any]:
+    async def get_msg(
+        self, block: bool = True, timeout: t.Optional[float] = None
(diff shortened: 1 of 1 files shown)


kevin-bates:
I was going to comment on whether or not we should deprecate the `block` parameter at this time (and remove it in 7.0) and/or document that `timeout` is only used when `block == True` (although the comment was the same previously).   Kind of an odd API and agree that the `timeout` value serves that purpose just fine.  We should probably default its value to `0` once `block` is removed (IMO).

davidbrochart:
Ah, sorry for merging so quickly. Actually, now that we decided to go for a major release and it's still a pre-release, we could remove it.
But defaulting `timeout` to `0` means we just poll for a message, which is the opposite of what we have now by default (we wait until we have a message)?

kevin-bates:
> But defaulting timeout to 0 means we just poll for a message, which is the opposite of what we have now by default (we wait until we have a message)?

I see.  I haven't looked at the socket poll() method, but now see what you're saying where `None` essentially means wait until we have a message.  Yeah, that should probably remain the default then.

davidbrochart:
I think we can get rid of the `block` parameter, since we can do anything with `timeout` only. I opened #671.

PR #646, opened 2021-04-27
Update test kernel with native coroutine, remove async_generator dependency

kevin-bates:
While updating the Kernel Provisioning PR with all of the recent changes, I ran into a couple of general issues that appear related to some recent updates to ipykernel (`6.0a5`).  Since we're using the `--pre` option to install test dependencies, I thought I'd add these changes to the primary branch since, in all likelihood, the next PR will encounter them.

1. This updates the test kernel subclass `SignalTestKernel` to use a native coroutine that is used in `TestKernelManagerShutDownGracefully`.
2. The return of the tuple from the previously required `async_generator`-decorated method has been removed (as has `yield_`) since we no longer support Python 3.5.
3. I had to conditionally skip the `TestKernelManagerShutDownGracefully` tests that use `shutdown_request` on Python 3.6 since the `ipykernel` updates are not available in that Python release.  These changes are essentially mutually exclusive to those in item 1, so this PR errs on the side of Python 3.7+.

### DIFF
--- a/jupyter_client/tests/signalkernel.py
+++ b/jupyter_client/tests/signalkernel.py
@@ -10,7 +10,6 @@ from subprocess import Popen
 from ipykernel.displayhook import ZMQDisplayHook
 from ipykernel.kernelapp import IPKernelApp
 from ipykernel.kernelbase import Kernel
-from tornado.web import gen
 
 
 class SignalTestKernel(Kernel):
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thanks!

davidbrochart (approved):

PR #648, opened 2021-05-12
Add status field to other replies in documentation

yitzchak:
The documentation states that all replies should have a status field and can pass errors also. The python kernel appears to do this and jupyter_kernel_test includes these in its message schema, although to does not currently require their presence. 

I did not add a status to the debug_reply since it follows the DAP.


blink1073 (approved):
Thank you!

PR #649, opened 2021-05-26
Added missing documentation for the inspectVariables request and resp…

JohanMabille:
…onse

Notice that this is not a new message to the protocol, but missing documentation for a message the has been part of the debugging protocol in Jupyter since the beginning.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1164,7 +1164,7 @@ In order to support page reloading, or a client connecting at a later stage, Jup
 
   The `source_breakpoint` schema is specified by the Debug Adapter Protocol.
 
-The `inspectVariables` is meant to retrieve the values of all the variables that have been defined in the notebook. It is a DAP `Request` with no extra argument.
+The `inspectVariables` is meant to retrieve the values of all the variables that have been defined in the kernel. It is a DAP `Request` with no extra argument.
 
   Content of the `inspectVariables` request::
 
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -1104,7 +1104,11 @@ Debug requests and replies are sent over the `control` channel to prevent queuin
 Additions to the DAP
 ~~~~~~~~~~~~~~~~~~~~
 
-The Jupyter debugger protocol makes two additions to the DAP, the `dumpCell` request and response, and the `debugInfo` request and response messages.
+The Jupyter debugger protocol makes several additions to the DAP:
+
+- the `dumpCell` request and response messages
+- the `debugInfo` request and response messages
+- the `inspectVariables` request and response messages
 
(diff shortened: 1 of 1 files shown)


docs/messaging.rst
```diff
@@ -1137,7 +1141,7 @@ In order to support page reloading, or a client connecting at a later stage, Jup
           'command' : 'debugInfo'
       }
 
-  Content of `debugInfo` response::
+  Content of the `debugInfo` response::
```
davidbrochart, line 1144: I think we should get rid of the two-space indentation here, and in other places too:
```suggestion
Content of the `debugInfo` response::
```

docs/messaging.rst
```diff
@@ -1160,6 +1164,32 @@ In order to support page reloading, or a client connecting at a later stage, Jup
 
   The `source_breakpoint` schema is specified by the Debug Adapter Protocol.
 
+The `inspectVariables` is meant to retrieve the values of all the variables that have been defined in the notebook. It is a DAP `Request` with no extra argument.
+
+  Content of the `inspectVariables` request::
```
davidbrochart, line 1169: Same here.

docs/messaging.rst
```diff
@@ -1160,6 +1164,32 @@ In order to support page reloading, or a client connecting at a later stage, Jup
...
+
+      {
+          'type' : 'request',
+          'command' : 'inspectVariables'
+      }
+
+  Content of the `inspectVariables` response::
```
davidbrochart, line 1176: Same here.

docs/messaging.rst
```diff
@@ -1160,6 +1164,32 @@ In order to support page reloading, or a client connecting at a later stage, Jup
 
   The `source_breakpoint` schema is specified by the Debug Adapter Protocol.
 
+The `inspectVariables` is meant to retrieve the values of all the variables that have been defined in the notebook. It is a DAP `Request` with no extra argument.
```
jtpio, line 1167: Maybe "kernel" as an alternative to "notebook" is more accurate?

```suggestion
The `inspectVariables` is meant to retrieve the values of all the variables that have been defined in the kernel. It is a DAP `Request` with no extra argument.
```

docs/messaging.rst
JohanMabille: If you look at the generated [doc](https://jupyter-client.readthedocs.io/en/stable/messaging.html#additions-to-the-dap), this allwos to easily identify the different messages added to the spec. I think we should either keep this indentation, or add subsections.

davidbrochart: > add subsections

Yes maybe as a bullet list?

JohanMabille: Let's do this reformating part in a dedicated PR since the indentation was here before. We really need to move on and merge this one since we have more fields to add in some messages.

davidbrochart (approved):

JohanMabille:
Thanks @davidbrochart !

PR #651, opened 2021-05-28
PR: Don't block on exit

impact27:
If the code tries to exit while calling `self.poller.poll`, the app will freeze until `until_dead` seconds have elapsed. This avoids that outcome.

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -105,7 +105,7 @@ class HBChannel(Thread):
 
         self.poller.register(self.socket, zmq.POLLIN)
 
-    def _poll(self, start_time: float) -> t.List[t.Any]:
+    def _poll(self) -> t.List[t.Any]:
         """poll for heartbeat replies until we reach self.time_to_dead.
 
         Ignores interrupts, and returns the result of poll(), which
@@ -113,9 +113,7 @@ class HBChannel(Thread):
         or the event tuple if there is a message to receive.
         """
         # Wait until timeout
-        until_dead = self.time_to_dead - (time.time() - start_time)
-        if until_dead > 0:
-            self._exit.wait(until_dead)
+        self._exit.wait(self.time_to_dead)
         if not self._exit.is_set():
             # 0 means return immediately.
             # http://api.zeromq.org/2-1:zmq-poll
@@ -145,7 +143,7 @@ class HBChannel(Thread):
             # either a recv or connect, which cannot be followed by EFSM
             await self.socket.send(b"ping")
             request_time = time.time()
-            ready = self._poll(request_time)
+            ready = self._poll()
             if ready:
                 self._beating = True
                 # the poll above guarantees we have something to recv
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -137,7 +137,7 @@ class HBChannel(Thread):
                 else:
                     raise
             else:
-                if events or self._exit.isSet():
+                if events or self._exit.is_set():
                     break
                 elif self.time_to_dead - (time.time() - start_time) < 0:
                     break
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -114,19 +114,15 @@ class HBChannel(Thread):
         will be an empty list if no messages arrived before the timeout,
         or the event tuple if there is a message to receive.
         """
-        # Check if an event is waiting
-        events = self.poller.poll(0)
-        if events:
-            return events
-
         # Wait until timeout
         until_dead = self.time_to_dead - (time.time() - start_time)
-        # ensure poll at least once
-        until_dead = max(until_dead, 1e-3)
-        self._exit.wait(until_dead)
+        if until_dead > 0:
+            self._exit.wait(until_dead)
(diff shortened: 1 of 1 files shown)


jupyter_client/channels.py
```diff
@@ -121,7 +121,7 @@ def _poll(self, start_time: float) -> t.List[t.Any]:
         events = []
         while True:
             try:
-                events = self.poller.poll(int(1000 * until_dead))
+                events = self.poller.poll(1)
```
davidbrochart, line 124: Does it mean we are going to iterate every milllisecond?

impact27: Well a bit longer than that because of the overhead, the time could be adjusted. How does 100ms sounds?

impact27: On second thought, I don't think the loop is necessary at all. I removed it.

jupyter_client/channels.py
```diff
@@ -114,30 +114,18 @@ def _poll(self, start_time: float) -> t.List[t.Any]:
         will be an empty list if no messages arrived before the timeout,
         or the event tuple if there is a message to receive.
         """
+        # Check if an event is waiting
+        events = self.poller.poll(0)
```
davidbrochart, line 118: Does it ensure we poll at least once?

impact27: If I understand http://api.zeromq.org/2-1:zmq-poll correctly:

> If none of the requested events have occurred on any zmq_pollitem_t item, zmq_poll() shall wait timeout microseconds for an event to occur on any of the requested items. If the value of timeout is 0, zmq_poll() shall return immediately. If the value of timeout is -1, zmq_poll() shall block indefinitely until a requested event has occurred on at least one zmq_pollitem_t. The resolution of timeout is 1 millisecond.

A timeout of 0 means that if a message is already there it will be polled.

davidbrochart: > A timeout of 0 means that if a message is already there it will be polled.

Thanks, although [pyzmq doesn't mention that](https://pyzmq.readthedocs.io/en/latest/api/zmq.html?highlight=poll#zmq.Poller.poll).

jupyter_client/channels.py
```diff
@@ -114,30 +114,18 @@ def _poll(self, start_time: float) -> t.List[t.Any]:
...
+        if events:
+            return events
 
+        # Wait until timeout
         until_dead = self.time_to_dead - (time.time() - start_time)
         # ensure poll at least once
         until_dead = max(until_dead, 1e-3)
```
davidbrochart, line 125: What is the point of waiting at least one millisecond?

davidbrochart:
I think that with your changes, this is not needed anymore:
https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/channels.py#L171-L174

jupyter_client/channels.py
```diff
@@ -114,30 +114,18 @@ def _poll(self, start_time: float) -> t.List[t.Any]:
         will be an empty list if no messages arrived before the timeout,
         or the event tuple if there is a message to receive.
         """
+        # Check if an event is waiting
+        events = self.poller.poll(0)
+        if events:
+            return events
```
davidbrochart, line 120: I don't think we should check if we have an event at the beginning, we should only wait until_dead and then check.

impact27:
@davidbrochart I applied your comments.

jupyter_client/channels.py
```diff
@@ -163,17 +143,13 @@ async def _async_run(self) -> None:
             # either a recv or connect, which cannot be followed by EFSM
             await self.socket.send(b"ping")
             request_time = time.time()
-            ready = self._poll(request_time)
+            ready = self._poll()
             if ready:
                 self._beating = True
```
davidbrochart, line 148:
```suggestion
            # Wait until timeout
            self._exit.wait(self.time_to_dead)
            # poll(0) means return immediately (see http://api.zeromq.org/2-1:zmq-poll)
            self._beating = bool(self.poller.poll(0))
            if self._beating:
```

jupyter_client/channels.py
```diff
@@ -163,17 +143,13 @@ async def _async_run(self) -> None:
...
                 await self.socket.recv()
-                # sleep the remainder of the cycle
-                remainder = self.time_to_dead - (time.time() - request_time)
-                if remainder > 0:
-                    self._exit.wait(remainder)
                 continue
-            else:
+            elif not self._exit.is_set():
                 # nothing was received within the time limit, signal heart failure
                 self._beating = False
```
davidbrochart, line 154:
```suggestion
```

jupyter_client/channels.py
```diff
@@ -107,38 +105,20 @@ def _create_socket(self) -> None:
 
         self.poller.register(self.socket, zmq.POLLIN)
 
-    def _poll(self, start_time: float) -> t.List[t.Any]:
+    def _poll(self) -> t.List[t.Any]:
```
davidbrochart, line 108: I don't think we need the `_poll()` method anymore.

davidbrochart (approved):
Awesome, thanks a lot for your efforts and patience!

PR #652, opened 2021-05-31
Change to edit_magic payload

yitzchak:
The Qt console and the ipython kernel both use a payload source of "edit_magic", not "edit".


blink1073 (approved):
Thank you!

PR #664, opened 2021-07-01
Use a default serializer that is not only for date types

martinRenou:
This would allow making more general changes here. I am also adding serialization of bytes type into base64. 

For more context, this is going in the direction of https://github.com/ipython/ipykernel/issues/706. This allows getting rid of the `json_clean` function, moving the bytes treatment from ipykernel to jupyter_client when actually doing the serializion of the object.

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 import re
 import warnings
+from binascii import b2a_base64
 from datetime import datetime
 from typing import Optional
 from typing import Union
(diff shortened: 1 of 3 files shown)


jupyter_client/jsonutil.py
```diff
@@ -91,9 +92,19 @@ def squash_dates(obj):
 
 
 def date_default(obj):
-    """default function for packing datetime objects in JSON."""
+    """DEPRECATED: Use jupyter_client.jsonutil.json_default"""
+    warnings.warn(
+        "date_default is deprecated. Use jupyter_client.jsonutil.json_default.", stacklevel=2
```
Carreau, line 97: Would you care to add a `since version XXX`, I don't know if you want to do 7.0 or backport. But I think that will help users to transition.

Carreau:
+1

jupyter_client/jsonutil.py
martinRenou: I wouldn't care :)

Should we do backports to `6.x.x`? I would actually not mind doing it, and it's not that big of a change, and not backward-incompatible.

martinRenou:
pip timed out in one of the builds. Could someone restart it?

davidbrochart:
Is it ready to be merged?

martinRenou:
I guess I still need to process https://github.com/jupyter/jupyter_client/pull/664#discussion_r662345972

But the question is what should be the comment? Should it be `since version 7.0`?

davidbrochart:
> But the question is what should be the comment? Should it be `since version 7.0`?

I think `since version 7.0` is fine.

martinRenou:
Done :)

davidbrochart:
Not quite yet :-)

davidbrochart:
Thanks!

martinRenou:
Thank you!

PR #666, opened 2021-07-09
Contain usage of nest-asyncio

SylvainCorlay:


davidbrochart (approved):

kevin-bates:
Hi @SylvainCorlay.  The changes look good to me so far.  Prior to merge, could you please add a description that includes the reason for these changes and what impact they might have on existing applications that subclass or use `KernelManager` directly?  I suspect they too would need to move to `BlockingKernelManager`.  Given this will likely be slated for 7.x (correct?), I don't see that being an issue.

IrfanJaya56 (changes requested):
>

blink1073:
nest-asyncio usage was removed in #835

PR #670, opened 2021-08-02
Fix up some async method aliases in KernelManager

kevin-bates:
In preparation for the 7.0 release, I was perusing the changes to `KernelManager` and noticed some "internal" methods (i.e., methods that won't typically be overridden and are called from within "public" methods) had been changed to `async` but didn't have a backward-compatible "non-async" alias.  As a result, applications that subclass `KernelManager` _**and**_ override any of the following methods will not have their overrides called.  

- `pre_start_kernel()`
- `post_start_kernel()`
- `request_shutdown()`

This pull request addresses this issue by introducing the appropriate aliasing thereby restoring the hierarchical call sequence.

Note that these methods were made `async` for `AsyncKernelManager`, but that will be called out in the 7.0 release notes - along with the other changes.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -274,7 +274,7 @@ class KernelManager(ConnectionFileMixin):
         self._control_socket.close()
         self._control_socket = None
 
-    async def pre_start_kernel(self, **kw) -> t.Tuple[t.List[str], t.Dict[str, t.Any]]:
+    async def _async_pre_start_kernel(self, **kw) -> t.Tuple[t.List[str], t.Dict[str, t.Any]]:
         """Prepares a kernel for startup in a separate process.
 
         If random ports (port=0) are being used, this method must be called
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!  I kicked the downstream build assuming the failure was spurious

kevin-bates:
Thanks @blink1073.  Yeah, all the tests had passed within my fork prior to creating the PR, and was about to do the same.

kevin-bates:
@davidbrochart - With these changes, I think the qtconsole update in https://github.com/jupyter/qtconsole/pull/494 should be rolled back as these are more correct and preserve "synchronicity" in `KernelManager`'s API that otherwise had been hybrid only with respect to these 3 updated methods.  I'm sorry I didn't catch this prior to these updates.

davidbrochart:
Right, I just run the qtconsole tests with jupyter_client master and with this PR they fail. Let's roll back https://github.com/jupyter/qtconsole/pull/494.

Issue #642, opened 2021-04-12
Tracking release 7.0 changes


kevin-bates:
Since we will be delivering a major release (7.0), I'm hoping we can use this issue to track changes that might affect existing applications.  We should, at this time, go ahead and remove deprecations that have been in place as well as other api-changes (preferably minor in nature). 

- [ ] Remove deprecated [`kernel_cmd`](https://github.com/jupyter/jupyter_client/blob/96e23a843102ccb8e81aa46d1a28d8841804a8f2/jupyter_client/manager.py#L148) trait
- [ ] Remove deprecated [`cleanup()`](https://github.com/jupyter/jupyter_client/blob/96e23a843102ccb8e81aa46d1a28d8841804a8f2/jupyter_client/manager.py#L438) method
- [ ] Remove `block` parameter from [`get_msg()`](https://github.com/jupyter/jupyter_client/blob/96e23a843102ccb8e81aa46d1a28d8841804a8f2/jupyter_client/channels.py#L258)
- [ ] Make kernel manager [`request_shutdown()`](https://github.com/jupyter/jupyter_client/blob/96e23a843102ccb8e81aa46d1a28d8841804a8f2/jupyter_client/manager.py#L381) `async`
- [ ] Rename `KernelSpecManager.whitelist` trait to `allowlist`, `allowed_kernels`, ?
- [ ] Document these changes and others in the docs

We should probably talk about branching as well.  I guess we'll keep working off the tip and create a 6.x branch against the appropriate tag once a 6.x PR is required.

kevin-bates:
Here's an attempt to document the API changes mostly brought about by the Kernel Provisioning changes (#612).  This content will need to be relocated but I wanted to get an initial form persisted somewhere.

## API Changes
All of the API changes for `KernelManager` and `AsyncKernelManager` in the 7.0 release were confined to _internal public_ methods, which we define as methods called from the _formally public_ methods but could be overridden in subclass implementations.  As a result, these changes may impact subclasses of `KernelManager` or `AsyncKernelManager` provided those implementations also implement or call these methods, but should not affect applications that call only the _formally public_ methods.

### `KernelManager`
The following internal methods had signature changes:
- `def pre_start_kernel(self, **kwargs) -> Tuple[List[str], Dict[str, Any]]:`
    - `pre_start_kernel` now returns a tuple consisting of the formatted kernel startup list and an updated set of keyword arguments.
- `def finish_shutdown(self, restart: bool = False):`
[...]

PR #671, opened 2021-08-03
Remove block parameter from get_msg()

davidbrochart:
See https://github.com/jupyter/jupyter_client/pull/641#issuecomment-891868964

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -216,17 +216,12 @@ class ZMQSocketChannel(object):
         ident, smsg = self.session.feed_identities(msg)
         return self.session.deserialize(smsg)
 
-    async def get_msg(
-        self, block: bool = True, timeout: t.Optional[float] = None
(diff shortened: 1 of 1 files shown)


kevin-bates (approved):
Thanks David - this LGTM.  Since this, if I'm understanding things correctly, re-introduces the reasons for the 6.x releases getting yanked from PyPi, is there anything clients need to do differently with 7.0?  If so, could you please jot down some notes in #671 that we'll gather into the release notes?

davidbrochart:
> is there anything clients need to do differently with 7.0?

Yes, they need to replace `block=False` with `timeout=0`.

> If so, could you please jot down some notes in #671 that we'll gather into the release notes?

I guess you meant #642. If so, sure I will do it.

kevin-bates:
> I guess you meant #642. If so, sure I will do it.

Yep - sorry about that. :smile:.  Thank you for the update.

PR #675, opened 2021-08-09
Force install jupyter_client master

davidbrochart:


davidbrochart:
As pointed out by @blink1073, we didn't test jupyter_client master but whatever version was installed by the downstream packages.
With this PR we do and we can see that:
- ipykernel still uses `get_msg(block=...)`. I opened https://github.com/ipython/ipykernel/pull/736 to get rid of the `block` parameter.
- nbclient still uses `km.cleanup`. I opened https://github.com/jupyter/nbclient/pull/152 to replace it with `km.cleanup_resources`.

blink1073:
Kicking CI

davidbrochart:
nbclient has not been released so we're not testing it with https://github.com/jupyter/nbclient/pull/152.

blink1073:
We can kick this again after upcoming releases of `ipykernel` and `nbclient` on Monday.

blink1073:
Looka like we're still failing one test in `nbclient`

blink1073:
Builds are passing in [nbclient](https://github.com/jupyter/nbclient/actions/runs/1134904828)

davidbrochart:
Looks like we also have all the downstream packages' dependencies pre-release installed, in particular `ipywidgets==8.0.0a5` makes the nbclient test fail.

davidbrochart:
I think this is the normal behavior of the `--pre` option. Maybe then we should install master versions of downstream packages (without `--pre`).

blink1073:
Thank you!

PR #681, opened 2021-08-17
Create migration.md

blink1073:
Add migration file with migration guide from 6.0 -> 7.0.   cf https://github.com/jupyter/jupyter_client/issues/642#issuecomment-891418487


kevin-bates (approved):
Great idea to have a migration guide - thanks Steve!

PR #683, opened 2021-08-19
Use formal method names when called internally

### LINKED ISSUE #682: jupyter_client v6.20 (and maybe also v7.0.0?) interrupts jupyter-enterprise-gateway on kernel change, instead of just interrupting the kernel
I ran into a strange issue this morning: I upgraded the version of jupyter_client in my jupyter-enterprise-gateway (JEG) deployment to v6.2.0 and then became unable to switch kernels. I could still start as many kernels as I liked normally, but the moment I tried to restart, stop, or change a kernel the gateway instance would crash. This was on a work machine, so unfortunately I can't copy/paste the error message, but essentially the `os.killpg` line here:

https://github.com/jupyter/jupyter_client/blob/c16c0c9b183c7aa209fe9e6e1a20bada78fbf7ae/jupyter_client/manager.py#L659-L664

was somehow sending `SIGINT` to JEG whenever a kernel change was requested, which then "gracefully" shut down due to "keyboard interrupt". Apparently the process group id of JEG itself is somehow getting passed into `os.killpg`.

For me the fix was to downgrade to jupyter_client v6.1.12, so I'm fairly certain that this is a jupyter_client issue and not a JEG problem. Since jupyter_client v6.2.0 has already been yanked from pypi for an unrelated version conflict issue, the question now is whether the problem is still present in jupyter_client v7.0.0. Though it moved around a bit, the relevant `os.killpg` command is still present in the latest HEAD:

https://github.com/jupyter/jupyter_client/blob/109e7ac137253b9245b672b27f0ca36c716b1be6/jupyter_client/provisioning/local_provisioner.py#L82-L86

so it may still cause problems for JEG.

For right now, this can't be directly tested as the latest version of JEG is pinned `jupyter_client<7`, since JEG is currently incompatible with the kernel provisioner stuff. So this issue is mostly just something to keep an eye out for in the near future.

pinging @kevin-bates

kevin-bates:
Fixup a couple more instances where we were calling the internal method name rather than the "aliased" (for lack of a better term) name.  This prevented some lower-level method overrides in subclasses from being called.  The test which was introduced for this purpose has been extended accordingly.

See https://github.com/jupyter/jupyter_client/issues/682#issuecomment-902260179 for more information.

Resolves #682

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -373,7 +373,7 @@ class KernelManager(ConnectionFileMixin):
         except asyncio.TimeoutError:
             self.log.debug("Kernel is taking too long to finish, terminating")
             self._shutdown_status = _ShutdownStatus.SigtermRequest
-            await self._async_send_kernel_sigterm()
+            await ensure_async(self._send_kernel_sigterm())
 
         try:
             await asyncio.wait_for(
@@ -531,7 +531,7 @@ class KernelManager(ConnectionFileMixin):
             assert self.kernel_spec is not None
             interrupt_mode = self.kernel_spec.interrupt_mode
             if interrupt_mode == "signal":
-                await self._async_signal_kernel(signal.SIGINT)
+                await ensure_async(self.signal_kernel(signal.SIGINT))
 
             elif interrupt_mode == "message":
                 msg = self.session.msg("interrupt_request", content={})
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thanks!

PR #686, opened 2021-08-30
Don't set event loop policy on Windows at import time

minrk:
Imports shouldn't have side effects like this.

I know it's annoying, but this kind of process initialization needs to be in *application* startup, not library import, otherwise you can end up overriding other code unintentionally.

Plus, with tornado 6.1 and pyzmq 22, the default proactor loop generally works anyway, even if selector is often a better fit for zmq-based applications.

Related to https://github.com/dask/distributed/issues/5272 I believe

### DIFF
--- a/jupyter_client/utils.py
+++ b/jupyter_client/utils.py
@@ -1,15 +1,11 @@
 """
 utils:
-- provides utility wrapeprs to run asynchronous functions in a blocking environment.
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):

blink1073:
I'll cut a patch release today, thanks @minrk!

blink1073:
https://github.com/jupyter/jupyter_client/releases/tag/v7.0.2

PR #688, opened 2021-08-31
use load_connection_info(info) when constructing a blocking client

minrk:
passing connection_info to constructor passes Session args (key, scheme) to the Client, which trigger warnings for unrecognized arguments

We could also add logic in `ConnectionFileMixin.__init__` to explicitly support all the fields of a connection file in the constructor, which might be nice. But we have the load_info method already.


blink1073 (approved):
Thank you!

Issue #689, opened 2021-09-14
kernel can only autorestart once


mofanke:
kernel can only autorestart once
Name: jupyter-client
Version: 7.0.2

kill kernel process first time, it will autorestart . kill it agian , nothing will happen

mofanke:
seems self.shutting_down in KernelManager never set to False ,after restart

kevin-bates:
Hmm - I'm not experiencing this.  If I kill the kernel pid, the auto-restart takes place each time (across all 3 attempts).  That said, I see what you mean regarding `self.shutting_down` and it seems it deserves a reset in the post start method.  Did you want to take a crack at this?  Since I'm not reproducing this, it would be good for you to at least check out the solution (from wherever that comes).

mofanke:
this issue occurred  when  jupyter-client==7.0.2,  and i add  "self.shutting_down = False"  in  funciton "async def _async_post_start_kernel(self, **kw) -> None:"    it will be ok

kevin-bates:
Hi @mofanke - thanks for the update. 

Yes, that sounds like the correct approach.  Did you want to go about creating a pull request for this?

mofanke:
fixed

PR #695, opened 2021-09-17
fix kernel can only restart once issue

mofanke:
fix kernel can only resstart once  issue
fix issue https://github.com/jupyter/jupyter_client/issues/689

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -286,6 +286,7 @@ class KernelManager(ConnectionFileMixin):
              keyword arguments that are passed down to build the kernel_cmd
              and launching the kernel (e.g. Popen kwargs).
         """
+        self.shutting_down = False
         self.kernel_id = self.kernel_id or kw.pop('kernel_id', str(uuid.uuid4()))
         # save kwargs for use in restart
         self._launch_args = kw.copy()


kevin-bates (commented):
Hi @mofanke - thank you for the pull request.  I was thinking the reset would occur at the start of the start_kernel() method (rather than its post-start).  Did you try that and was it problematic?

mofanke:
Hi, @kevin-bates  I have tried the reset in start_kernel(), pre_start_kernel() and post_start_kernel() , all the three ways can git rid of the problem.    "self.shutting_down = True"  setted in _async_shutdown_kernel()  just before "self.stop_restarter()" so I put "self.shutting_down = False"  as the reset just before "self.start_restarter()" in _async_post_start_kernel(). 

shoud I update the PR , and set "self.shutting_down = False"  in start_kernel()?

kevin-bates:
Hi @mofanke - thanks for looking into the various solutions.  In looking at the history of `self.shuttting_down` I must have inadvertently removed its reset to `False` when introducing kernel provisioners.  It was previously located at the [start of `pre_start_kernel()`](https://github.com/jupyter/jupyter_client/commit/4d62c5b3733e8c4d54ceba743fb11e13f3c56ffc#diff-a3ba5852c735811a3643b5461c200d95819fe806dba20d1e4e692ecdba5cb62cR251), which makes the most sense to me since, at that point, it's no longer shutting down.  As such, I think its best to place it in that same relative location.  Thank you.

kevin-bates (approved):
LGTM - thank you @mofanke!

PR #697, opened 2021-09-22
Clean up the pending task

### LINKED ISSUE #696: run_sync does not clean up the pending task after run_until_complete is interrupted
When the wrapped method (like `BlockingKernelClient.get_shell_msg`) is interrupted by KeyboardInterrupt(Ctrl+C), the event loop is interrupted, and the task ( `asyncio.Task`  ) becomes a pending state.
If I call the wrapped method again, it does not return.

I think `run_sync` should cancel the pending task.
https://stackoverflow.com/questions/30765606/whats-the-correct-way-to-clean-up-after-an-interrupted-event-loop

shingo78:
Fix #696

### DIFF
--- a/jupyter_client/utils.py
+++ b/jupyter_client/utils.py
@@ -21,9 +21,9 @@ def run_sync(coro):
         future = asyncio.ensure_future(coro(*args, **kwargs))
         try:
             return loop.run_until_complete(future)
-        except KeyboardInterrupt:
+        except BaseException as e:
             future.cancel()
-            raise
+            raise e
 
     wrapped.__doc__ = coro.__doc__
     return wrapped
--- a/jupyter_client/utils.py
+++ b/jupyter_client/utils.py
@@ -18,7 +18,12 @@ def run_sync(coro):
         import nest_asyncio  # type: ignore
 
         nest_asyncio.apply(loop)
-        return loop.run_until_complete(coro(*args, **kwargs))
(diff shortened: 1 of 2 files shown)


jupyter_client/utils.py
```diff
@@ -18,7 +18,12 @@ def wrapped(*args, **kwargs):
         import nest_asyncio  # type: ignore
 
         nest_asyncio.apply(loop)
-        return loop.run_until_complete(coro(*args, **kwargs))
+        future = asyncio.ensure_future(coro(*args, **kwargs))
+        try:
+            return loop.run_until_complete(future)
+        except KeyboardInterrupt:
+            future.cancel()
+            raise
```
davidbrochart, line 26: Maybe we want to handle any kind of exception?

```suggestion
        except BaseException as e:
            future.cancel()
            raise e
```

davidbrochart (commented):
Thanks, that looks good. It would be great to add a test.

jupyter_client/utils.py
shingo78: I agree with your suggestion.

davidbrochart:
Great, do you feel like adding a test?

shingo78:
Yes, I would like to add a test.

shingo78:
Please review the test I pushed.

davidbrochart (approved):

davidbrochart:
Thanks @shingo78, that looks great.

PR #698, opened 2021-09-27
Fix json_default so that it's closer to what ipykernel had before

martinRenou:
ipykernel's `json_clean` used to support generators, iterables, and int/float like types, this adds support for them

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -1,9 +1,11 @@
 """Utilities to manipulate JSON objects."""
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+import numbers
 import re
 import warnings
 from binascii import b2a_base64
+from collections.abc import Iterable
 from datetime import datetime
 from typing import Optional
 from typing import Union
@@ -106,7 +108,17 @@ def json_default(obj):
     if isinstance(obj, datetime):
         obj = _ensure_tzinfo(obj)
         return obj.isoformat().replace('+00:00', 'Z')
-    elif isinstance(obj, bytes):
+
+    if isinstance(obj, bytes):
         return b2a_base64(obj).decode('ascii')
-    else:
-        raise TypeError("%r is not JSON serializable" % obj)
+
+    if isinstance(obj, Iterable):
+        return list(obj)
+
+    if isinstance(obj, numbers.Integral):
+        return int(obj)
+
+    if isinstance(obj, numbers.Real):
+        return float(obj)
+
+    raise TypeError("%r is not JSON serializable" % obj)
--- a/jupyter_client/tests/test_jsonutil.py
+++ b/jupyter_client/tests/test_jsonutil.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 import datetime
 import json
+import numbers
 from datetime import timedelta
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


jupyter_client/jsonutil.py
```diff
@@ -108,5 +109,14 @@ def json_default(obj):
         return obj.isoformat().replace('+00:00', 'Z')
     elif isinstance(obj, bytes):
         return b2a_base64(obj).decode('ascii')
+    elif (
+        isinstance(obj, (set, types.GeneratorType)) or
+        hasattr(obj, '__iter__') and hasattr(obj, '__next__')
+    ):
+        return list(obj)
+    elif hasattr(obj, '__int__'):
+        return int(obj)
```
davidbrochart, line 118: A `float` also has a `__int__` method, so this will convert all floats to ints?

martinRenou: The test for floats is passing so I don't believe so

martinRenou: But I changed the check anyway

martinRenou: Thanks for the review!

jupyter_client/jsonutil.py
```diff
@@ -106,7 +108,21 @@ def json_default(obj):
...
-        raise TypeError("%r is not JSON serializable" % obj)
+
+    if (
+        isinstance(obj, (set, types.GeneratorType))
+        or hasattr(obj, '__iter__')
+        and hasattr(obj, '__next__')
+    ):
+        return list(obj)
+
+    if isinstance(obj, numbers.Integral):
```
davidbrochart, line 122: Why not:
```suggestion
    if isinstance(obj, int):
```

martinRenou: Then it wouldn't work with:

```python
class MyInt(object):
    def __int__(self):
        return 389

numbers.Integral.register(MyInt)
```

davidbrochart: :+1:

jupyter_client/jsonutil.py
```diff
@@ -106,7 +108,21 @@ def json_default(obj):
...
+    if isinstance(obj, bytes):
         return b2a_base64(obj).decode('ascii')
-    else:
-        raise TypeError("%r is not JSON serializable" % obj)
+
+    if (
+        isinstance(obj, (set, types.GeneratorType))
+        or hasattr(obj, '__iter__')
+        and hasattr(obj, '__next__')
+    ):
```
davidbrochart, line 119: Why not (with `from collections.abc import Iterable`):
```suggestion
    if isinstance(obj, Iterable):
```

martinRenou: Thanks!

jupyter_client/jsonutil.py
```diff
@@ -106,7 +109,17 @@ def json_default(obj):
...
         obj = _ensure_tzinfo(obj)
         return obj.isoformat().replace('+00:00', 'Z')
-    elif isinstance(obj, bytes):
+
+    if isinstance(obj, bytes):
         return b2a_base64(obj).decode('ascii')
-    else:
-        raise TypeError("%r is not JSON serializable" % obj)
+
+    if isinstance(obj, (set, types.GeneratorType, Iterable)):
```
davidbrochart, line 116: Sets and generators are iterables.
```suggestion
    if isinstance(obj, Iterable):
```

martinRenou: Thanks!

davidbrochart (approved):

davidbrochart:
Thanks!

PR #706, opened 2021-09-30
Fix NaN and Inf serialization

martinRenou:
cc. @SylvainCorlay @minrk 

This is fixing serialization of `math.nan` and `math.inf`, preventing: `ValueError: Out of range float values are not JSON compliant`

We used to clean NaNs and Inf in `json_clean` from ipykernel before serializing them, but since https://github.com/ipython/ipykernel/pull/708 we need to take care of them

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -12,7 +12,6 @@ Sessions.
 # Distributed under the terms of the Modified BSD License.
 import hashlib
 import hmac
-import json
 import logging
 import os
 import pickle
@@ -27,6 +26,7 @@ from hmac import (
     compare_digest,
 )  # We are using compare_digest to limit the surface of timing attacks
 
+import simplejson as json
 import zmq
 from traitlets import Any  # type: ignore
 from traitlets import Bool
@@ -96,7 +96,7 @@ def json_packer(obj):
         obj,
         default=json_default,
         ensure_ascii=False,
-        allow_nan=False,
+        ignore_nan=True,
     ).encode("utf8")
 
 
--- a/jupyter_client/tests/test_jsonutil.py
+++ b/jupyter_client/tests/test_jsonutil.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 import datetime
 import json
+import math
 import numbers
 from datetime import timedelta
 from unittest import mock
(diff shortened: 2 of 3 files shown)


jupyter_client/session.py
```diff
@@ -27,6 +26,7 @@
     compare_digest,
 )  # We are using compare_digest to limit the surface of timing attacks
 
+import simplejson as json
```
minrk, line 29: why the added dependency? I think we should be able to do this with stdlib json, since we always have before.

martinRenou: Before, we had `json_clean` in ipykernel that was going through all messages (making copies of lists and dicts) and replacing `math.inf` and `math.nan` manually.

Unfortunately the standard json lib only allows two options:
- `allow_nan=True`: nan is serialized to `NaN`, inf is serialized to `Infinity`, which are both valid JS but not valid JSON
- `allow_nan=False`: throw an exception for nan and inf without letting us a change to serialize them to `null` (valid JSON) in the `default` function

minrk: Arg, that's frustrating. And support for nan/inf is required somewhere? This is an artifact of earlier "never raise" goals, but I think maybe something that can't actually be serialized should raise. What breaks if these raise?

martinRenou: Both bqplot and pythreejs are failing now due to this. Arguably, those libraries should maybe provide valid JSON.

minrk: simplejson does the wrong thing for bytes:

```
In [2]: simplejson.dumps({'buf': b'bytes',}, default=repr)
Out[2]: '{"buf": "bytes"}'
```

we should have a test that verifies this, because the switch to simplejson should have caused tests to fail.

Can you add

```python
        ({"key": b"\xFF"}, {"key": "/w==\n"}),
```

to the list of pairs in `test_json_default`? And make sure the same json implementation is used?

I think the bytes behavior is more important than the unsupported floats. matplotlib figures rely on it, for one.

What I think is the ideal behavior is to trigger a DeprecationWarning when these values are seen, but still coerce to null. I don't see a good way to do either with simplejson *or* stdlib json, though.

If supporting these values is deprecated, maybe the thing to do is to fallback on the old inefficient path:

```python
try:
    dumps(obj, default=json_default)
except ValueError as e:
    # maybe double check that it's the nan/inf error
    # DeprecationWarning about inf/nan support
    obj = json_clean(obj)
    return dumps(obj, default=json_default)
```

What do you think?

martinRenou: > and push a fix to ipykernel

Actually, should we do this fallback to `json_clean` in `jupyter_client`? It's `jupyter_client` that does the dumps

minrk: > Actually, should we do this fallback to json_clean in jupyter_client? It's jupyter_client that does the dumps

Yeah, I think doing it here makes sense. I'm not sure ipykernel can add a catch in the right place.

Issue #701, opened 2021-09-28
`jupyter_client` >= 7.0.0 breaks `ipydatagrid`


ibdafna:
Hi team,

This issue surfaced during a routine PR in the `ipydatagrid` repo. There were no code changes to trigger this on the `ipydatagrid` side, and our CI only shows this as failing for Python >= 3.7, Python 3.6 seems to work fine with >=7.0.0. Wondering if there's anything we need to change on the `ipydatagrid` side, or if it's an issue with `jupyter_client`.

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_2726/322807178.py in <module>
     42     layout={"height": "180px"},
     43     header_renderer=header_renderer,
---> 44     default_renderer=default_renderer,
     45 )
     46 

~/work/ipydatagrid/ipydatagrid/.tox/py37/lib/python3.7/site-packages/ipydatagrid/datagrid.py in __init__(self, dataframe, **kwargs)
    355 
    356         self.data = dataframe
--> 357         super().__init__(**kwargs)
    358         self._cell_click_handlers = CallbackDispatcher()
    359         self._cell_change_handlers = CallbackDispatcher()

~/work/ipydatagrid/ipydatagrid/.tox/py37/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in __init__(self, **kwargs)
    413 
    414         Widget._call_widget_constructed(self)
--> 415         self.open()
    416 
    417     def __del__(self):

~/work/ipydatagrid/ipydatagrid/.tox/py37/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in open(self)
    436                 args['comm_id'] = self._model_id
    437 
--> 438             self.comm = Comm(**args)
    439 
    440     @observe('comm')

~/work/ipydatagrid/ipydatagrid/.tox/py37/lib/python3.7/site-packages/ipykernel/comm/comm.py in __init__(self, target_name, data, metadata, buffers, **kwargs)
     55             if self.primary:
     56                 # I am primary, open my peer.
---> 57                 self.open(data=data, metadata=metadata, buffers=buffers)
     58             else:
     59                 self._closed = False

~/work/ipydatagrid/ipydatagrid/.tox/py37/lib/python3.7/site-packages/ipykernel/comm/comm.py in open(self, data, metadata, buffers)
     92                               data=data, metadata=metadata, buffers=buffers,
     93                               target_name=self.target_name,
[...]

PR #708, opened 2021-09-30
Fallback to the old ipykernel "json_clean" if we are not able to serialize a JSON message

martinRenou:
As discussed in https://github.com/jupyter/jupyter_client/pull/706#discussion_r719308921

Fix https://github.com/jupyter-widgets/pythreejs/issues/366 and fix https://github.com/jupyter/jupyter_client/issues/701

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -102,18 +102,21 @@ def json_packer(obj):
         ).encode("utf8")
     except ValueError as e:
         # Fallback to trying to clean the json before serializing
+        packed = json.dumps(
+            json_clean(obj),
+            default=json_default,
+            ensure_ascii=False,
+            allow_nan=False,
+        ).encode("utf8")
+
         warnings.warn(
             f"Message serialization failed with:\n{e}\n"
             "Supporting this message is deprecated in jupyter-client 7, please make "
             "sure your message is JSON-compliant",
             stacklevel=2,
         )
-        return json.dumps(
-            json_clean(obj),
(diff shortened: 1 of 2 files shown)


jupyter_client/session.py
```diff
@@ -92,12 +93,27 @@ def squash_unicode(obj):
...
+    try:
+        return json.dumps(
+            obj,
+            default=json_default,
+            ensure_ascii=False,
+            allow_nan=False,
+        ).encode("utf8")
+    except ValueError as e:
+        # Fallback to trying to clean the json before serializing
+        warnings.warn(
```
minrk, line 105: Maybe show this warning only after the second try at serialization post json_clean succeeds? It might get confusing to get a deprecation warning if it really should fail.

jupyter_client/session.py
```diff
@@ -92,12 +93,27 @@ def squash_unicode(obj):
...
+            f"Message serialization failed with:\n{e}\n"
+            "Supporting this message is deprecated, please make "
```
minrk, line 107:
```suggestion
            "Supporting this message is deprecated in jupyter-client 7, please make "
```

martinRenou:
Thanks a lot for the review @minrk

minrk:
Great, thanks!

blink1073:
@davidbrochart do you want to try an automated release with this change?

davidbrochart:
Sure, but maybe a bit later.

PR #710, opened 2021-10-04
Avoid trying to set the sticky bit of a regular file

ci4ic4:
As it is , the paths list starts with the saved .json file, subsequently setting the sticky bit of its directory. There is no need to set the sticky bit of a regular file, it only generates otherwise a harmless message. No modern UNIX(-like) OS supports sticky bit functionality now, as far as I am aware.

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -167,7 +167,7 @@ def write_connection_file(
     if hasattr(stat, "S_ISVTX"):
         # set the sticky bit on the file and its parent directory
         # to avoid periodic cleanup
-        paths = [fname]
+        paths = []
         runtime_dir = os.path.dirname(fname)
         if runtime_dir:
             paths.append(runtime_dir)


jupyter_client/connect.py
```diff
@@ -167,7 +167,7 @@ def write_connection_file(
     if hasattr(stat, "S_ISVTX"):
         # set the sticky bit on the file and its parent directory
         # to avoid periodic cleanup
-        paths = [fname]
+        paths = []
```
kevin-bates, line 170: Since we're only concerned about the hosting directory, could this be simplified to work directly against `runtime_dir`, fixup comments, etc.?

kevin-bates (commented):
Thanks @ci4ic4 - I agree.  Just had a comment regarding taking this a tad further.

jupyter_client/connect.py
ci4ic4: I've created another pull request which gets rid of the for loop and sets the sticky bit only on the containing directory.

ci4ic4: ... which had a left-over wrong comparison, now removed.

PR #712, opened 2021-10-18
Add support for pending kernels

blink1073:
Add `MultiKernelManager.use_pending_kernels` to use a created kernel before the process has started, based on original discussion in [`jupyter_server`](https://github.com/jupyter-server/jupyter_server/issues/592).

Add a `KernelManager.ready` property to that resolves when the process has started for the first time.

Refuse to restart a kernel that has not been fully started - we could also decide to wait if it has been launched but not fully started.

Consumers using `use_pending_kernels` should wait for the `ready` future before attempting to communicate with the kernel.

This change will correspond to a change in [`jupyter_server`](https://github.com/jupyter-server/jupyter_server/pull/593), where the server will use the `ready` future before connecting the websocket.  

Web clients will have to be updated to handle errors on WebSocket startup and surface them to the user.

As a result, the use of `use_pending_kernels` will be opt-in for now.

- [x] Add `MultiKernelManager.use_pending_kernels`
- [x] Add `KernelManager.ready`
- [x] Make sure a kernel that died at startup can be handled in `MultiKernelManager.shutdown_all`
- [x] Add test for bad kernelspec causing startup errors
- [x] Make sure existing tests pass
- [x] Verify functionality in `jupyter_server` - https://github.com/jupyter-server/jupyter_server/pull/593
- [x] Verify functionality in `JupyterLab` - https://github.com/jupyterlab/jupyterlab/pull/11358

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,6 +112,7 @@ jobs:
           pip freeze
 
       - name: Check types
+        if: ${{ matrix.python-version != '3.6' }}
         run: mypy jupyter_client --exclude '\/tests|kernelspecapp|ioloop|runapp' --install-types --non-interactive
 
       - name: Run the tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,7 +112,6 @@ jobs:
           pip freeze
 
       - name: Check types
-        if: ${{ matrix.python-version != '3.6' }}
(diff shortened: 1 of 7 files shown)


blink1073:
Thanks @davidbrochart.  I marked it as draft for discussion and because I still need to add tests for the new feature.

kevin-bates (approved):
LGTM - thanks Steve!

blink1073:
Thanks for the review!  I'll make a release with this on Monday

PR #723, opened 2021-11-24
Fix install instructions in README

palewire:
I had to add these quotes to get it to work. Without the quotes, the install failed.


blink1073 (approved):
Thanks!

Issue #726, opened 2021-12-13
Renamed default branch to main


blink1073:
Anyone who has the repository already cloned will likely need to update it. These are the commands Github suggests:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

davidbrochart:
Should "master" be replaced with "main"?
https://github.com/jupyter/jupyter_client/blob/52828f0e3ae16757a69a1df6b7234add95ad2950/.github/workflows/check-release.yml#L4

blink1073:
Good catch, yes

vidartf:
Probably ok to close now?

PR #729, opened 2021-12-21
Replace master with main

davidbrochart:
See https://github.com/jupyter/jupyter_client/issues/726#issuecomment-998965728


blink1073 (approved):
Thanks!

PR #731, opened 2021-12-29
Test against all kernels in jupyter kernel test and clean up CI

### LINKED ISSUE #728: Add a CI test against another kernel
### Problem
We are currently only testing against `ipykernel`.  If we were testing against at least one other kernel we can be more confident that our changes are not adversely affecting other kernels.

### Proposed Solution
Add a CI job that installs IRKernel in a conda environment and runs a test that launches IRKernel and executes code.  What we could do is run [test_irkernel.py](`https://github.com/jupyter/jupyter_kernel_test/blob/main/test_irkernel.py`) using a local checkout of `jupyter_kernel_test`, which we are currently doing only for `test_ipykernel.py`.

### Additional context
Use [jupyter_kernel_test](https://github.com/jupyter/jupyter_kernel_test/blob/518c326f5bfc964604f7b798d2b4e6e5179e2902/.github/workflows/tests.yml#L35) as an example for the CI setup.
<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

blink1073:
- Fixes #728 
- Adds testing against Python 3.10
- Uses `base-setup` from `jupyterlab/maintainer-tools` for cleanup

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -17,38 +17,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Install Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
-          architecture: "x64"
-      - name: Install node
-        uses: actions/setup-node@v2
-        with:
-          node-version: "14.x"
-      - name: Get pip cache dir
-        id: pip-cache
-        run: |
-          echo "::set-output name=dir::$(pip cache dir)"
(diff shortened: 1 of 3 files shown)


blink1073:
Adding the Julia test upstream, will rerun this after: https://github.com/jupyter/jupyter_kernel_test/pull/69

blink1073:
I wasn't able to get `IJulia` working, so I went with `xeus-cling`

PR #732, opened 2022-01-05
Further improvements to pending kernels managment

Zsailer:
Two changes in this PR, summarized below. This work is a follow-up to #712 

### Make `shutdown_kernel` a pending state

Makes `shutdown_kernel` also show the kernel in a pending state. Since the KernelManager is still managing a process while it's shutting down, which might take a long time, I think this should show as a pending state too.

### Make `KernelManager` only responsible for reporting kernel pending state 

Also, after working with pending kernels a bit, I believe it makes the most sense to make the `KernelManager` responsible for _reporting_ the kernel's pending state, while the layer that sits above the KernelManager, e.g. `MultiKernelManager`, responsible for _reacting_ to that state.

This essentially means removing *all* `self.ready` checks in the individual KernelManager. For example, 
https://github.com/jupyter/jupyter_client/blob/30823666344c9d03daa7330a81caedd04a6a53a6/jupyter_client/manager.py#L455-L457
should be removed; rather, a MultiKernelManager whose `use_pending_kernels` attribute is `True` would determine if shutdown is a passthrough, etc. 

Another example is https://github.com/jupyter/jupyter_client/blob/30823666344c9d03daa7330a81caedd04a6a53a6/jupyter_client/manager.py#L506-L507
The MultiKernelManager would be responsible for handling what to do when a kernel restart happens while a kernel is in a pending state.

### DIFF
--- a/docs/pending-kernels.rst
+++ b/docs/pending-kernels.rst
@@ -1,0 +1,36 @@
+Pending Kernels
+===============
+
+*Added in 7.1.0*
+
+In scenarios where an kernel takes a long time to start (e.g. kernels running remotely), it can be advantageous to immediately return the kernel's model and ID from key methods like ``.start_kernel()`` and ``.shutdown_kernel()``. The kernel will continue its task without blocking other managerial actions.
+
+This intermediate state is called a **"pending kernel"**.
+
+How they work
+-------------
+
+When ``.start_kernel()`` or ``.shutdown_kernel()`` is called, a ``Future`` is created under the ``KernelManager.ready`` property. This property can be awaited anytime to ensure that the kernel moves out of its pending state, e.g.:
+
+.. code-block:: python
+
+    # await a Kernel Manager's `.ready` property to
+    # block further action until the kernel is out
+    # of its pending state.
+    await kernel_manager.ready
+
+Once the kernel is finished pending, ``.ready.done()`` will be ``True`` and either 1) ``.ready.result()`` will return ``None`` or 2) ``.ready.exception()`` will return a raised exception
+
+Using pending kernels
+---------------------
+
+The most common way to interact with pending kernels is through the ``MultiKernelManager``—the object that manages a collection of kernels—by setting its ``use_pending_kernels`` trait to ``True``. Pending kernels are "opt-in"; they are not used by default in the ``MultiKernelManager``.
+
+When ``use_pending_kernels`` is ``True``, the following changes are made to the ``MultiKernelManager``:
+
+1. ``start_kernel`` and ``stop_kernel`` return immediately while running the pending task in a background thread.
+2. The following methods raise a ``RuntimeError`` if a kernel is pending:
+    * ``restart_kernel``
+    * ``interrupt_kernel``
+    * ``shutdown_kernel``
+3. ``shutdown_all`` will wait for all pending kernels to become ready before attempting to shut them down.
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -239,10 +239,11 @@ class MultiKernelManager(LoggingConfigurable):
         restart : bool
             Will the kernel be restarted?
         """
-        # If the kernel in a pending state?
         self.log.info("Kernel shutdown: %s" % kernel_id)
+        # If we're using pending kernels, block shutdown when a kernel is pending.
         if self._using_pending_kernels() and kernel_id in self._pending_kernels:
             raise RuntimeError("Kernel is in a pending state. Cannot shutdown.")
+        # If the kernel isn't in a ready state, wait for it to be ready.
         elif kernel_id in self._pending_kernels:
             kernel = self._pending_kernels[kernel_id]
             try:
@@ -251,6 +252,10 @@ class MultiKernelManager(LoggingConfigurable):
                 self.remove_kernel(kernel_id)
                 return
         km = self.get_kernel(kernel_id)
+        # If a pending kernel raised an exception, remove it.
+        if km.ready.exception():
+            self.remove_kernel(kernel_id)
+            return
         stopper = ensure_async(km.shutdown_kernel(now, restart))
         fut = asyncio.ensure_future(self._remove_kernel_when_ready(kernel_id, stopper))
         self._pending_kernels[kernel_id] = fut
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -222,6 +222,22 @@ class MultiKernelManager(LoggingConfigurable):
 
     start_kernel = run_sync(_async_start_kernel)
 
+    async def _shutdown_kernel_when_ready(
+        self,
+        kernel_id: str,
+        now: t.Optional[bool] = False,
+        restart: t.Optional[bool] = False,
+    ) -> None:
+        """Wait for a pending kernel to be ready
+        before shutting the kernel down.
+        """
+        # Only do this if using pending kernels
+        if self._using_pending_kernels():
+            kernel = self._kernels[kernel_id]
+            await kernel.ready
+        # Once out of a pending state, we can call shutdown.
+        await ensure_async(self.shutdown_kernel(kernel_id, now=now, restart=restart))
+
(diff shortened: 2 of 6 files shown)


blink1073:
This seems reasonable to me.  Zach and I discussed this today and agreed that we should release a version of `jupyter_server` that supports these changes, get downstream tests here passing, and then release a minor release with these refinements.

jupyter_client/tests/test_multikernelmanager.py
```diff
@@ -29,6 +29,14 @@
 TIMEOUT = 30
 
 
+async def now(awaitable):
+    """Use this function ensure that this awaitable
+    happens before other awaitables defined after it.
+    """
+    (out,) = await asyncio.gather(awaitable)
```
kevin-bates, line 36: I guess I don't understand why this method is necessary - especially for a single-item gather.  Isn't awaiting the awaitable sufficient to prevent the execution of follow-on awaitables in the same execution path?

kevin-bates (commented):
Thanks @Zsailer - this looks good.  I just had a question regarding the `now` method in the tests.

It might be good to also update the help-string of the [`ready` property](https://github.com/jupyter/jupyter_client/blob/30823666344c9d03daa7330a81caedd04a6a53a6/jupyter_client/manager.py#L151).  Perhaps something like:
```python
"""A future that resolves when the kernel has completed its startup or shutdown"""
```

jupyter_client/tests/test_multikernelmanager.py
Zsailer: Great question, @kevin-bates. 

That's what's supposed to happen; however, I'm seeing some weird behavior that *I think* is coming from the `gen_test` decorator. Basically, I'm not seeing the `await`s happening in the order they are defined.

Weirdly, if I replace the async/await syntax with the old `yield` syntax, these tests work fine (no `now` needed). Again, this suggest some weird interference happening from Tornado's async testing. I'm not sure how to fix this.

Zsailer: By enforcing a `gather` call in each await statement, everything happens in the order they are called. Super strange!

kevin-bates: Thanks Zach.  That is interesting.

Zsailer: Looks like nested coroutines can be a bit difficult in unit tests. It appears that inner coroutines never get scheduled even when the outer coroutine is `await`ed. Adding `ensure_future` in multiple places in the tests ensures that these coroutines get scheduled and tests pass again.

Zsailer:
It looks like all tests are (finally) passing except the downstream tests. For that, https://github.com/jupyter-server/jupyter_server/pull/654 is required.

I'll work on getting that merged and then ask for a final review here.

Zsailer:
I've also added some documentation for pending kernels to this PR

jupyter_client/multikernelmanager.py
```diff
@@ -258,13 +315,18 @@ def remove_kernel(self, kernel_id: str) -> KernelManager:
...
+        # When using "shutdown all", all pending kernels
+        # should be awaited before exiting this method.
+        if self._using_pending_kernels():
+            for km in self._kernels.values():
+                await km.ready
 
     shutdown_all = run_sync(_async_shutdown_all)
 
-    @kernel_method
+    # @kernel_method
```
blink1073, line 329: Can this be reinstated?

Zsailer: No, the commented line needs to be removed.

Zsailer:
I'm not sure why the `nbclient` downstream tests are failing. When I run these locally, they run successfully.

davidbrochart:
The nbclient failures should be fixed by https://github.com/jupyter/nbclient/pull/190.
I could release nbclient if you want to be sure.

davidbrochart:
nbclient v0.5.10 is on PyPI with the fix.

Zsailer:
Thanks, @davidbrochart! I really appreciate it!

Zsailer:
Waiting on https://github.com/jupyter-server/jupyter_server/pull/662 to be merged and released to fix the downstream tests.

Zsailer:
All green!

Zsailer:
This was discussed extensively at the last Jupyter Server meeting: https://github.com/jupyter-server/team-compass/issues/15#issuecomment-1012349510

Merging away!

docs/pending-kernels.rst
```diff
@@ -0,0 +1,36 @@
+Pending Kernels
+===============
+
+*Added in 7.1.0*
+
+In scenarios where an kernel takes a long time to start (e.g. kernels running remotely), it can be advantageous to immediately return the kernel's model and ID from key methods like ``.start_kernel()`` and ``.shutdown_kernel()``. The kernel will continue its task without blocking other managerial actions.
```
kevin-bates, line 6:
```suggestion
In scenarios where a kernel takes a long time to start (e.g. kernels running remotely), it can be advantageous to immediately return the kernel's model and ID from key methods like ``.start_kernel()`` and ``.shutdown_kernel()``. The kernel will continue its task without blocking other managerial actions.
```

jupyter_client/multikernelmanager.py
```diff
@@ -97,7 +97,12 @@ def create_kernel_manager(*args, **kwargs) -> KernelManager:
 
     context = Instance("zmq.Context")
 
-    _starting_kernels = Dict()
+    _pending_kernels = Dict()
+
+    @property
+    def _starting_kernels(self):
+        """A shim for backwards compatibility."""
+        return self._pending_kernels
```
kevin-bates, line 105: Should this be marked as deprecated?

jupyter_client/multikernelmanager.py
```diff
@@ -215,15 +256,31 @@ async def _async_shutdown_kernel(
             Will the kernel be restarted?
         """
         self.log.info("Kernel shutdown: %s" % kernel_id)
-        if kernel_id in self._starting_kernels:
+        # If we're using pending kernels, block shutdown when a kernel is pending.
+        if self._using_pending_kernels() and kernel_id in self._pending_kernels:
+            raise RuntimeError("Kernel is in a pending state. Cannot shutdown.")
+        # If the kernel is still starting, wait for it to be ready.
+        elif kernel_id in self._starting_kernels:
```
kevin-bates, line 263: Is the use of `_starting_kernels` intentional?  Technically, this could also contain kernels that are shutting down now, but only when pending kernels are enabled - so I suspect this was for hinting that, in this case, the kernel can only be starting.

Perhaps we could be more explicit (and save an existence check [nit]) via: 
```python
        if kernel_id in self._pending_kernels:
            if self._using_pending_kernels():
                 raise RuntimeError("Kernel is in a pending state. Cannot shutdown.")
            else:  # kernel is still starting, wait for its startup
                kernel = self._pending_kernels[kernel_id]
                try:
                    await kernel
                except Exception:
                    self.remove_kernel(kernel_id)
```

jupyter_client/multikernelmanager.py
```diff
@@ -215,15 +256,31 @@ async def _async_shutdown_kernel(
...
+            kernel = self._starting_kernels[kernel_id]
             try:
-                await self._starting_kernels[kernel_id]
+                await kernel
             except Exception:
                 self.remove_kernel(kernel_id)
```
kevin-bates, line 268: Do we want to update the `_pending_kernels` list as well here?

kevin-bates (commented):
Just had a couple of suggestions and comments, but this looks good.

kevin-bates:
Hmm - sorry, it looks like this was merged while I was reviewing.  The comments aren't earth-shattering so I'll leave it to you to determine if they have any merit.

Thanks for all the work on this @Zsailer - good stuff!

jtpio:
Looks like the `7.1.1` release of `jupyter_client` affected the JupyterLab tests: https://github.com/jupyterlab/jupyterlab/issues/11886

The check is defined here:


Pinning to `7.1.0` fixes it in https://github.com/jupyterlab/jupyterlab/pull/11887.

Reporting here for reference.

PR #733, opened 2022-01-09
Update consoleapp.py

you-n-g:
Supporting setting sshkey via command line 

Example
```
jupyter qtconsole --existing C:\Users\<user>\qtconfig.json --ssh <user>@<host>:10022 --sshkey <key_path>
```


kevin-bates (approved):
@you-n-g - This seems like a useful change and in line with the existing `--ssh` alias.  Thank you.

PR #740, opened 2022-01-21
Await `kernel.ready` in `_async_shutdown_kernel`

jtpio:
Attempt at fixing https://github.com/jupyterlab/jupyterlab/issues/11886.

According to the comment above:

https://github.com/jupyter/jupyter_client/blob/263173095ceab87a572937f21f843886ffe356f1/jupyter_client/multikernelmanager.py#L262

This branch of the code should be waiting for the kernel to be ready.

This change seems to be fixing the issue noticed in JupyterLab in https://github.com/jupyterlab/jupyterlab/issues/11886.

Example run: https://github.com/jupyterlab/jupyterlab/runs/4894046802?check_suite_focus=true from https://github.com/jupyterlab/jupyterlab/pull/11887

### DIFF
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -264,7 +264,8 @@ class MultiKernelManager(LoggingConfigurable):
             kernel = self._pending_kernels[kernel_id]
             try:
                 await kernel
-                await kernel.ready
+                km = self.get_kernel(kernel_id)
+                await km.ready
             except Exception:
                 self.remove_kernel(kernel_id)
                 return
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -260,10 +260,11 @@ class MultiKernelManager(LoggingConfigurable):
         if self._using_pending_kernels() and kernel_id in self._pending_kernels:
             raise RuntimeError("Kernel is in a pending state. Cannot shutdown.")
         # If the kernel is still starting, wait for it to be ready.
-        elif kernel_id in self._starting_kernels:
-            kernel = self._starting_kernels[kernel_id]
+        elif kernel_id in self._pending_kernels:
+            kernel = self._pending_kernels[kernel_id]
             try:
                 await kernel
+                await kernel.ready
             except Exception:
                 self.remove_kernel(kernel_id)
                 return


jupyter_client/multikernelmanager.py
```diff
@@ -260,10 +260,11 @@ async def _async_shutdown_kernel(
         if self._using_pending_kernels() and kernel_id in self._pending_kernels:
             raise RuntimeError("Kernel is in a pending state. Cannot shutdown.")
         # If the kernel is still starting, wait for it to be ready.
-        elif kernel_id in self._starting_kernels:
-            kernel = self._starting_kernels[kernel_id]
+        elif kernel_id in self._pending_kernels:
```
jtpio, line 263: Since `_starting_kernels` is a shim for `_pending_kernels` defined here:

https://github.com/jupyter/jupyter_client/blob/263173095ceab87a572937f21f843886ffe356f1/jupyter_client/multikernelmanager.py#L102-L105

We might as well use `_pending_kernels` directly.

jtpio:
Looks like the CI checks have been stuck for a couple of hours.

Gonna close and reopen in a couple of minutes.

blink1073:
I just kicked a recent [build](https://github.com/jupyter/jupyter_client/actions/runs/1699288179) and it passes, so the CI failures are relevant.

jupyter_client/multikernelmanager.py
```diff
@@ -260,10 +260,11 @@ async def _async_shutdown_kernel(
...
+            kernel = self._pending_kernels[kernel_id]
             try:
                 await kernel
+                await kernel.ready
```
Zsailer, line 267: I think I see the issue—you are on the right track. The naming of things is getting a little messy.

In this line, `kernel` is not the kernel manager, like you'd expect. It's the `Future` returned by `_add_kernel_when_ready`. Waiting this `Future` doesn't ensure that kernel has started, it just ensures that the `start_kernel` method returned. When pending kernels are enabled, `start_kernel` returns immediately, but the kernel might not be ready. To check that the kernel is ready, we need to call `await ... .ready` on the actual kernel manager.


Calling `.ready` on this future is redundant. It might be falsely helping the JupyterLab tests pass because it slows the method enough to dodge the race condition.

Zsailer, line 267:
```suggestion
                km = self.get_kernel(kernel_id)
                await km.ready
```

blink1073 (approved):
Thanks y'all!

Zsailer:
Thank you, @jtpio, for helping with this! This saved me a ton of time!

jtpio:
Thank you @Zsailer for finishing this one!

PR #743, opened 2022-02-03
Prefer sending signals to kernel process group

kevin-bates:
Per [this comment in #104](https://github.com/jupyter/jupyter_client/issues/104#issuecomment-1027506791), we should restore the behavior that signaling the process group is preferred over the Popen `kill()` and `terminate()` methods (which essentially signal the process).  This pull request updates the LocalProvisioner to send `SIGKILL`/`SIGTERM` to the process group, prior to falling back to Popen `kill()`/`terminate()`, respectively.

As also noted in the comment, this functionality is masked by the fact that we still prefer a message-based shutdown unless the message-based shutdown fails or an auto-restart event occurs (when the kernel process can no longer be tracked), in which case the "immediate" form of restart occurs (using the signal-based approach).

### DIFF
--- a/jupyter_client/provisioning/local_provisioner.py
+++ b/jupyter_client/provisioning/local_provisioner.py
@@ -85,34 +85,54 @@ class LocalProvisioner(KernelProvisionerBase):
                     os.killpg(self.pgid, signum)
                     return
                 except OSError:
-                    pass
-            try:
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Nice, thanks!

PR #749, opened 2022-02-28
Add session async recv

davidbrochart:
We are trying to drop Tornado from ipykernel in https://github.com/ipython/ipykernel/pull/876. Because we now use asyncio's event loop, a socket's `recv_multipart` must be awaited.
This change consists of adding `Session.async_recv`, which is the async version of `Session.recv`, so that [the debugger can await it](https://github.com/ipython/ipykernel/blob/d6744f9e423dacc6b317b1d31805304e89cbec5d/ipykernel/debugger.py#L387).

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -936,6 +936,46 @@ class Session(Configurable):
             # TODO: handle it
             raise e
 
+    async def async_recv(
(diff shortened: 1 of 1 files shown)


blink1073:
This was handled in #835

PR #751, opened 2022-03-06
Handle warnings in tests

blink1073:
Done so far:

- Close fds in `LocalProvisioner.wait` to avoid unclosed fids
- Destroy kernel client's context when calling `stop_channels` to avoid unclosed zmq context
- Use separate future for shutdown to avoid conflict: `kernel.shutdown_ready`
- Add guard around interrupting in `KernelManager.shutdown`
- Clean up handling of event loops
- Increase the number of available fds in tests
- Miscellaneous teardown cleanups

So far the only warnings we can't easily work around are due to https://github.com/pytest-dev/pytest-asyncio/issues/77.  We'd have to rewrite all of the tests not to use `TestCase`.


.github/workflows/main.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         run: mypy jupyter_client --exclude '\/tests|kernelspecapp|ioloop|runapp' --install-types --non-interactive
 
       - name: Run the tests with coverage
-        run: pytest --cov jupyter_client -v jupyter_client
+        run: pytest --cov jupyter_client -vv jupyter_client || || pytest --cov jupyter_client -vv jupyter_client --lf
```
kevin-bates: Extra OR?
```suggestion
        run: pytest --cov jupyter_client -vv jupyter_client || pytest --cov jupyter_client -vv jupyter_client --lf
```

jupyter_client/client.py
```diff
@@ -311,6 +318,8 @@ def stop_channels(self) -> None:
             self.hb_channel.stop()
         if self.control_channel.is_alive():
             self.control_channel.stop()
+        if self._created_context:
+            self.context.destroy()
```
kevin-bates, line 322: Should `self._create_context` be unset here?

jupyter_client/manager.py
```diff
@@ -476,7 +489,10 @@ async def _async_shutdown_kernel(self, now: bool = False, restart: bool = False)
         # Stop monitoring for restarting while we shutdown.
         self.stop_restarter()
 
-        await ensure_async(self.interrupt_kernel())
+        # If the kernel has already started, interrupt it to give it a
+        # chance to clean up.
+        if self.has_kernel:
+            await ensure_async(self.interrupt_kernel())
```
kevin-bates, line 495: Nice - make _this_ call to interrupt() _conditional_ on start status.  Good catch.

jupyter_client/tests/conftest.py
```diff
@@ -7,9 +7,31 @@
...
+    DEFAULT_SOFT = 4096
+    if hard >= DEFAULT_SOFT:
+        soft = DEFAULT_SOFT
+
+    old_soft, old_hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    hard = old_hard
+    if old_soft < soft:
+        if hard < soft:
+            hard = soft
+        resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
```
kevin-bates, line 32: I find this a little difficult to understand.  I think the double calls to `getrlimit` are part of the confusion.  Perhaps a comment as to the goal would be helpful.  It appears to ensure a minimal soft limit of `DEFAULT_SOFT` if the current hard limit is at least that much.

It seems like the `hard < soft` (L30) condition will never be true because `old_soft < soft` (L29) can only be true if `soft` was increased to `DEFAULT_SOFT`, which implies `hard >= soft`, which leads me to wonder some kind of update to `hard` is missing.

kevin-bates (commented):
Thanks for spending the necessary time on this @blink1073 - it's non-trivial and much appreciated!  Just had a couple of comments.

Should we open an issue to move the remaining tests to pytest to rid ourselves of `TestCase`?  Perhaps someone with some bandwidth could take that on as a good intro?

.github/workflows/main.yml
blink1073: Already fixed  ;)

jupyter_client/manager.py
blink1073: 👍🏼

jupyter_client/tests/conftest.py
blink1073: Cleaned up

blink1073:
> Should we open an issue to move the remaining tests to pytest to rid ourselves of TestCase? Perhaps someone with some bandwidth could take that on as a good intro?

Yeah that sounds good

blink1073:
It looks like this breaks the server tests, so I'll have to do both PRs at once.

blink1073:
Okay, this part is done, stopping for today.

jupyter_client/manager.py
```diff
@@ -52,33 +52,38 @@ class _ShutdownStatus(Enum):
...
+    def inner(method):
+        @functools.wraps(method)
+        async def wrapper(self, *args, **kwargs):
+            # Create a future for the decorated method
+            try:
+                asyncio.get_running_loop()
+                fut = Future()
+            except RuntimeError:
+                # No event loop running, use concurrent future
+                fut = CFuture()
```
minrk, line 70: If this is inside an `async def`, can this except branch ever occur? There should always be a running loop from inside a coroutine.

The previous strategy of setting `_ready` _outside_ a coroutine might not have a running loop, but now that you've moved it into the coroutine (good, I think), there should always be a running loop.

Returning two types of Future depending on context would be tricky because they aren't interchangeable (CFuture is not awaitable without calling `asyncio.wrap_future(cfuture)`).

minrk: An important difference of this strategy, though: the attribute will not be set until an arbitrarily-later point in time due to asyncness. The previous strategy ensured `._ready` was defined _immediately_ before any waiting, whereas putting it in the coroutine means the attribute will _not_ be defined immediately.

Two new things to consider (may well both be covered):

1. previous waits for `self._ready` that may now be called when `_ready` is not yet
2. can `_ready` be set and then re-set? If so, waits for `_ready` may get a 'stale' state before the new `_ready` future is attached. This can be avoided with a non-async `delattr` prior to the async bit.

blink1073: Originally `_ready` was only defined in the constructor, which is why we had the fallback option.  It seems like these futures are becoming problematic in general.  Maybe what we really want is a state and a signal when the state changes, like we have on the JavaScript side.

blink1073: Alternatively, if we want to define `ready` and `shutdown_ready` as "the most recent" starting and shutting down futures, we could also add an `is_ready` and `is_shutdown_ready` for when there are no pending of either.  We would use this trick that tornado used to work around the new deprecation warnings:

```python
loop = asyncio.get_event_loop_policy().get_event_loop()   # this always returns the same loop
future = aysncio.Future(loop=loop)
```

We could then make `.ready` and `.shutdown` ready promises only available when using the `AsyncKernelManager`.

blink1073: I think part of the problem is that `jupyter_client` is a library, not an application, in the same way that `tornado` is.   It looks like we also can't rely on `asyncio_run` in the future, since it relies on `get_event_loop` and `set_event_loop`.  I think here's a sketch of what we need to do:

- The base classes will be asynchronous and call the the private async versions of functions directly as needed.  We will only use futures and tasks from within async methods.
- The synchronous classes will work by wrapping the asynchronous public methods using a decorator that runs a new private asyncio loop as `return asyncio.new_event_loop().run_until_complete(async_method)`.
- We remove the `.ready` future and add new methods called `async def wait_for_pending_start()` and `async def wait_for_pending_shutdown()` to handle pending kernels.  
 
With this new structure, the synchronous managers could also support pending kernels, breaking it into two synchronous steps.

blink1073: We can handle the ready future in this PR and defer the other parts to a separate refactor PR.

minrk: btw, concurrent.futures are generally more compatible and flexible than asyncio futures - they are threadsafe, reusable across loops, etc. If you always use a cf.Future, you can await it in a coroutine with `asyncio.wrap_future(cf_future)`. That may be the simplest solution here. It's the _inconsistency_ that I saw as a problem (`self._ready` _may_ be awaitable). Something like:

```python
# earlier, _not_ in a coroutine
self._ready = cf.Future()

async def wait_for_ready(self):
    if self._ready is None:
        raise RuntimeError("haven't started yet!")
    # wrap cf.Future self._ready to make it awaitable
    await asyncio.wrap_future(self._ready)
```

ought to be more reliable.


To always use new event loops (or `asyncio.run`) for each sync method call, one important consideration is that all your methods completely resolve by the end of the coroutine (i.e. not hold references to the current loop for later calls, which means requiring [pyzmq 22.2](https://pyzmq.readthedocs.io/en/v22.2.0/changelog.html#id1), among other things). A possibly simpler alternative is to maintain your _own_ event loop reference. `asyncio.get_event_loop()` is deprecated because it did too many things, but that doesn't mean _nobody_ should hold onto persistent loop references. It just means that they don't want to track a _global_ not-running loop. If each instance had a `self._loop = asyncio.new_event_loop()` and always ran the sync wrappers in that loop (`self._loop.run_until_complete(coro())`), it should work fine. That avoids the multiple-loop problem for sync uses because it's still a single persistent loop. Each instance may happen to use a different loop, but that should not be a problem in the sync case.

blink1073: Sweet, thanks for the insights Min.  That sounds like a good approach.

blink1073: More thoughts on ready future handling:

- Remove the public ready attribute(s).
- Internally, store current ready and pending flags for start and shutdown.
- In `start()`, check for `_start_pending` and error if true.  Set `_start_pending` and set a new `_start_ready` future and clear `_start_pending` when ready.
- In `shutdown()`, check for `_shutdown_pending` and bail if true.  Set `_shutdown_pending` and set a new `_shutdown_ready` future.  Clear `_shutdown_pending` when ready.
- In `wait_for_start_ready()` and `wait_for_shutdown_ready()` store current ready, and wait for it.  If new current ready is different from the stored one, recurse.

blink1073:
Rather than introduce more flags and more confusion, I propose we actually use a simple [state machine](https://dev.to/karn/building-a-simple-state-machine-in-python) as @echarles suggested.

Currently server adds "execution_state" to the manager object based on [kernel status](https://github.com/jupyter-server/jupyter_server/blob/49dfe2e331277a8e57e5e475234fd3252717e4c3/jupyter_server/services/kernels/kernelmanager.py#L521) messages.  I think eventually we should move that logic here, but it doesn't need to be in this PR.

The states used here could mirror the ones used in [`JupyterLab`](https://github.com/jupyterlab/jupyterlab/blob/5f4e3fabf33bed101e41708021a245ca59410511/packages/services/src/kernel/messages.ts#L516) :

  - Unknown
  - Starting
  - Running (Maps to Idle and Busy execution states)
  - Terminating
  - Restarting
  - Dead (with optional traceback)

We are `Unknown` to start.   State transitions are as follows:

  - Unknown -> Starting
  - Starting -> Running or Dead
  - Running -> Terminating, Restarting, or Dead
  - Terminating -> Dead
  - Restarting -> Running or Dead
  - Dead -> Starting

Then we have a `state: Unicode()` trait and a convenience `wait_for_state()` method that takes 
an optional desired state or returns for any state change.  It would also raise an exception if it
gets to `Dead` state and it wasn't the expected state, along with the traceback.

The `state` trait can also be `observed` directly.

We also have an `exception: Unicode()` trait that is set when an exception leads to a `Dead` status.

echarles:
> Rather than introduce more flags and more confusion, I propose we actually use a simple [state machine](https://dev.to/karn/building-a-simple-state-machine-in-python) as @echarles suggested.

Maybe the first step would be to draw and share that state machine that would be highly beneficial to jupyter client, but also to e.g. the jupyterlab services package evolution.

Having a SVG format for that drawing would allow everyone to use their favorite editor (e.g. I use inkscape).

davidbrochart:
I have not looked at this PR in details, but it seems to me that supporting a sync API is becoming problematic. At the time we switched jupyter_client to async, I had the feeling that keeping a sync API was kind of mandatory, although it involved using nest-asyncio which is not without problems. But today, could jupyter_client be only async? That would simplify the library a lot.

blink1073:
Good call @echarles.  I think we should adopt the [`transitions`](https://github.com/pytransitions/transitions) library.  It is MIT licensed, already on conda-forge, and lets you produce [`diagrams`](https://github.com/pytransitions/transitions#-diagrams) from your state machine.  

We could add a CI job that create the diagram and compares it to the one in the documentation, and fails and asks you to rebuild if it has changed.

blink1073:
@davidbrochart, I think there are still valid use cases for having a synchronous manager.  What we could say is that only the async class is meant to be overridable, and offer a function that wraps the async class to become synchronous.  We would have a section like:

```
# These methods are meant to be overridden by subclasses
async def _start_kernel(...)
```

And the public functions would be async in the base class, but made synchronous in the sync class.
We would create our own async manager(s) using this wrapper.
I think this would have to be an 8.0 release.

echarles:
> Good call @echarles. I think we should adopt the [transitions](https://github.com/pytransitions/transitions) library. It is MIT licensed, already on conda-forge, and lets you produce [diagrams](https://github.com/pytransitions/transitions#-diagrams) from your state machine.

Code-to-doc, that sounds exciting. Looks like `transitions` uses has not svg native support and uses graphviz to generate the diagram. As graphviz can [generate SVG](https://graphviz.org/docs/outputs/svg/), that is all good.

> We could add a CI job that create the diagram and compares it to the one in the documentation, and fails and asks you to rebuild if it has changed.

That sounds exiciting.

BTW I have been looking recently at [xstate](https://xstate.js.org) on the frontend are. Looks like state machine architecture is trendy atm.

echarles:
> @davidbrochart, I think there are still valid use cases for having a synchronous manager

+1 Although async + sync is harder to maintain in `jupyter-client`, it benefits basic consumers of this library who don't have to worry about `asynchronicity`.

minrk:
I think keeping sync is important. 

It's also possible to avoid nest_asyncio if you run 'asyncio-for-sync' in a dedicated IO thread. `call_soon_threadsafe` and concurrent.futures.Future make this pretty doable, and in my experience simpler and more robust than using nest_asyncio. ipyparallel's sync client has worked that way for a long time.

blink1073:
I opened https://github.com/jupyter/jupyter_client/issues/755 to track the `nest_asyncio` problem.

blink1073:
Closing this one in favor of #760 and a follow up in #763

PR #752, opened 2022-03-07
Json packer: handle TypeError and fallback to old json_clean

martinRenou:
`TypeError` are raised when the dictionaries contain invalid keys, which `json_clean` used to handle so we can fallback to it.

cc. @minrk


blink1073 (approved):
Thanks!

PR #754, opened 2022-03-14
Test coroutine interruption with async client

davidbrochart:
See https://github.com/jupyterlab/jupyterlab/issues/12194.

It looks like interrupting a cell execution which consists of awaiting a coroutine, e.g.:
```python
import asyncio

async def main():
    await asyncio.sleep(5)

await main()
```
doesn't take effect immediately, but once the coroutine completes execution. This only happens with an async client. It doesn't happen on the Notebook because I think it uses a blocking client.


davidbrochart:
Actually it looks more like an ipykernel issue, since akernel stops immediately after interruption.
I tried xeus-python (not raw xeus-python since it doesn't allow awaiting at the top-level), but interrupting such a coroutine leads in the kernel dying.

rgbkrk:
Closing as stale. Thanks @davidbrochart

PR #758, opened 2022-03-28
Mock is not needed

hroncok:
Since 3f8599f56d1d898e54a8116550b024226c985ac4 we always use unittest.mock


davidbrochart:
Thanks @hroncok !

PR #760, opened 2022-03-29
Handle Warnings

blink1073:
Reboot of #751 without any changes to the `KernelManager`.


blink1073:
Downstream test failure is due to https://github.com/jupyterlab/maintainer-tools/issues/74

jupyter_client/utils.py
```diff
@@ -11,14 +11,14 @@
 def run_sync(coro):
     def wrapped(*args, **kwargs):
         try:
-            loop = asyncio.get_event_loop()
+            loop = asyncio.get_running_loop()
```
Carreau, line 14: This and only this line seem to be breaking some projects. 
in particular https://github.com/Carreau/inplace_restarter.

It seem to be because `asyncio.get_event_loop()` can return the current event loop even if it's not (yet) running. 

I'm not sure what a proper fix coulc be.

blink1073: Thanks for the heads up, I put a workaround in [`0a2c228` (#772)](https://github.com/jupyter/jupyter_client/pull/772/commits/0a2c2289347771d69cac7dda69e5a29dc4b445a9)

Carreau: Thanks , I also fixed my project upstream, I'll try to have a look at #772

PR #766, opened 2022-04-07
Include py.typed file

blink1073:
cf https://peps.python.org/pep-0561/

cf https://github.com/jupyter-server/jupyter_server/pull/779


davidbrochart:
Thanks!

Issue #763, opened 2022-03-30
Add State Machine Handling to KernelManager


blink1073:
In #751 we explored better internal state management for the `KernelManager` class.

We decided there to adopt the [`transitions`](https://github.com/pytransitions/transitions) library.  It is MIT licensed, already on conda-forge, and lets you produce [`diagrams`](https://github.com/pytransitions/transitions#-diagrams) from your state machine.
We can produce the state machine graph image automatically, backed by CI. 

A rough sketch looks like:

Currently server adds "execution_state" to the manager object based on [kernel status](https://github.com/jupyter-server/jupyter_server/blob/49dfe2e331277a8e57e5e475234fd3252717e4c3/jupyter_server/services/kernels/kernelmanager.py#L521) messages.  I think eventually we should move that logic here, but it doesn't need to be in this PR.

The states used here could mirror the ones used in [`JupyterLab`](https://github.com/jupyterlab/jupyterlab/blob/5f4e3fabf33bed101e41708021a245ca59410511/packages/services/src/kernel/messages.ts#L516) :

  - Unknown
  - Starting
  - Running (Maps to Idle and Busy execution states)
  - Terminating
  - Restarting
  - Dead (with optional traceback)

We are `Unknown` to start.   State transitions are as follows:

  - Unknown -> Starting
  - Starting -> Running or Dead
  - Running -> Terminating, Restarting, or Dead
  - Terminating -> Dead
  - Restarting -> Running or Dead
  - Dead -> Starting

Then we have a `state: Unicode()` trait and a convenience `wait_for_state()` method that takes 
an optional desired state or returns for any state change.  It would also raise an exception if it
gets to `Dead` state and it wasn't the expected state, along with the traceback.

The `state` trait can also be `observed` directly.

We also have an `exception: Unicode()` trait that is set when an exception leads to a `Dead` status.

kevin-bates:
This is interesting, thanks for opening this issue @blink1073.

If we were to include the _triggers_ that initiate these transitions, it seems like we'd have the following ...

| Trigger | From State | To State | Notes |
|---------|------------|----------|-------|
| `start_kernel()` | Unknown | Starting | Successful startup initiation|
| `ready.set_result()`| Starting | Running | Successful startup completion |
| `ready.set_exception()`| Starting | Dead | Unsuccessful startup |
| `shutdown_kernel()` | Running | Terminating | Successful shutdown initiation|
[...]

PR #782, opened 2022-05-02
Manage Kernel status/state with a state machine

Zsailer:
Implements #763

This is an early draft of the work. I wanted to open-source this early to encourage feedback/discussion.

Adds a state machine to the default `KernelManager` implementation . This exposes an API to see what state the kernel is in at any given time. This replaces the previous "pending kernel" API we added recently in favor of something more robust.

You'll notice it takes very little code. The `transitions` library does most of the heavy lifting here. 

Todo
------

- [ ] Deprecate the `.ready` API for pending kernels; add a backwards compatible API for the deprecation cycle.
- [ ] Add unit tests
- [ ] Add documentation

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -2,7 +2,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import asyncio
-import functools
 import os
 import re
 import signal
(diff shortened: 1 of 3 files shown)


blink1073:
We decided to move this work to `jupyter_server` (post 2.0), and leave the `.ready` promise as-is.

PR #785, opened 2022-05-08
Check that channels exist before asking if they are alive

ccordoba12:
This extra check fixes issue https://github.com/spyder-ide/spyder/issues/17776 in Spyder, which appeared with Jupyter-client 7.3.0.

It only happens when Spyder is closed, but I don't understand very well why. Since it's a small patch, I hope it'll be accepted so that we don't have to subclass `QtKernelClient` on our side for this.

### DIFF
--- a/jupyter_client/client.py
+++ b/jupyter_client/client.py
@@ -337,11 +337,11 @@ class KernelClient(ConnectionFileMixin):
     def channels_running(self) -> bool:
         """Are any of the channels created and running?"""
         return (
-            self.shell_channel.is_alive()
-            or self.iopub_channel.is_alive()
-            or self.stdin_channel.is_alive()
-            or self.hb_channel.is_alive()
-            or self.control_channel.is_alive()
+            (self._shell_channel and self.shell_channel.is_alive())
+            or (self._iopub_channel and self.iopub_channel.is_alive())
+            or (self._stdin_channel and self.stdin_channel.is_alive())
+            or (self._hb_channel and self.hb_channel.is_alive())
+            or (self._control_channel and self.control_channel.is_alive())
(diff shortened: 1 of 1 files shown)


jupyter_client/client.py
```diff
@@ -337,11 +337,11 @@ def stop_channels(self) -> None:
...
-            self.shell_channel.is_alive()
-            or self.iopub_channel.is_alive()
-            or self.stdin_channel.is_alive()
-            or self.hb_channel.is_alive()
-            or self.control_channel.is_alive()
+            (self._shell_channel and self.shell_channel.is_alive())
+            or (self._iopub_channel and self.iopub_channel.is_alive())
+            or (self._stdin_channel and self.stdin_channel.is_alive())
+            or (self._hb_channel and self.hb_channel.is_alive())
+            or (self._control_channel and self.control_channel.is_alive())
```
davidbrochart, line 344: These changes seem more in line with the docstring `"Are any of the channels created and running?"`, e.g. with only `self.shell_channel.is_alive()` we are creating a shell channel if it doesn't already exists, so we are only checking if it is running.
Your changes ensure that a channel already exists before checking if it is running, which seems like the right thing to do IMO.

davidbrochart (approved):
Thanks @ccordoba12, although I'm not sure how this method is used in the wild, I think that your changes are correct.

jupyter_client/client.py
ccordoba12: Thanks @davidbrochart!

ccordoba12:
Thanks a lot for your prompt reply @davidbrochart and for releasing 7.3.1 with my fix!

PR #794, opened 2022-05-31
Ensure KernelClient's API is complete

davidbrochart:
Otherwise type checkers complain about a kernel client not having the required methods.


rgbkrk:
Would you want to dust this one off? I'd love to see more types.

PR #803, opened 2022-06-08
Revert latest changes to `ThreadedZMQSocketChannel` because they break Qtconsole

ccordoba12:
- Those changes were done in commit https://github.com/jupyter/jupyter_client/commit/0057186519ed356b822b1b1a0052e4068fccff12, released in version 7.3.2.
- They break Qtconsole and hence Spyder, but I checked locally that a simple revert fixes the problem.
- I added the qtconsole test suite here to avoid future breakages.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -79,3 +79,40 @@ jobs:
           conda install -c conda-forge xeus-cling
           pip install -e ".[test]"
           python -m unittest -v
+
+  qtconsole:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.9"
+          architecture: "x64"
+
+      - name: Install System Packages
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


blink1073:
Hi @ccordoba12, I think this might be an error in the typings in [`pyzmq`](https://github.com/zeromq/pyzmq/blob/29aded1bf017385866dcbf7b92a954f272360060/zmq/eventloop/zmqstream.py#L150).    I added a type ignore for now.

blink1073 (approved):
Thanks very much for adding the qtconsole test!

blink1073:
Docs failure is unrelated, I'll pick that up in a separate PR.

ccordoba12:
> I think this might be an error in the typings in [pyzmq](https://github.com/zeromq/pyzmq/blob/29aded1bf017385866dcbf7b92a954f272360060/zmq/eventloop/zmqstream.py#L150). I added a type ignore for now.

Great! Thanks for your help with that.

> Thanks very much for adding the qtconsole test!

No prob. Jupyter-client is critical for us, so that's something we really needed to do (I'll also do the same in IPykernel once I have some time).

davidbrochart:
Thanks @blink1073 and @ccordoba12 for taking care of that.
Something is still not clear to me, is the callback passed to `ZMQStream.on_recv` an awaitable or not in `pyzmq` v23.0?
I raised that point [in the original PR](https://github.com/jupyter/jupyter_client/pull/791#issuecomment-1138344313). Would love some feedback from @minrk.

PR #819, opened 2022-07-28
add `AsyncKernelClient` to `doc/api/client.rst`

helioz11:


davidbrochart (commented):
Thanks @helioz11.
I'm wondering if it's worth adding the API of `AsyncKernelClient`, since it's the same as `BlockingKernelClient` except that it's async.
Maybe we should have the complete API for `BlockingKernelClient` and just say that `AsyncKernelClient` is identical but async?

helioz11:
Sure, how about if I move the `automethod` from `AsyncKernelClient` to `BlockingKernelClient` and add extra information in `AsyncKernelClient`?

davidbrochart:
Yes, that sounds good.

docs/api/client.rst
```diff
@@ -49,3 +47,7 @@ client - communicating with kernels
    .. automethod:: wait_for_ready
 
    .. automethod:: is_alive
+
+.. autoclass:: AsyncKernelClient
+
+   The available class methods are identical with :class:`BlockingKernelClient` but async.
```
davidbrochart, line 53:
```suggestion
   :class:`AsyncKernelClient` is identical to :class:`BlockingKernelClient` but the methods described above are async.
```

davidbrochart (changes requested):

davidbrochart (approved):
Thanks @helioz11.

PR #826, opened 2022-08-26
Use pytest_asyncio fixture

davidbrochart:
Test `test_start_new_async_kernel` failed in https://github.com/conda-forge/jupyter_client-feedstock/pull/81 because pytest's asyncio mode is `strict`. I don't know why it doesn't fail here.
We can either mark the `start_async_kernel` fixture as a `pytest_asyncio` one, or globally set the mode to `auto`. This PR does the former.


blink1073 (approved):
Thanks!

PR #638, opened 2021-04-08
Fix qtconsole issues

davidbrochart:
@ccordoba12 this should fix the issues you were having with v6.1.13 in spyder. I checked that qtconsole works fine, but it might be worth trying spyder too. Actually I should open PRs in upstream projects with this jupyter_client.

PR #831, opened 2022-09-09
Avoid running nested runloops in ThreadedZMQSocketChannel

marc-etienne:
Hi!

I'm the maintainer of a project called [IPyIDA](https://github.com/eset/ipyida), which integrates the QtConsole inside [IDA Pro](https://hex-rays.com).

I'm using lots of plumbing to make this all work, and use [QtKernelManager and QtKernelClient](https://github.com/eset/ipyida/blob/508f542686da0b5379f269fa779c6d3fc45451eb/ipyida/ida_qtconsole.py#L140-L144) (which inherits from ThreadKernelClient) to connect to the kernel.

Since jupyter_client version 7, IPyIDA doesn't work and raises a RuntimeError and starting the client (see eset/ipyida#44). My temporary solution was to [require working version of jupyter_client](https://github.com/eset/ipyida/blob/508f542686da0b5379f269fa779c6d3fc45451eb/setup.py#L36) until I can find the source of the problem.

After much hours (probably too much :D) trying to understand the problem I think I found a proper solution. My understanding is that the `SocketStream.on_recv` was designed to be used with (pre-v6) Tornado runloop. Since Tornado 6, the Tornado runloop is just a wrapper around asyncio's. `on_recv` still works with the asyncio runloop but receives the (completed) Future object with the result. Trying to await the Future in another runloop crashes because the callback is called from an already running runloop.

The suggested commit fixes the problem with minimal changes because I didn't want to break other things I may not have full understanding about.

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -1,7 +1,6 @@
 """ Defines a KernelClient that provides thread-safe sockets with async callbacks on message
 replies.
 """
-import asyncio
 import atexit
 import errno
 import time
@@ -9,6 +8,7 @@ from threading import Event
 from threading import Thread
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Optional
 
 import zmq
@@ -63,7 +63,7 @@ class ThreadedZMQSocketChannel(object):
         def setup_stream():
             assert self.socket is not None
             self.stream = zmqstream.ZMQStream(self.socket, self.ioloop)
-            self.stream.on_recv(self._handle_recv)  # type:ignore[arg-type]
+            self.stream.on_recv(self._handle_recv)
             evt.set()
 
         assert self.ioloop is not None
@@ -107,13 +107,11 @@ class ThreadedZMQSocketChannel(object):
         assert self.ioloop is not None
         self.ioloop.add_callback(thread_send)
 
-    def _handle_recv(self, future_msg: asyncio.Future) -> None:
+    def _handle_recv(self, msg_list: List[bytes]) -> None:
         """Callback for stream.on_recv.
 
         Unpacks message, and calls handlers with it.
         """
-        assert future_msg.done()
-        msg_list = future_msg.result()
         assert self.session is not None
         ident, smsg = self.session.feed_identities(msg_list)
         msg = self.session.deserialize(smsg)
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -1,17 +1,15 @@
 """ Defines a KernelClient that provides thread-safe sockets with async callbacks on message
 replies.
 """
+import asyncio
 import atexit
 import errno
 import time
 from threading import Event
 from threading import Thread
 from typing import Any
-from typing import Awaitable
 from typing import Dict
-from typing import List
 from typing import Optional
-from typing import Union
 
 import zmq
 from traitlets import Instance
@@ -109,7 +107,7 @@ class ThreadedZMQSocketChannel(object):
         assert self.ioloop is not None
         self.ioloop.add_callback(thread_send)
 
-    def _handle_recv(self, future_msg: Awaitable) -> None:
+    def _handle_recv(self, future_msg: asyncio.Future) -> None:
         """Callback for stream.on_recv.
 
         Unpacks message, and calls handlers with it.
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -1,7 +1,6 @@
 """ Defines a KernelClient that provides thread-safe sockets with async callbacks on message
 replies.
 """
-import asyncio
 import atexit
 import errno
 import time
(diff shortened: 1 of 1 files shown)


marc-etienne:
ccing @ccordoba12 because he seems to be a user for Spyder and may want to test the change.

This issue is related to #803 and #638.

blink1073:
Hi @marc-etienne, can you please enable maintainer edits so I can push a fix for the linter?

ccordoba12:
Hey @marc-etienne, thanks for the ping! Qtconsole tests run here, so that should be enough, but given that this kind of changes can be very disruptive, I'll open a PR in Spyder and set it up to run against your PR to be extra sure.

About the changes themselves, I think @davidbrochart is the best to review them.

blink1073:
Thanks @marc-etienne and @ccordoba12!

jupyter_client/threaded.py
```diff
@@ -114,13 +107,13 @@ def thread_send():
         assert self.ioloop is not None
         self.ioloop.add_callback(thread_send)
 
-    def _handle_recv(self, future_msg: Awaitable) -> None:
+    def _handle_recv(self, future_msg: asyncio.Future) -> None:
```
davidbrochart, line 110: Looking at [pyzmq's documentation](https://pyzmq.readthedocs.io/en/latest/api/zmq.eventloop.zmqstream.html#zmq.eventloop.zmqstream.ZMQStream.on_recv), shouldn't the argument already be the (multipart) message, not a `Future`?
```suggestion
    def _handle_recv(self, msg_list: List[bytes]) -> None:
```

jupyter_client/threaded.py
```diff
@@ -114,13 +107,13 @@ def thread_send():
...
         """Callback for stream.on_recv.
 
         Unpacks message, and calls handlers with it.
         """
-        assert self.ioloop is not None
-        msg_list = self.ioloop._asyncio_event_loop.run_until_complete(get_msg(future_msg))
+        assert future_msg.done()
+        msg_list = future_msg.result()
```
davidbrochart, line 116:
```suggestion
```

davidbrochart (commented):
Thanks @marc-etienne, I think you're right about not awaiting the message in another event loop.
I would actually go further, because I think the `on_recv` callback is called with the message itself, not a `Future` for it.
I'll push a commit with my suggestions to see it the Qtconsole tests pass, first.

davidbrochart:
Qtconsole's tests [fail](https://github.com/jupyter/jupyter_client/runs/8293961249?check_suite_focus=true), suggesting that the argument to the `on_recv` callback is indeed an `asyncio.Future`, so your changes seem right @marc-etienne.
Pinging @minrk to confirm.

marc-etienne:
I think the pyzmq doc on `on_recv` is outdated. Or this could be a bug in pyzmq itself, maybe it should be pyzmq's responsibility to call `.result()`... I think the callback arg differs depending on the pyzmq context type of the stream (I think Jupyter Client uses [the asyncio one](https://github.com/jupyter/jupyter_client/blob/4bb7ac52ecb726c4ae964e727209eb0157462271/jupyter_client/client.py#L94-L101)?). A more general solution which doesn't depend on the context type could be to test the type of the received arg of `_handle_recv` using `asyncio.isfuture(msg)` and changing the method signature to something like `def _handle_recv(self, msg: Union[List[bytes], asyncio.Future]) -> None:`.

davidbrochart:
Let's clarify all that, and not try to guess the type of the argument at run-time.

minrk:
The short answer is: ThreadedKernelClient shouldn't use `zmq.asyncio.Context`, it should use `zmq.Context`.

`on_recv` should indeed only get the message frames, but technically what it does is get the return value for `socket.recv_multipart()`.

The problem is the switch here to `zmq.asyncio.Context`, which wasn't accounted for in ThreadedZMQChannel (or ThreadedKernelClient). So an async socket is being passed to a non-async API (ZMQStream), which means the async socket's result is getting passed to on_recv instead of the actual recv.

If there's a bug in pyzmq, it's that is creation of `ZMQStream(zmq.asyncio.Socket)` doesn't fail with a TypeError.

A class that uses ThreadedZMQSocketChannel should _not_ be creating asyncio Sockets. You can always create multiple Python Socket objects for a single zmq socket with e.g. `zmq.Socket.shadow(asyncio_socket)` (or the other way around: `zmq.asyncio.Socket.shadow(other_socket)`).

davidbrochart:
Thanks for the explanation @minrk.
If I understand correctly, this PR should include https://github.com/jupyter/jupyter_client/pull/831/commits/bb2319c57a6d0749a8ed7070a0334ed2ac421289, and the changes you mentioned should be done in downstream projects (Qtconsole, Spyder, etc.). Is that right?

blink1073:
As part of #835, we properly use a synchronous socket for `ThreadedZMQSocketChannel`.

PR #832, opened 2022-09-12
Add Jupyter Events and emit events from the Kernel Manager

Zsailer:
~This requires a release of Jupyter Events...~ done :+1:

- [x] Adds an event schema for the KernelManager
- [x] Adds events to various methods in the default KernelManager implementation.
- [x] Adds tests to the kernelmanager's lifecycle unit test to ensure events are emmitted.


ccordoba12:
This looks like a major change, so hopefully for Jupyter-client 8, right?

Zsailer:
@ccordoba12, yeah, I think that's fair. This adds an additional dependency, [jupyter_events](https://github.com/jupyter/jupyter_events), which I think warrants a major release.

blink1073:
Good call @ccordoba12, I've opened #833

jupyter_client/event_schemas/kernel_manager/v1.yaml
```diff
@@ -0,0 +1,36 @@
...
+  Record actions on kernels by the KernelManager.
+type: object
+required:
+  - kernel_id
+  - action
+properties:
+  kernel_id:
+    oneOf:
+      - type: string
+      - type: "null"
```
kevin-bates, line 14: This looks good @Zsailer - just have one question.

Is this in _anticipation_ of supporting other actions, because all of the currently defined actions fire with a non-None `kernel_id`?  If we retained that approach, we could then apply a format or pattern, allowing only UUID formats (and remove the `oneOf` and `null`).  This might allow event consumers to have an easier experience as well.

Zsailer: > because all of the currently defined actions fire with a non-None kernel_id?

I don't think `start_kernel` or `pre_start_kernel` methods strictly require a kernel_id to start, right? [This logic](https://github.com/jupyter/jupyter_client/blob/ff85055a3b0ab198e3b1250185628b5c4929615e/jupyter_client/manager.py#L342) seems to suggest that if `None` is provided, it creates a uuid on the fly.

tests/test_manager.py
```diff
@@ -32,3 +32,13 @@ def test_connection_file_real_path():
         km._launch_args = {}
         cmds = km.format_kernel_cmd()
         assert cmds[4] == "foobar"
+
+
+def test_kernel_manager_event_logger(jp_event_handler, jp_read_emitted_events):
+    action = "start"
```
blink1073, line 38:
```suggestion
    action = "pre_start"
```

.github/workflows/main.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Run the tests on pypy and windows
         if: ${{ startsWith(matrix.python-version, 'pypy') || startsWith(matrix.os, 'windows') }}
         run: |
-          python -m pytest -vv || python -m pytest -vv --lf
+          python -m pytest --pyargs jupyter_client -vv || python -m pytest -vv --lf
```
blink1073, line 79:
```suggestion
          python -m pytest -vv || python -m pytest -vv --lf
```

blink1073 (approved):
LGTM, thanks!

jupyter_client/event_schemas/kernel_manager/v1.yaml
kevin-bates: That is correct (on both counts; non-required kernel_id and auto-create logic).  The provisioners should have access to the kernel-id from the get-go so they can do indexing, etc. as necessary.

Issue #755, opened 2022-03-14
Clean up synchronous managers


blink1073:
Continued discussion from #751.

We also can't rely `asyncio_run` in the future, since it relies on the deprecated global `get_event_loop` and `set_event_loop`.

Options for moving forward:

- Use a single private ioloop instance per manager
  - The base classes will be asynchronous and call the the private async versions of functions directly as needed. We will only use futures and tasks from within async methods.
  - The synchronous classes will work by wrapping the asynchronous public methods using a decorator that runs a new private asyncio loop as return asyncio.new_event_loop().run_until_complete(async_method).
  - We would offer a function that wraps the async class to become synchronous, and use this ourselves
  - We would have a section like:

```
 # These methods are meant to be overridden by subclasses
async def _start_kernel(...)
```

- Run 'asyncio-for-sync' in a dedicated IO thread. `call_soon_threadsafe` and `concurrent.futures.Future` make this pretty doable, and in my experience simpler and more robust than using nest_asyncio. ipyparallel's sync client has worked that way for a long time.  (per @minrk)

davidbrochart:
>   * The synchronous classes will work by wrapping the asynchronous public methods using a decorator that runs a new private asyncio loop as return asyncio.new_event_loop().run_until_complete(async_method).

We may already be running inside an event loop, so this might not be possible, e.g.:
```python
import asyncio

loop2 = asyncio.new_event_loop()

async def main2():
    await asyncio.sleep(1)
    print("main2 done")

async def main1():
    loop2.run_until_complete(main2())
    print("main1 done")

asyncio.run(main1())
```
```python-traceback
Traceback (most recent call last):
  File "/home/david/github/davidbrochart/jupyterlab/foo.py", line 13, in <module>
    asyncio.run(main1())
  File "/home/david/mambaforge/envs/jupyterlab-dev/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/david/mambaforge/envs/jupyterlab-dev/lib/python3.10/asyncio/base_events.py", line 641, in run_until_complete
    return future.result()
  File "/home/david/github/davidbrochart/jupyterlab/foo.py", line 10, in main1
    loop2.run_until_complete(main2())
  File "/home/david/mambaforge/envs/jupyterlab-dev/lib/python3.10/asyncio/base_events.py", line 617, in run_until_complete
    self._check_running()
[...]

PR #835, opened 2022-09-14
Remove nest-asyncio dependency

blink1073:
- Remove the need for `nest_asyncio` by only calling asynchronous methods on self from within an asynchronous method and using the task runner method described in https://github.com/jupyter/jupyter_client/issues/755

- The blocking channel/client/manager methods call `run_sync` to run ad-hoc and background loop(s) as needed, and use `zmq.Context`.  The async channel/client/manager use `async/await` and `zmq.asyncio.Context`.   

- For session objects, we make all of the methods synchronous and check for futures returned by socket methods, converting them to values as needed.  We had to subclass `ZMQStream` to do the same, but perhaps this logic should go in `pyzmq` itself.

- Add tests for async and threaded clients

### DIFF
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -60,7 +60,17 @@ class _ZMQStream(ZMQStream):
 def as_zmqstream(f):
     def wrapped(self, *args, **kwargs):
         socket = f(self, *args, **kwargs)
-        return _ZMQStream(socket, self.loop)
+        save_socket_type = None
+        # zmqstreams only support sync sockets
+        if self.context._socket_type is not zmq.Socket:
+            save_socket_type = self.context._socket_type
+            self.context._socket_type = zmq.Socket
+        try:
+            return ZMQStream(socket, self.loop)
+        finally:
+            if save_socket_type:
+                # restore default socket type
+                self.context._socket_type = save_socket_type
 
     return wrapped
 
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -1,60 +1,16 @@
 """A kernel manager with a tornado IOLoop"""
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-import asyncio
-
 import zmq
 from tornado import ioloop
 from traitlets import Instance
 from traitlets import Type
-from zmq.eventloop.zmqstream import gen_log
 from zmq.eventloop.zmqstream import ZMQStream
 
 from .restarter import AsyncIOLoopKernelRestarter
 from .restarter import IOLoopKernelRestarter
 from jupyter_client.manager import AsyncKernelManager
 from jupyter_client.manager import KernelManager
-
-
-class _ZMQStream(ZMQStream):
-    def _handle_recv(self):
-        """Handle a recv event."""
-        if self._flushed:
-            return
-        try:
-            msg = self.socket.recv_multipart(zmq.NOBLOCK, copy=self._recv_copy)
-            if isinstance(msg, asyncio.Future):
-                msg = msg.result()
-        except zmq.ZMQError as e:
-            if e.errno == zmq.EAGAIN:
-                # state changed since poll event
-                pass
-            else:
-                raise
-        else:
-            if self._recv_callback:
-                callback = self._recv_callback
-                self._run_callback(callback, msg)
-
-    def _handle_send(self):
-        """Handle a send event."""
-        if self._flushed:
-            return
-        if not self.sending():
-            gen_log.error("Shouldn't have handled a send event")
-            return
-
-        msg, kwargs = self._send_queue.get()
-        try:
-            status = self.socket.send_multipart(msg, **kwargs)
-            if isinstance(status, asyncio.Future):
-                status = status.result()
-        except zmq.ZMQError as e:
-            gen_log.error("SEND Error: %s", e)
-            status = e
-        if self._send_callback:
-            callback = self._send_callback
-            self._run_callback(callback, msg, status)
 
 
 def as_zmqstream(f):
--- a/jupyter_client/ioloop/manager.py
+++ b/jupyter_client/ioloop/manager.py
@@ -7,6 +7,7 @@ import zmq
 from tornado import ioloop
 from traitlets import Instance
 from traitlets import Type
+from zmq.eventloop.zmqstream import gen_log
 from zmq.eventloop.zmqstream import ZMQStream
 
 from .restarter import AsyncIOLoopKernelRestarter
@@ -15,13 +16,15 @@ from jupyter_client.manager import AsyncKernelManager
 from jupyter_client.manager import KernelManager
 
 
-class AsyncZMQStream(ZMQStream):
+class _ZMQStream(ZMQStream):
     def _handle_recv(self):
         """Handle a recv event."""
         if self._flushed:
             return
         try:
             msg = self.socket.recv_multipart(zmq.NOBLOCK, copy=self._recv_copy)
+            if isinstance(msg, asyncio.Future):
+                msg = msg.result()
         except zmq.ZMQError as e:
             if e.errno == zmq.EAGAIN:
                 # state changed since poll event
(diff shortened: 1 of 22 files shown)


davidbrochart:
Looks like you're having fun in there :smile:

blink1073:
> Looks like you're having fun in there 😄

Ha, yes and no.

blink1073:
Current status:
- [ ] There is one hanging test on Ubuntu
- [ ] Jupyter server requires a slight tweak to asyncio handling: https://github.com/jupyter-server/jupyter_server/pull/988
- [ ] IPykernel will require changes to the inprocess kernel
- [ ] Nbclient has one failing test - will have to debug in nbclient
- [ ] Nbconvert is hanging on the execution test - will have to debug in nbconvert

minrk:
> We had to subclass ZMQStream to do the same

You shouldn't need to subclass, since you can make a  zmqstream by creating a shadow socket (this should be simpler):

```
async_socket = ...
blocking_socket = zmq.Socket.shadow(asocket.underlying)
stream = ZMQStream(blocking_socket)
```


> but perhaps this logic should go in pyzmq itself.

It definitely shouldn't do what it does now. I'm not 100% sure what it should do, though. If I add support for this, what I'll probably do is the above - shadow async sockets to sync before hooking up events. That's the only thing that makes sense to me. Maybe with a warning about unexpected socket types. That's technically a backward-incompatible change since you could use ZMQStream with async sockets, even though that doesn't make much sense, and it would change what `on_recv` is called with (from the inadvertent `Future` to the correct value). I'm pretty sure `on_recv` doesn't accept coroutines right now, either, and that should be easy to fix.

When writing the asyncio APIs, I didn't really expect applications to use both - I assumed it would be _either_ asyncio Sockets _or_ ZMQStream. I'll have to think more about what makes the most sense in a world with both in terms of support, deprecations, or both.

jupyter_client/ioloop/manager.py
```diff
@@ -12,10 +16,51 @@
...
+        if self._send_callback:
+            callback = self._send_callback
+            self._run_callback(callback, msg, status)
+
+
 def as_zmqstream(f):
     def wrapped(self, *args, **kwargs):
         socket = f(self, *args, **kwargs)
-        return ZMQStream(socket, self.loop)
+        return _ZMQStream(socket, self.loop)
```
minrk, line 63: I've meant to make this easier/more public for ages, but this is the official way to instantiate a non-default socket type for a context, rather than needing a custom `_ZMQStream`:

```suggestion
        save_socket_type = None
        # zmqstreams only support sync sockets
        if self.context._socket_type is not zmq.Socket:
            save_socket_type = self.context._socket_type
            self.context._socket_type = zmq.Socket
        stream = ZMQStream(socket, self.loop)
        if save_socket_type:
            # restore default socket type
            self.context._socket_type = save_socket_type
        return stream
```

minrk: (`with mock.patch.object` doesn't work because of a bug in `Context.__delattr__`)

jupyter_client/session.py
```diff
@@ -849,10 +850,14 @@ def send(
         if stream and buffers and track and not copy:
             # only really track when we are doing zero-copy buffers
             tracker = stream.send_multipart(to_send, copy=False, track=True)
+            if isinstance(tracker, asyncio.Future):
+                tracker = tracker.result()
```
minrk, line 854: This isn't going to work if the future's not done, is it? It should raise `InvalidStateError: Result is not set.`

blink1073: Yeah, I think I got lucky so far in that it was always done when I checked, looks like I'll have to use the shadow trick for `Session`.

blink1073:
I think we introduced a failure:

```
Error: TestKernelManager.test_ipc_cinfo

AttributeError: Context has no such option: _SOCKET_TYPE
```

jupyter_client/ioloop/manager.py
```diff
@@ -15,7 +16,17 @@
...
+        # zmqstreams only support sync sockets
+        if self.context._socket_type is not zmq.Socket:
+            save_socket_type = self.context._socket_type
+            self.context._socket_type = zmq.Socket
+        try:
+            return ZMQStream(socket, self.loop)
+        finally:
+            if save_socket_type:
+                # restore default socket type
+                self.context._socket_type = save_socket_type
```
minrk, line 29: Sorry! This is what I get for having too much mypy on the brain. It's `socket_class` not `socket_type`

```suggestion
        save_socket_class = None
        # zmqstreams only support sync sockets
        if self.context._socket_class is not zmq.Socket:
            save_socket_class = self.context._socket_class
            self.context._socket_class = zmq.Socket
        try:
            return ZMQStream(socket, self.loop)
        finally:
            if save_socket_class:
                # restore default socket class
                self.context._socket_class = save_socket_class
```

jupyter_client/ioloop/manager.py
```diff
@@ -15,6 +16,19 @@
...
         socket = f(self, *args, **kwargs)
+        save_socket_class = None
+        # zmqstreams only support sync sockets
+        if self.context._socket_class is not zmq.Socket:
+            save_socket_class = self.context._socket_class
+            self.context._socket_class = zmq.Socket
+        try:
+            orig_socket = socket
+            socket = f(self, *args, **kwargs)
+            orig_socket.close()
```
minrk, line 27: Why is `orig_socket` created and closed, instead of removed entirely?

jupyter_client/session.py
```diff
@@ -807,6 +807,10 @@ def send(
             # ZMQStreams and dummy sockets do not support tracking.
             track = False
 
+        if isinstance(stream, zmq.asyncio.Socket):
+            assert stream is not None
+            stream = zmq.Socket.shadow(stream.underlying)
```
minrk, line 812: 👍

jupyter_client/session.py
```diff
@@ -807,6 +807,10 @@ def send(
             # ZMQStreams and dummy sockets do not support tracking.
             track = False
 
+        if isinstance(stream, zmq.asyncio.Socket):
+            assert stream is not None
```
minrk, line 811: I assume leftover debug, this will never be None if it's an asyncio.Socket

```suggestion
```

blink1073: This is there to satisfy `mypy`

jupyter_client/ioloop/manager.py
blink1073: Didn't we need to create it to determine it's type?

jupyter_client/session.py
minrk: type checkers, always making code better!

jupyter_client/ioloop/manager.py
minrk: No, we only check `self.context._socket_class` which we can do before creating sockets

blink1073: Okay, I removed the throwaway socket

minrk:
https://github.com/zeromq/pyzmq/pull/1785 introduces support in ZMQStream for both asyncio sockets (by immediately shadowing them to sync sockets) and allowing coroutines in on_recv callbacks.

It also makes it easier to explicitly create a particular socket class per `ctx.socket` call with e.g. `ctx.socket(zmq.PUSH, socket_class=zmq.asyncio.Socket)`

tests/test_restarter.py
```diff
@@ -191,7 +192,7 @@ async def test_async_restart_check(config, install_kernel, debug_logging):
     km = AsyncIOLoopKernelManager(kernel_name=install_kernel, config=config)
 
     cbs = 0
-    restarts = [asyncio.Future() for i in range(N_restarts)]
+    restarts = [asyncio.futures.Future() for i in range(N_restarts)]
```
minrk, line 195: This should still be `asyncio.Future`. That's the correct, public name for this. `asyncio.futures` is an implementation detail, which can change.

blink1073: We needed to use concurrent futures in the non-async tests, and asyncio futures in the async ones, I'll update to use `asyncio.Future`

blink1073:
Downstream errors are addressed, I'm going to merge and iterate.

PR #852, opened 2022-10-11
Fix ready promise and session send

blink1073:
- Preserves previous use of `ready` promises, but preserves the promise created in the constructor for initial use.
- Backports a fix from #835 to handle asyncio sockets in `Session.send`


Zsailer (approved):

Zsailer:
Looks good, thanks!

PR #860, opened 2022-10-21
Fix assertion in `TestSession.test_serialize`

samrat:
There was a typo due to which the type of a value was getting compared to itself, so it would always pass.


blink1073 (approved):
Thank you!

PR #861, opened 2022-10-22
Workaround for launch bug

blink1073:
cc @kevin-bates 

This prevents a race condition where the kernel launch process is attempting to read the connection file, and we overwrite the file while it is being read.   The symptom is that the kernel fails to start because it cannot decode the JSON in the connection file.

We should backport this fix to 7.x, but I think for 8.x we should change it so that the provisioner is entirely responsible for the handling of the connection file, even if the information lives on the kernel manager.   Nothing would change for the local provisioner.

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -158,6 +158,18 @@ def write_connection_file(
     cfg["signature_scheme"] = signature_scheme
     cfg["kernel_name"] = kernel_name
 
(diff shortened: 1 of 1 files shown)


kevin-bates (approved):
LGTM.
I completely agree with your comment regarding the  provisioner's role relative to the connection file.

blink1073:

PR #879, opened 2022-11-15
Reconcile connection information

kevin-bates:
Per the [side discussion between myself and @blink1073](https://github.com/jupyter/jupyter_client/pull/863#issuecomment-1311913872) on PR #863 that was reworking #861, this PR checks that the existing connection file (should one exist) matches the connection information returned from the provisioner and prevents overwriting the connection file in such cases (to avoid the race condition addressed via #861).

These changes also ensure that once the connection file is written (if it didn't already exist) the connection information stored in attributes of `KernelManager` (via the `ConnectionFileMixin`) also matches that returned from the provisioner (which is considered the _truth_).

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -157,21 +157,9 @@ def write_connection_file(
     cfg["signature_scheme"] = signature_scheme
     cfg["kernel_name"] = kernel_name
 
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thank you!

blink1073:

PR #882, opened 2022-11-16
Fix connection reconciliation to handle restarts

kevin-bates:
The changes delivered in #879 did not account for scenarios when ports have changed during restarts - which tends to occur with remote kernels.  As a result, we need to reset the connection file when the connection information returned from the provisioner differs from the file's contents, rather than raise a `ValueError`.  

Although this is probably not an issue in typical situations, it is something that should be fixed.  I apologize to @blink1073 for having not caught this in my initial changes in #879 since two different releases were cut following those changes.


blink1073 (approved):
Thanks!  I'll cut these releases in the morning.

blink1073:

PR #889, opened 2022-11-24
Import ensure_async and run_sync from jupyter_core

davidbrochart:
Needs https://github.com/jupyter/jupyter_core/pull/315.


davidbrochart:
You're too fast @blink1073, each time I want to push a commit I see you already did it one minute ago.

PR #893, opened 2022-12-03
Make sure loop is marked with nest_asyncio

impact27:
I know that this is already obsolete with https://github.com/jupyter/jupyter_client/pull/835.  Sometimes, this loop is not marked by nest_asyncio (when calling `run_sync`) so this makes sure that the two eventloops (`asyncio.new_event_loop()` and the other created in `ioloop.IOLoop()`) do not make the system crash. Unfortunately, I can not really give you a reproducible example as I was messing with spyder code when it happened. Feel free to close this PR if you think this is not needed.

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -3,7 +3,6 @@ replies.
 """
 import asyncio
 import atexit
-import nest_asyncio
 import errno
 import time
 from threading import Event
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!  I'll cut a release with this on Monday.

PR #905, opened 2022-12-23
MAINT: [_async_]start_kernel should only take kwarg only.

Carreau:
The base cases only take **kw, so you can't have subclass also take positional arguments.

It is problematic to allow positional arguments, as is someone code against a subclass it is not swappable for another one.

The latest version of jupyter_server also seem to sometime have subclasses that have the signature with kernel_name first and other with kernel_id first, with is a recipe for disaster if we don't make it kwarg only.

This is I belove not caught by mypy because of multiple reasons:

1) the use of run_sync in a couple of place,
2) method assignement might not be handled by mypy.


blink1073 (approved):
Thank you!

PR #906, opened 2022-12-23
MAINT: Proper typing and cast

Carreau:
Cast has minimal if no runtime cost, and is slightly more precise that just ignoring types


blink1073 (approved):
Thanks!

Carreau:
thanks

PR #842, opened 2022-09-29
Remove some deprecated zmq imports

minrk:
The zmq.eventloop.ioloop shim has been deprecated since pyzmq 17

jupyter-client requires pyzmq 23

PR #910, opened 2023-01-03
Require pyzmq v25

davidbrochart:


minrk:
Is there a reason to require 25? This PR was opened before 25 was even out, and I think it's good to be as conservative as possible with support for dependencies like this one.

davidbrochart:
> Is there a reason to require 25?

No, I just wanted to see what was going to break.

blink1073:
Closing as addressed by #915  and #842

PR #912, opened 2023-01-05
MAINT: consistently use relative imports.

Carreau:
It seem like jupyter_client uses a mix of absolute and relative imports, even in same files, which seem a bit weird.

Normalise everything to relative imports, which should make it a bit easier to detect potential cycles, or higher-level imports.

I've found 6 places that now use `..` to imports somethings, where would potentially want to make some refactor at some point in the future to not import something closer to the root.


blink1073 (approved):
Thanks!

Issue #482, opened 2019-09-26
kernel client blocks indefinitely during gc.collect() call


markwaddle:
after a `BlockingClient` execution request that times out, `gc.collect()` hangs indefinitely

see the repro at https://gist.github.com/markwaddle/af87e1d314fad4021d8d25c0b7c59cb7

```python
import gc
from jupyter_client import KernelManager


def start_kernel():
    kernel_manager = KernelManager()
    kernel_manager.start_kernel()
    kernel_client = kernel_manager.blocking_client()
    kernel_client.execute_interactive(code='', silent=False, allow_stdin=False,
            output_hook=lambda x: print('output:', x), timeout=.1)


if __name__ == '__main__':
    try:
        start_kernel()
    except Exception as e:
        print('exception:', e)
    print('garbage collecting - process hangs indefinitely here')
    gc.collect()
    print('done - never reaches here')
```

markwaddle:
interestingly, if you place the body of the `start_kernel` function directly under main, instead of calling the function, the hang does not occur.

markwaddle:
and this is where it hangs:
```
Exception ignored in: <function Context.__del__ at 0x1040d7d08>
Traceback (most recent call last):
  File "/Users/mark/.local/share/virtualenvs/kernel_test-kiveVYPU/lib/python3.7/site-packages/zmq/sugar/context.py", line 50, in __del__
    self.term()
  File "zmq/backend/cython/context.pyx", line 91, in zmq.backend.cython.context.Context.term
  File "zmq/backend/cython/checkrc.pxd", line 12, in zmq.backend.cython.checkrc._check_rc
KeyboardInterrupt
```

markwaddle:
on a side note, does anyone know why the `BlockingClient. execute_interactive` call times out? is there something else i should be doing before making that call?

markwaddle:
i've determined that a call to `kernel_client.stop_channels()`, prior to the gc.collect(), will allow the collect to complete

PR #914, opened 2023-01-12
Make Qtconsole work with PyZMQ 25

ccordoba12:
- It seems PyZMQ streams return now a list of bytes and not a future, so this was not handled properly.
- This doesn't seem necessary on the `main` branch.

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -71,7 +71,7 @@ class ThreadedZMQSocketChannel(object):
         def setup_stream():
             assert self.socket is not None
             self.stream = zmqstream.ZMQStream(self.socket, self.ioloop)
-            self.stream.on_recv(self._handle_recv) 
+            self.stream.on_recv(self._handle_recv)
             evt.set()
 
         assert self.ioloop is not None
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -71,7 +71,7 @@ class ThreadedZMQSocketChannel(object):
         def setup_stream():
             assert self.socket is not None
             self.stream = zmqstream.ZMQStream(self.socket, self.ioloop)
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'm also seeing this warning printed in the terminal when Qtconsole starts:

```
/home/carlos/Projects/jupyter_client/jupyter_client/threaded.py:73: RuntimeWarning: ZMQStream only supports the base zmq.Socket class.

                Use zmq.Socket(shadow=other_socket)
                or `ctx.socket(zmq.DEALER, socket_class=zmq.Socket)`
                to create a base zmq.Socket object,
                no matter what other kind of socket your Context creates.
                
  self.stream = zmqstream.ZMQStream(self.socket, self.ioloop)
```

but I don't know how to fix it.

ccordoba12:
Ok, I think this also requires #482 to make the tests pass. Pinging @blink1073 and @minrk about it.

blink1073 (approved):
Thanks!

ccordoba12:
Thanks @blink1073 for your help with this! It seems the ipykernel failure is caused by the new Qt eventloop integration, but I don't know how to fix it.

minrk:
For tracking, the change is https://github.com/zeromq/pyzmq/pull/1785 which 'fixes' ZMQStream compatibility with async sockets (with a warning, because passing an async socket to ZMQStream indicates an odd mixture of async/sync APIs).

ZMQStream never supported wrapping async sockets before pyzmq 25, but instead of raising it would fire callbacks with the wrong type (usually a Future instead of the message). Now, the `on_recv` callback will get the same arguments, no matter what socket class it's been called with.

The general fix that will work on all pyzmq versions is to ensure that ZMQStream is passed a regular zmq socket, which I believe was done in #835.

PR #915, opened 2023-01-12
Remove deprecated zmq imports

blink1073:
closes #842

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -19,7 +19,6 @@ jobs:
       - uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: ipykernel
-          package_spec: "pyqt5 ipykernel[test]"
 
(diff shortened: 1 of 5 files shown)


blink1073:
Hmm, it appears we might have other problems with pyzmq 25.

minrk:
I'll look into this now! I wouldn't expect issues with this very small release, but nobody ever _expects_ issues.

minrk:
After isolating pyzmq, it seems this is related to pyqt5 integration changes, and not related to pyzmq at all.

blink1073:
Okay, we do have to fix one test to not pass an asyncio socket to `ZMQStream`, but you're right it looks like the underlying issue is in `ipykernel`.

blink1073:
Aha!  The problem is that we were explicitly installing `pyqt5` in our downstream `ipykernel` tests (here and in 7.x), which now requires a display buffer for testing.

PR #919, opened 2023-01-24
MAINT: Don't format log in log call.

Carreau:
In particular f-string:

There are many reason not to use f-string in logs,
 - F-strings are eager, so might be slow.
 - prevent structured logging or handler to highlight.
 - Security (untrusted input can lead to DOS on formatting, https://discuss.python.org/t/safer-logging-methods-for-f-strings-and-new-style-formatting/13802)
 - Performance (we can filter before formatting).

But also % formatting in a couple of places.


blink1073 (approved):
Thanks!

Issue #930, opened 2023-02-12
Trailing newline in base64 encoded data


xl0:
I've noticed that Jupyter notebook will append a "\n" to any base64 encoded data in the output of the cell. This does not cause problems, except that VScode-jupyter strips the "\n".

This results in excessively large diffs when Jupyter notebook and vsccde-jupyter are used on the same codebase. I understand that Jupyter was there first, but to me not having the extra "\n" feels like a better solution than having it. Unless I'm missing a good reason for having it, would it be possible to not add this trailing "\n"?

In my understanding the "\n" is introduced by calling b2a_base64 from `session.py:json_default`:

```py
def json_default(obj):
    [...]
    if isinstance(obj, bytes):
        return b2a_base64(obj).decode('ascii')
    [...]
```

Since python 3.6 b2a_base64() accepts a `newline=True` argument.
The description of the function in 3.5 documentation is:
> binascii.b2a_base64(data)
> Convert binary data to a line of ASCII characters in base64 coding. The return value is the converted line, including a newline char.
> The newline is added because the original use case for this function was to feed it a series of 57 byte input lines to get output lines
> that conform to the MIME-base64 standard. Otherwise the output conforms to [RFC 3548](https://tools.ietf.org/html/rfc3548.html).

Would it be possible to not include that extra newline by passing `newline=False` to `b2a_base64()`?

To reptoduce:
```py
from PIL import Image
Image.new("RGB", (10, 10), "green")
```

results in
```
   "outputs": [
    {
     "data": {
      "image/png": "iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAIAAAACUFjqAAAAE0lEQVR4nGNkaGDAA5jwSY5caQCnUgCUBZU3vQAAAABJRU5ErkJggg==\n",
      "text/plain": [       "<PIL.Image.Image image mode=RGB size=10x10>"      ]     },
    ...
```

PR #931, opened 2023-02-12
Don't emit a trailng newline in base64-encoded data like 'image/png'

xl0:
This should address #930

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -111,7 +111,7 @@ def json_default(obj):
         return obj.isoformat().replace('+00:00', 'Z')
 
     if isinstance(obj, bytes):
-        return b2a_base64(obj).decode('ascii')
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

kevin-bates:
@xl0 thanks for the pull request and analysis, it all makes sense to me. 

My only concern, given the time that a training `\n` has been in use, is that some applications may be expecting it or will not tolerate its absence gracefully and i wonder if this is something that should be configurable, at least initially. 

I'm unable to check the history of that call but I would imagine it's been around since prior to python 3.6. 

This configurable could be something akin to the message debug option where, iirc, it isn't a configurable trait. That would at least give intolerant applications a way forward. 

(Sorry, I've access to only my phone at the moment but wanted to raise this topic.)

blink1073:
I would say anyone consuming these values would have most likely called `strip()` on them, which would still work.

PR #933, opened 2023-02-15
fix kernelspec print output

minrk:
typo in log -> print migration (#923) resulted in output like:

```
Available kernels:
  %s    %s python3                                                           /Users/minrk/conda/share/jupyter/kernels/python3
```

and keep errors written to stderr, not stdout


kevin-bates (approved):
Good catch.  These changes LGTM.  Thanks Min.

PR #936, opened 2023-03-16
ThreadedZMQStream: close stream before socket

minrk:
not just the underlying socket, because the stream represents event listeners on the socket, and can produce errors or at least warnings if the socket is closed before being unregistered. Or worst case, continue to register the wrong callbacks on a subsequent socket if it uses the freed FD before it's noticed.

related to https://github.com/spyder-ide/spyder/issues/20381

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -4,12 +4,14 @@ replies.
 import asyncio
 import atexit
 import time
+from concurrent.futures import Future
 from threading import Event, Thread
 from typing import Any, Dict, List, Optional
 
 import zmq
 from tornado.ioloop import IOLoop
 from traitlets import Instance, Type
+from traitlets.log import get_logger
 from zmq.eventloop import zmqstream
 
 from .channels import HBChannel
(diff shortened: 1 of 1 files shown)


jupyter_client/threaded.py
```diff
@@ -80,6 +82,29 @@ def stop(self) -> None:
...
+                    self.socket.close(linger=0)
+                    self.socket = None
+                except Exception as e:
+                    f.set_exception(e)
+                else:
+                    f.set_result(None)
+
+            self.ioloop.add_callback(close_stream)
+            try:
+                f.result(timeout=5)
```
minrk, line 102: This pattern should prevent hangs in case of errors where we use `Event.wait()` without timeouts elsewhere in this file.

jupyter_client/threaded.py
```diff
@@ -80,6 +82,29 @@ def stop(self) -> None:
 
     def close(self) -> None:
         """ "Close the channel."""
```
ccordoba12, line 84:
```suggestion
        """Close the channel."""
```

ccordoba12 (commented):
Small fix in a docstring, the rest looks good to me. Thanks @minrk!

blink1073:
Thanks!  I'll cut a release tomorrow.

PR #940, opened 2023-03-29
use c.f.Future to wait across threads

minrk:
Future allows us to propagate errors across threads, instead of threading.Event, which only waits forever and can result in uninformative hangs if there are errors.

this is related to https://github.com/spyder-ide/spyder/issues/20381 where flush is being called on a closed stream, but the traceback doesn't track to the actual caller of `channel.flush()` because of how it's scheduled.

This shouldn't actually change the behavior, only improve error reporting when e.g. calling flush on a closed channel.

Alternately, we could _choose_ to turn this error into a warning, which would in fact be a change in behavior.

### DIFF
--- a/jupyter_client/threaded.py
+++ b/jupyter_client/threaded.py
@@ -5,7 +5,8 @@ import asyncio
 import atexit
 import time
 from concurrent.futures import Future
-from threading import Event, Thread
+from functools import partial
+from threading import Thread
 from typing import Any, Dict, List, Optional
 
 import zmq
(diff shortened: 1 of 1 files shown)


ccordoba12:
> Alternately, we could choose to turn this error into a warning, which would in fact be a change in behavior.

I think it's better it's better to raise an error just to be sure that we're doing things as expected.

blink1073 (approved):
Thank you!

Issue #955, opened 2023-07-05
Create a kernel with specified custom ports


jkitchin:
I am trying to start a jupyter kernel like this:

```
jupyter-kernel \
       --ip=0.0.0.0 \
       --KernelManager.control_port=56406 \
       --KernelManager.hb_port=56407 \
       --KernelManager.iopub_port=56408 \
       --KernelManager.shell_port=56409 \
       --KernelManager.stdin_port=56410
```
It starts, and I see this output:

[KernelApp] Starting kernel 'python3'
[KernelApp] Connection file: /Users/jkitchin/Library/Jupyter/runtime/kernel-892382b5-8115-44ba-95cd-d8172ba848ac.json
however, the kernel json contains:

```
cat /Users/jkitchin/Library/Jupyter/runtime/kernel-892382b5-8115-44ba-95cd-d8172ba848ac.json
{
  "shell_port": 55505,
  "iopub_port": 55506,
  "stdin_port": 55507,
  "control_port": 55509,
  "hb_port": 55508,
  "ip": "0.0.0.0",
  "key": "0a163fda-2b4963c994abc4277bdf58fa",
  "transport": "tcp",
  "signature_scheme": "hmac-sha256",
  "kernel_name": ""
}
```

which do not match the ports I specified. Am I missing something to make this happen?

jkitchin:
I can confirm the issue is with KernelManager.cache_ports. At https://github.com/emacs-jupyter/jupyter/issues/481#issuecomment-1622725743 I patched this and then it works as one would expect it to. I don't know if it is a general solution for others who want to use this, but it probably beats forking it.

kevin-bates:
Hi @jkitchin - thanks for taking a closer look.  

I think we can make `cache_ports` configurable by adding the `config=True` keyword argument into its "declaration" on [L185](https://github.com/jupyter/jupyter_client/blob/e7ebada6044f4b269b120e561bdba0cef5587a68/jupyter_client/manager.py#L185) and similar to the [`shutdown_wait_time`](https://github.com/jupyter/jupyter_client/blob/e7ebada6044f4b269b120e561bdba0cef5587a68/jupyter_client/manager.py#L155C5-L155C23) declaration.  For completeness, we might as well add the default value of `False` argument prior to `config=True`. 

This would then enable the use of `--KernelManager.cache_ports=False` on your command line, allowing custom port definitions to occur.

[...]

Issue #878, opened 2022-11-13
Cannot configure ports if 'transport' is 'tcp'


yves-surrel:
Changing the shell_port, iopub_port etc. in jupyter_console_config.py does not work, neither does using `jupyter console --iopub:xxxx` in the command line.

After investigation, it comes from the `cache_ports` attribute of `KernelManager`:
```
    def _default_cache_ports(self) -> bool:
        return self.transport == "tcp"
```
in `manager.py' and
```
            if km.cache_ports and not self.ports_cached:
                lpc = LocalPortCache.instance()
                km.shell_port = lpc.find_available_port(km.ip)
                km.iopub_port = lpc.find_available_port(km.ip)
                km.stdin_port = lpc.find_available_port(km.ip)
                km.hb_port = lpc.find_available_port(km.ip)
                km.control_port = lpc.find_available_port(km.ip)
                self.ports_cached = True
```
in `async def pre_launch` in `local_provisioner.py`.

Maybe related to #492

Jupiter_client v 7.3.5

kevin-bates:
Hi @yves-surrel - it looks like you've identified a possible issue.  I'm not familiar with `jupyter console` but see the `consoleapp.py` file in the repo - although there is no corresponding "console-script" in the set of scripts in `pyproject.toml` (or has there been in the past), so I'm not sure where `jupyter console` is coming from.  Could you clarify how that application is installed/exposed?

I'm curious, assuming the `KernelManager` can be configured in this application (as looks to be the case in [`consoleapp.py`](https://github.com/jupyter/jupyter_client/blob/9f1c379b6f50012cad50037617d1ad3ac0ec47c3/jupyter_client/consoleapp.py#L97)), whether you can disable the port caching on the `KernelManager` prior to your invocation, and whether or not that moves you forward.   Perhaps try on the CLI: `--KernelManager.cache_ports = False` or in the file: `c.KernelManager.cache_ports = False`?  Thanks.

cc: @martinRenou for possible insights

yves-surrel:
Setting `--KernelManager.cache_ports = False` does not change anything, and it's the same behavior in `jupyter qtconsole`, which is no surprise as `cache_ports` is not `Config`-able, as stated in #492. BTW, have you an idea why this 'port caching' mechanism does exist for transport = 'tcp'?

kevin-bates:
[...]

PR #956, opened 2023-07-06
Make cache_ports configurable with default value of False.

jkitchin:
This addresses issue #878 and #955. Suggested by @kevin-bates.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -185,7 +185,7 @@ class KernelManager(ConnectionFileMixin):
     cache_ports: Bool = Bool(
         False,
         config=True,
-        help="True if the MultiKernelManager should cache ports for this KernelManager instance"
+        help="True if the MultiKernelManager should cache ports for this KernelManager instance",
     )
 
     @default("cache_ports")  # type:ignore[misc]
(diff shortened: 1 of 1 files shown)


jkitchin:
I guess I should have added a label. I don't know what this would be considered, I guess an enhancment?

kevin-bates:
>  I don't know what this would be considered, I guess an enhancement?

Given the previous discussion a while ago, I view this change as something that should have been part of the original change and have applied the `bug` label.

kevin-bates (approved):
LGTM - thank you @jkitchin

jkitchin:
Does anything else need to happen to this?

kevin-bates:
We usually give folks a day or two to review.  However, the build issue is also unwarranted but I believe it is probably unrelated.

@blink1073 - is the coverage issue due to some external version update or something like that?  (Sorry, I haven't been "close" to the repos lately.)

blink1073:
The coverage thing is a weird artifact from switching away from codecov where it seems like coverage isn't honoring `relative_files = true` in some cases.  I've seen this on `nbconvert` as well.

blink1073:
I'll cut a new release early next week.

kevin-bates:
Thank you Steve!

Issue #957, opened 2023-07-07
Report exit status if kernel unexpectedly dies


tmaxwell-anthropic:
If the Jupyter kernel process unexpectedly dies, Jupyter will show a message in the frontend. However, it doesn't report the exit status of the dead process. The exit status would be useful for diagnosing why the process crashed -- for example, I recently spent a while trying to figure out why a kernel process crashed, before eventually realizing it had been SIGKILL'ed by the Linux OOM killer.

I'm not sure what's the best way to implement this. `KernelManager.is_alive` doesn't provide a way to report the reason for death. `jupyter_server.ServerKernelManager` adds a `reason` field; it's not currently used for this purpose, but perhaps it could be? There would also need to be a way to report the message to the frontend, perhaps via a new field in the kernel status message? And then the frontend would need to be updated to render that field.

For now I'm working around this by subclassing KernelManager and ZMQChannelsWebsocketConnection to report the exit code to the frontend via the stderr: https://gist.github.com/tmaxwell-anthropic/de7a54753d312ccce831dbfd48ef2b12 But it would be nice if this could be a first-class feature.

jakkdl:
Shouldn't this be reported primarily in https://github.com/jupyter-server/jupyter_server? Only once the server has support for it could any eventual client-side support be handled.

tmaxwell-anthropic:
My understanding is that `jupyter_client` is the "client" of the kernel, not the "client" of the server. I.e. `jupyter_client` is the component of `jupyter_server` which is responsible for talking to the kernel process. For example, the logic that checks if a kernel is alive is defined here in `jupyter_client`: https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/manager.py#L647-L654

I figured this feature would first need support in `jupyter_client`, then `jupyter_server`, then the actual frontends (e.g. https://github.com/jupyter/nbclassic). Does that sound plausible to you?

jakkdl:
Ah, my bad, I missed the distinction between `ServerKernelManager` and `KernelManager`. And thanks for the clarification of the client/server distinction, think I'm following now.

The interfaces of `KernelManager` (in client) and `GatewayKernelManager` (in server) looks quite messy at the moment, with several fields being `getattr`d and/or `type: ignore[attr-defined]`. But `reason` seems to only be set at initialization, so promoting it to a proper property of `KernelManager` and also using it when the server is killed makes sense to me - and will at least partly clean up the interface mess.

[...]

PR #958, opened 2023-07-19
add KernelManager.exit_status

jakkdl:
Quick draft implementation of #957 in the way suggested by @kevin-bates

* I'm not sure what value to return `if not self.has_kernel`, both `0` and `None` seem plausible - and could also go with a positive value.
  * or the value can maybe be saved away so it can be accessed even after doing cleanup.

I tested all possible signals, and a bunch of them were ignored - including `signal.SIGINT`. Not sure if that's intended.

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -655,6 +655,17 @@ class KernelManager(ConnectionFileMixin):
 
     is_alive = run_sync(_async_is_alive)
 
+    async def _async_exit_status(self) -> int | None:
+        """Returns 0 if there's no kernel or it exited gracefully,
+        None if the kernel is running, or a negative value `-N` if the
+        kernel was killed by signal `N` (posix only)."""
+        if not self.has_kernel:
+            return 0
+        assert self.provisioner is not None
+        return await self.provisioner.poll()
+
+    exit_status = run_sync(_async_exit_status)
+
     async def _async_wait(self, pollinterval: float = 0.1) -> None:
         # Use busy loop at 100ms intervals, polling until the process is
         # not alive.  If we find the process is no longer alive, complete
--- a/tests/test_kernelmanager.py
+++ b/tests/test_kernelmanager.py
@@ -160,6 +160,46 @@ class TestKernelManagerShutDownGracefully:
         assert km._shutdown_status in expected
 
 
+class TestKernelManagerExitStatus:
+    @pytest.mark.skipif(sys.platform == "win32", reason="Windows doesn't support signals")
+    @pytest.mark.parametrize('_signal', [signal.SIGHUP, signal.SIGTERM, signal.SIGKILL])
+    async def test_exit_status(self, _signal):
+        # install kernel
+        _install_kernel(name="test_exit_status")
+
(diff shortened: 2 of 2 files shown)


jakkdl:
@tmaxwell-anthropic any opinions on the detailed behaviour, or in what ways it would be maximally useful for cases such as yours? See in-line comments.

Fixed CI issues*, and testing to see how signals that are defined on windows behave.


\* why is mypy not configured to be run in pre-commit? And `python_version=3.8` should probably be set in the config. I could open up a separate PR for that.

tmaxwell-anthropic:
Thanks for looking into this!

I don't know much about Jupyter internals, but here are some thoughts:

- What happens if the kernel dies, but then the auto-restarter restarts it? Ideally, the user should be notified that the kernel was restarted, and also why it was restarted (which signal killed it). It looks like the restarter delivers a callback every time the kernel dies or is restarted. Maybe we want to include the exit code in that callback?

- I don't know how the kernel restarter interacts with intentional shutdown. Ideally the user can tell the difference between whether Jupyter intentionally shut down the kernel, versus the kernel unexpectedly calling something like `os._exit(0)`.

kevin-bates:
The [auto-restarter is stopped](https://github.com/jupyter/jupyter_client/blob/508982f11a3175089b572a0c8277542d9fbd0729/jupyter_client/manager.py#L500) for any flavor of intentional "shutdown" (either directly or on behalf of intentional restarts).  It lives only to detect unexpected terminations.

jakkdl:
>     * What happens if the kernel dies, but then the auto-restarter restarts it? Ideally, the user should be notified that the kernel was restarted, and also why it was restarted (which signal killed it). It looks like the restarter delivers a callback every time the kernel dies or is restarted. Maybe we want to include the exit code in that callback?
> 

Adding a parameter with the exit code to the callback would either break all existing code, or require [inspecting the callable](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) to see if it can take a parameter with the signature or not. The latter is definitely doable, but not sure if it's the best way go to about it.

tmaxwell-anthropic:
Another option might be to add a second set of callbacks, with different names, which are equivalent except that they take an extra parameter. So the component that's registering the callback could indicate whether it wants a parameter based on which callback name it registers for.

jakkdl:
> Another option might be to add a second set of callbacks, with different names, which are equivalent except that they take an extra parameter. So the component that's registering the callback could indicate whether it wants a parameter based on which callback name it registers for.

Ah that's pretty nice. Or probably even cleaner, add a parameter `accepts_exit_code: bool = False` to `add_restart_callback` and save it alongside it.

jakkdl:
made a somewhat messy (esp when you look at the typing) implementation for restarter callback taking the exit status, and wrote (copy-paste+modified) a test for it. A more creative test would kill the process in several different ways, but that should be covered by `TestKernelManagerExitStatus``

rgbkrk:
Closing as stale. Feel free to make a new PR or re-open if needed. Thanks @jakkdl

PR #961, opened 2023-07-26
Support external kernels

davidbrochart:
See https://github.com/jupyter-server/jupyter_server/pull/1305.

### DIFF
--- a/jupyter_client/multikernelmanager.py
+++ b/jupyter_client/multikernelmanager.py
@@ -2,11 +2,13 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import asyncio
+import json
 import os
 import socket
 import typing as t
 import uuid
 from functools import wraps
+from pathlib import Path
 
 import zmq
 from traitlets import Any, Bool, Dict, DottedObjectName, Instance, Unicode, default, observe
@@ -15,7 +17,7 @@ from traitlets.utils.importstring import import_item
 
 from .kernelspec import NATIVE_KERNEL_NAME, KernelSpecManager
 from .manager import KernelManager
-from .utils import ensure_async, run_sync
+from .utils import ensure_async, run_sync, utcnow
 
 
 class DuplicateKernelError(Exception):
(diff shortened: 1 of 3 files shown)


jupyter_client/multikernelmanager.py
```diff
@@ -123,6 +130,56 @@ def __del__(self):
...
+                    km.shell_port = connection_info["shell_port"]
+                    km.iopub_port = connection_info["iopub_port"]
+                    km.stdin_port = connection_info["stdin_port"]
+                    km.control_port = connection_info["control_port"]
+                    km.hb_port = connection_info["hb_port"]
+                    km.ip = connection_info["ip"]
+                    km.transport = connection_info["transport"]
+                    km.session.key = connection_info["key"].encode()
+                    km.session.signature_scheme = connection_info["signature_scheme"]
+                    km.kernel_name = connection_info["kernel_name"]
```
kevin-bates, line 178: Could we leverage the [`load_connection_info()`](https://github.com/jupyter/jupyter_client/blob/508982f11a3175089b572a0c8277542d9fbd0729/jupyter_client/connect.py#L530) method on the newly-created kernel manager instance?

davidbrochart: Good point, thanks!

jupyter_client/multikernelmanager.py
```diff
@@ -105,9 +108,14 @@ def _context_default(self) -> zmq.Context:
         return zmq.Context()
 
     connection_dir = Unicode("")
+    external_connection_dir = Unicode(None, allow_none=True)
```
kevin-bates, line 111: Since we're silent if this value is set, but not an existing directory, it seems like a validate-handler would be helpful here that, when set, it ensures the value corresponds to an existing directory and presents a helpful message otherwise.

davidbrochart: On the other side, validating the existence of the directory at start time is kind of restrictive - it doesn't allow for directories that are created after startup. Also, validating the directory only at start time doesn't check if it's removed afterwards.
So I'd prefer to keep the flexibility of the current code.

davidbrochart:
@blink1073 I'd like to make a release, but it looks like I don't have admin right on this repository.

blink1073:
I've updated your rights on the repo.

PR #972, opened 2023-09-11
Do not use datetime.utcnow() that is deprecated in Python 3.12

ellert:
* Do not use datetime.utcnow() that is deprecated in Python 3.12
* Import session.utcnow() into utils instead of reimplementing it

### DIFF
--- a/jupyter_client/session.py
+++ b/jupyter_client/session.py
@@ -197,7 +197,7 @@ def default_secure(cfg: t.Any) -> None:  # pragma: no cover
 
 def utcnow() -> datetime:
     """Return timezone-aware UTC timestamp"""
-    return datetime.now(utc)  # noqa
+    return datetime.now(utc)
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


davidbrochart (approved):
Thanks @ellert.
I don't know why we had this implementation of `utcnow` in `utils.py`, but it looks like it is not needed indeed.

jupyter_client/session.py
```diff
@@ -197,7 +197,7 @@ def default_secure(cfg: t.Any) -> None:  # pragma: no cover
 
 def utcnow() -> datetime:
     """Return timezone-aware UTC timestamp"""
-    return datetime.utcnow().replace(tzinfo=utc)  # noqa
+    return datetime.now(utc)  # noqa
```
davidbrochart, line 200:
```suggestion
    return datetime.now(utc)
```

jupyter_client/utils.py
```diff
@@ -4,10 +4,11 @@
 - vendor functions from ipython_genutils that should be retired at some point.
 """
 import os
-from datetime import datetime, timedelta, tzinfo
 
 from jupyter_core.utils import ensure_async, run_sync  # noqa: F401  # noqa: F401
 
+from .session import utcnow
```
davidbrochart, line 10:
```suggestion
from .session import utcnow  # noqa
```

PR #978, opened 2023-10-10
Test on python 3.12

blink1073:


tornaria:
Let me point out that this is actually failing *exactly* as I reported in #979.

Indeed, check out: https://github.com/jupyter/jupyter_client/actions/runs/6463461968/job/17571024062?pr=978#step:4:679

Your workflow is actually ignoring the check failure, but it is there in the log.

tornaria:
> We're not ignoring it, we're retrying. The event loop situation is not ideal, but we are doing our best.

Sure. I was just trying to be helpful.

PR #984, opened 2023-10-16
Enable strict typing

blink1073:
Leave a few rules on for now that would need a lot more work:

- warn_return_any = false
- disallow_any_generics = false
- no_implicit_reexport = false

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -5,6 +5,8 @@ related to writing and reading connections files.
 """
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+from __future__ import annotations
+
 import errno
 import glob
 import json
@@ -14,7 +16,7 @@ import stat
 import tempfile
 import warnings
 from getpass import getpass
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, Union, cast
 
 import zmq
 from jupyter_core.paths import jupyter_data_dir, jupyter_runtime_dir, secure_write
@@ -34,7 +36,7 @@ KernelConnectionInfo = Dict[str, Union[int, str, bytes]]
 
 
 def write_connection_file(
-    fname: Optional[str] = None,
+    fname: str | None = None,
     shell_port: int = 0,
     iopub_port: int = 0,
     stdin_port: int = 0,
@@ -46,7 +48,7 @@ def write_connection_file(
     signature_scheme: str = "hmac-sha256",
     kernel_name: str = "",
     **kwargs: Any,
-) -> Tuple[str, KernelConnectionInfo]:
+) -> tuple[str, KernelConnectionInfo]:
     """Generates a JSON config file, including the selection of random ports.
 
     Parameters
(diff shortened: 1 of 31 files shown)


jupyter_client/connect.py
```diff
@@ -24,12 +26,17 @@
...
+
+    from .session import Session
+
 # Define custom type for kernel connection info
 KernelConnectionInfo = Dict[str, Union[int, str, bytes]]
 
 
 def write_connection_file(
-    fname: Optional[str] = None,
+    fname: str | None = None,
```
Carreau, line 39: I dont' think we should do this if we still support 3.8 and 3.9, this beciome valid syntax, but `typing.get_type_hints()` will fail on `|` for Union syntax on 3.8 and 3.9, so we risk breaking things.

Carreau:
```
from __future__ import annotations
from typing import Optional, get_type_hints


def write_connection_file(fname: str | None = None) -> None:
    pass


get_type_hints(write_connection_file)
```
---
```
$ python f.py
  File "/Users/bussonniermatthias/dev/jupyter_client/f.py", line 9, in <module>
    get_type_hints(write_connection_file)
  File "/Users/bussonniermatthias/miniforge3/envs/39/lib/python3.9/typing.py", line 1386, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/Users/bussonniermatthias/miniforge3/envs/39/lib/python3.9/typing.py", line 254, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/Users/bussonniermatthias/miniforge3/envs/39/lib/python3.9/typing.py", line 493, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

Carreau: Yeah, not your fault, it's a common confusion. I wouldn't have caught it if not for a similar problem recently.

blink1073: Well now I'm not so sure what to do, because this also applies to `type[]`, `set[]`, `list[]`, `dict[]`, etc.  I don't think we need to claim to support *type hints* for Python 3.8 or 3.9.

blink1073: At least not at run time.

blink1073: Nothing prevents folks from calling `get_type_hints` on their own code, I just verified.

Carreau: as you made the PR i think it's fine, but maybe we ca avoid this for subsequent libs.

blink1073: Yeah that sounds like a good plan.  I added this, which I'll use as a template:

```
  # non-pep585-annotation
  "UP006",
  # non-pep604-annotation
  "UP007",
```

Carreau:
BTW @blink1073 when I run mypy directly instead of via hatch, I get [many more errors](https://gist.github.com/Carreau/33980adcb37886a54ce08a89112e254a). Any idea why ?

blink1073:
The hatch version has some min version pins to pick up newer typings.

Carreau:
> The hatch version has some min version pins to pick up newer typings.

Yeah it helps to update the versions, but I still get some errors.

That's fine I can use hatch for now.

PR #986, opened 2023-10-23
Unused `*args` in `KernelManager`'s `__init__`

Carreau:
I think that if it is unused, then it is likely a mistake from the caller that meant to pass a kwarg.

Should it maybe get a deprectation period ?

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -9,6 +9,7 @@ import signal
 import sys
 import typing as t
 import uuid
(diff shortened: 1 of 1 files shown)


blink1073:
Yeah I think it should be deprecated and removed in 9.0.

Carreau:
Rebased on main, should fix linting.

blink1073 (approved):
Thanks!

PR #987, opened 2023-10-23
Allow to update kernels env in between restart.

Carreau:
This allow to update the running env of kernels.

This will allow resolves some inconsistencies between kernel restart, and actually stopping and starting a kernel session.

In parsley via a few lines in Jupyter Server in update_session, this will let us update the `__session__` which is supposed to reflect some information about the current session, and help correct ipython/ipykernel#1102 where renaming a notebook is not reflected in the kernel unless you manually stop/start the session, or restart the server.


blink1073 (approved):
Thanks!

PR #989, opened 2023-10-25
BUG: Fix Kwarg only in update_env

Carreau:
closes jupyterlab/jupyterlab#15301

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -19,6 +19,7 @@ from jupyter_core.utils import run_sync
 from traitlets import (
     Any,
     Bool,
+    Dict,
     DottedObjectName,
     Float,
     Instance,
@@ -206,7 +207,7 @@ class KernelManager(ConnectionFileMixin):
         return self.kernel_name in {"python", "python2", "python3"}
 
     # Protected traits
-    _launch_args: Any = Any()
+    _launch_args: t.Optional[Dict] = Dict(allow_none=True)
     _control_socket: Any = Any()
 
     _restarter: Any = Any()
@@ -319,7 +320,7 @@ class KernelManager(ConnectionFileMixin):
 
         if self.kernel_spec:  # type:ignore[truthy-bool]
             ns["resource_dir"] = self.kernel_spec.resource_dir
-
+        assert self._launch_args is not None
         ns.update(self._launch_args)
 
         pat = re.compile(r"\{([A-Za-z0-9_]+)\}")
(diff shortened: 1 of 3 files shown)


blink1073:
Looks like there's still an issue with the downstream server tests.

```python
    File "/home/runner/test_venv/lib/python3.11/site-packages/tornado/web.py", line 1786, in _execute
      result = await result
               ^^^^^^^^^^^^
    File "/home/runner/test_venv/downstream_test/jupyter_server/jupyter_server/services/sessions/handlers.py", line 172, in patch
      await sm.update_session(session_id, **changes)
    File "/home/runner/test_venv/downstream_test/jupyter_server/jupyter_server/services/sessions/sessionmanager.py", line 469, in update_session
      self.kernel_manager.update_env(kernel_id=kernel_id, env=self.get_kernel_env(path, name))
    File "/home/runner/test_venv/lib/python3.11/site-packages/jupyter_client/multikernelmanager.py", line 226, in update_env
      self._kernels[kernel_id].update_env(env=env)
    File "/home/runner/test_venv/lib/python3.11/site-packages/jupyter_client/manager.py", line 284, in update_env
      self._launch_args['env'].update(env)
      ~~~~~~~~~~~~~~~~~^^^^^^^
  KeyError: 'env'
```

blink1073:
I'm reverting this change in server for now.

Carreau:
I pushed a check that env exists.

blink1073:
Have you tested this with server to 2.8.1?

Carreau:
Ok, I've been more agressive on the condition, checking that everything exists and are dicts, and added all the corresponding test cases.

I've tested on Jupyter_server  2.7.0, 2.9.1 (does not crash, but does nothing), 
and on 2.9.0 (it does work an update `__session__`).

Carreau:
Let me see if I can fix the typing of `_launch_args`, that is set to `Any`, I think we can make it t.Optional[Dict]

Carreau:
ok, I can't make it ` t.Optional[Dict[str, Any]]`, as `Dict` is not subscriptable, but I did make it `t.Optional[Dict]`

jupyter_client/manager.py
```diff
@@ -314,7 +320,7 @@ def format_kernel_cmd(self, extra_arguments: t.Optional[t.List[str]] = None) ->
 
         if self.kernel_spec:  # type:ignore[truthy-bool]
             ns["resource_dir"] = self.kernel_spec.resource_dir
-
+        assert self._launch_args is not None
```
Carreau, line 323: This helps mypy, the line below would fail anyway if it was None.

blink1073: | ok, I can't make it  t.Optional[Dict[str, Any]], as Dict is not subscriptable

You can use  `t.Optional["Dict[str, Any]"]` to get around the runtime issue.

Carreau: Ah ! True, let's try

blink1073:
Papermill failure is known: https://github.com/nteract/papermill/issues/737

blink1073 (approved):
Thanks!

PR #997, opened 2023-11-18
Remove direct usage of tornado

blink1073:
See https://github.com/jupyter-server/jupyter_server/issues/1362 for discussion.  This will require a major version change since the loops are public properties.

- [x] Make the app use `asyncio.main`
- [x] Move the `get_event_loop` function to `jupyter_core` (renamed to `ensure_event_loop`)
- [x] Get unit tests passing
- [x] Get downstreams passing
- [x] Document changes in migration guide


blink1073:
Needs some direct tests for the new `ZMQStream` class, including the ability to pass through, such as `if stream is not None and not stream.closed()`.

PR #1010, opened 2024-02-18
Improve handling of async sockets

blink1073:
Attempt to fix https://github.com/voila-dashboards/voila/issues/1428#issuecomment-1927025620


blink1073:
@trungleduc this might address the issue you're seeing.

PR #1023, opened 2024-05-14
Use non-blocking zmq Poller

### LINKED ISSUE #1022: zmq.Poller is not async
The line https://github.com/jupyter/jupyter_client/blob/0a19278ed8d4090080a44760498a38ad9d4e0d99/jupyter_client/client.py#L533 makes the code blocking.

We should use `zmq.asyncio.Poller`.

fcollonval:
Fixes #1022

### DIFF
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -220,12 +220,11 @@ class ZMQSocketChannel:
         ident, smsg = self.session.feed_identities(msg)
         return self.session.deserialize(smsg)
 
-    def get_msg(self, timeout: t.Optional[int] = None) -> t.Dict[str, t.Any]:
+    def get_msg(self, timeout: t.Optional[float] = None) -> t.Dict[str, t.Any]:
         """Gets a message if there is one that is ready."""
         assert self.socket is not None
-        if timeout is not None:
-            timeout *= 1000  # seconds to ms
-        ready = self.socket.poll(timeout)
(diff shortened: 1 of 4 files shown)


davidbrochart (approved):
Failures are unrelated, see #1024.

ianthomas23:
Pinning `pytest<8.2` helps with the test failures.

jupyter_client/channels.py
```diff
@@ -220,7 +220,7 @@ def _recv(self, **kwargs: t.Any) -> t.Dict[str, t.Any]:
         ident, smsg = self.session.feed_identities(msg)
         return self.session.deserialize(smsg)
 
-    def get_msg(self, timeout: t.Optional[float] = None) -> t.Dict[str, t.Any]:
+    def get_msg(self, timeout: t.Optional[int] = None) -> t.Dict[str, t.Any]:
```
minrk, line 223: This should accept a float. `get_msg(timeout=0.25)` is valid. The cast to int should happen after multiplying by 1000 to get ms.

minrk (commented):
👍 to the headline Poller-related change, but the unrelated get_msg type change isn't right, I think.

minrk (approved):

krassowski:
It would be great to see this released :) Is the timeout in one of the Windows tests related?

davidbrochart:
I don't think it's related. I'd like to merge but I cannot bypass the rules. Maybe @blink1073 can help?

PR #1028, opened 2024-06-13
Parameterized kernel specs

AnastasiaSliusar:
## References

This work is in progress as JEP [jupyter/enhancement-proposals#87](https://github.com/jupyter/enhancement-proposals/pull/87).
This PR should be reviewed when the JEP has been accepted

## Code changes
 
 - Replacing kernel spec variables for arg and env that are in `kernel.json` on values from `custom_kernel_specs` parameter which is configured by a user.  `custom_kernel_specs` is passed from Jupyter Lab through jupyter_server.
 - Clearing kernel spec variables that are in `kernel.json` if they are present into this file but `custom_kernel_specs` parameter is missed
 - Adding basic validation to check whether custom_kernel_specs does not include anything wrong
 - Adding support if a kernel.json file includes customization for a kernel but a user has not selected anything then a parameterized kernel will be run by using default values from its kernel.json file

## How to run
 - Please use this PR with next PRs together:
   https://github.com/jupyterlab/jupyterlab/pull/16487
   https://github.com/jupyter-server/jupyter_server/pull/1431

### DIFF
--- a/jupyter_client/kernelspec.py
+++ b/jupyter_client/kernelspec.py
@@ -240,18 +240,18 @@ class KernelSpecManager(LoggingConfigurable):
         print(kspec.argv)
         is_secure = self.check_kernel_is_secure(kspec=kspec)
         if is_secure == True:
-            print('---is_secure---')
-            return kspec # a kernel spec is allowed
+            print("---is_secure---")
+            return kspec  # a kernel spec is allowed
         else:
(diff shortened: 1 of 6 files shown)


rgbkrk:
Would you like to rebase this PR to help it move along? I'd love to get the JEP accepted so we can move forward and not have this grow stale.

PR #1036, opened 2024-09-17
test on python 3.13-dev

andife:


.github/workflows/main.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.8", "3.12"]
+        python-version: ["3.8", "3.12","3.13.0-rc.2"]
```
minrk, line 41: Thanks! `3.13-dev` will get the latest prerelease without needing to update here:

```suggestion
        python-version:
          - "3.8"
          - "3.12"
          - "3.13-dev"
```

andife:
LGTM thanks

https://peps.python.org/pep-0719/ 3.13 is released in two weeks?

PR #1040, opened 2024-09-21
don't include link-local addresses (169.254) in public_ips

### LINKED ISSUE #1039: Psutil changes in `public_ips()` causes breaking changes
Introduced in [5d4888116a69b8f8edb69e51a22b8d4df3a54741](https://github.com/jupyter/jupyter_client/commit/5d4888116a69b8f8edb69e51a22b8d4df3a54741#diff-4191d4a06a30c2b9acffc3ba0ae4f8dc02c8ce01478a650fb7d4ba0eb40b4133R160) in `jupyter_client/localinterfaces.py`, the changes made to `public_ips()` method breaks existing deployments.

Previously, the following yielded the correct public ip address:

```python
ip = public_ips()[0]
print(ip)  # 10.0.XX.XX
```

Now, it yields the linked local address:

```python
ip = public_ips()[0]
print(ip)  # 169.254.XXX.X
```

For now, I am manually filtering out IPs that are one of: ipv4 link local ips (unicast), ipv6 ips (unicast), ipv4 multicast ips, or ipv6 multicast ips, but would love to see this get fixed upstream.

minrk:
`public_ips()` is meant to return addresses which might be connectable from other machines, which link-local are not.

closes #1039

### DIFF
--- a/jupyter_client/localinterfaces.py
+++ b/jupyter_client/localinterfaces.py
@@ -8,11 +8,11 @@ import re
 import socket
 import subprocess
 from subprocess import PIPE, Popen
-from typing import Any, Callable, Iterable, Sequence
+from typing import Any, Callable, Iterable, Mapping, Sequence
 from warnings import warn
 
-LOCAL_IPS: list = []
-PUBLIC_IPS: list = []
+LOCAL_IPS: list[str] = []
+PUBLIC_IPS: list[str] = []
 
 LOCALHOST: str = ""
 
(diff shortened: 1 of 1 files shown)


ivanov (approved):
Thanks, Min!

not to code-golf things, too much, but any reason not to add 
```python
or  addr.startswith("169.254.")
```
to the lines that already have `addr.startswith("127.")` logic (possibly with an appropriate group, where line 89 would become 

```python
if not (ip.startswith("127.") or ip.startswith("169.254.")):`
```

minrk:
No worries! I added the conditions to avoid link-local addresses being assigned to LOCALHOST. So I could add the `or`, but then I'd have to move the LOCALHOST to its own condition with the same match minus 169z

There is a lot of duplication here, so maybe I'll try to consolidate the whole thing.

minrk:
consolidated to `_populate_from_dict` since netifaces and psutil were essentially the same.

ivanov:
thanks, Min, for the patch, the explanation, and the additional cleanup :tada:

minrk:
Thanks for the review and merge!

PR #1045, opened 2024-11-18
Add messaging docs for optional features and kernel subshells

ianthomas23:
This adds messaging documentation for optional features [JEP92](https://github.com/jupyter/enhancement-proposals/pull/92) and kernel subshells [JEP91](https://github.com/jupyter/enhancement-proposals/pull/91/files).

Summary of changes:

- Optional features [JEP92](https://github.com/jupyter/enhancement-proposals/pull/92):
  - Return optional `supported_features` in `kernel_info_reply`.
  - Deprecate use of `debugger` in `kernel_info_reply` as this is replaced by `supported_features=['debugger']`. It will have to be a long deprecation period whilst other libraries in the Jupyter ecosystem adopt use of `supported_features`.

- Kernel subshells [JEP91](https://github.com/jupyter/enhancement-proposals/pull/91/files):
  - Return `supported_features=['kernel subshells']` in `kernel_info_reply`.
  - New control request/reply messages for `create`, `delete` and `list` subshells.
  - Add optional `subshell_id` to message header. 

The version added/changed for all of these is 5.5. I am unsure of the protocol around official releases here.

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -118,6 +118,8 @@ so that frontends can label the various messages in a meaningful way.
         "msg_type": str,
         # the message protocol version
         "version": "5.0",
+        # Optional subshell_id
+        "subshell_id": str | None,
     }
 
 .. note::
@@ -139,6 +141,11 @@ so that frontends can label the various messages in a meaningful way.
     from the kernel have a different kernel session id than prior to the disconnect,
     the client should assume that the kernel was restarted.
 
+    The ``subshell_id`` is only used in shell messages of kernels that support
+    subshells (:ref:`kernel_subshells`). If it is not included or is ``None`` then the
+    shell message is handled by the parent subshell (main shell), if it is a string
+    subshell id then it is handled by the subshell with that id.
+
 .. versionchanged:: 5.0
 
     ``version`` key added to the header.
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):
Thanks @ianthomas23.
Just one minor comment: should we use capital letters for "ID"? I see we use both "ID" and "id" in the current documentation, but I would prefer "ID".

ianthomas23:
> Just one minor comment: should we use capital letters for "ID"? I see we use both "ID" and "id" in the current documentation, but I would prefer "ID".

Good idea. I've switched to using "ID" throughout the messaging page for consistency.

ianthomas23:
I don't have merge rights here.

docs/messaging.rst
```diff
@@ -1313,6 +1334,86 @@ of the debugger to the ``global`` scope to inspect it after debug session.
...
 
+.. _kernel_subshells:
+
+Kernel subshells
+----------------
+
+Kernel subshells are separate threads of execution within the same kernel process that
+were introduced by
+`Jupyter Enhancement Proposal 91 <https://github.com/jupyter/enhancement-proposals/pull/91>`_.
+Kernels supporting subshells must return ``'supported_features'=['kernel subshells']``
```
davidbrochart, line 1345: Maybe say that kernels supporting subshells must include `'kernel subshells'` in `'supported_features'`, since there could be other features in there?

ianthomas23: Done.

PR #1050, opened 2025-01-07
Update _load_ips_psutil() docstring

cben:
Followup to #1033, apparently docstring got copy-pasted from _load_ips_netifaces


minrk:
thanks!

PR #1052, opened 2025-01-13
Handle automatic docstring dedenting in Python 3.13

juliangilbey:
Python 3.13 has changed the way it handles docstrings; see https://docs.python.org/3/whatsnew/3.13.html#other-language-changes and https://github.com/python/cpython/issues/81283

As a result of this, the `reqrep()` function in `jupyter_client.client` gives a poorly indented docstring, and as a result Sphinx fails.

This PR addresses this by ensuring that the docstring changes that `reqrep()` performs respect the indentation of the docstring it has been passed.  It has been tested on Python 3.12 and Python 3.13.


rgbkrk:
Can you rebase this PR against main to get it up to date?

PR #1054, opened 2025-02-08
Set protocol version to 5.4

JamesWrigley:
I believe this was incorrectly left as 5.3 since the blocking and async clients both expect the shutdown message to come on the control channel, e.g. https://github.com/jupyter/jupyter_client/blob/6ea834d69d9404b9b67fbc066ee8bdf1d7d2ede6/jupyter_client/blocking/client.py#L70-L71


JamesWrigley:
(bump)

Issue #1018, opened 2024-03-13
Add `jupyter kernelspec remove --missing` to remove all missing kernels.


wpk-nist-gov:
I often install kernels from project virtualenvs (`my-project/.venv/`) that eventually get removed.  I can manually remove a kernelspec with `jupyter kernelspec remove my-project` but forget to do this, then forget the missing kernel name, etc.  Instead, I use a script that checks if the python executable the kernel points to exists, and if it doesn't, removes the kernel.  

I think such functionality would be nice to have directly from `jupyter kernelspec remove`.  I'd be happy to attempt a PR with this feature if you agree.  Thanks!

PR #1056, opened 2025-02-24
feat: added `--missing` flag to list and remove subcommands

wpk-nist-gov:
Addresses #1018.

This adds a flag `--missing` to `list` and `remove` subcommands of `jupyter kernelspec`.  By "missing", we mean that the interpreter (the first argument of spec["spec"]["argv"] of the json specification) does not exist.  This simplifies removing kernels pointing to missing environments.

I'm not sure how best to test this functionality.  Basic smoke tests seem to work.  I'm happy to put some more work into this if it seems worthwhile to include in `jupyter-client`.

Thanks!

### DIFF
--- a/jupyter_client/kernelspecapp.py
+++ b/jupyter_client/kernelspecapp.py
@@ -8,6 +8,7 @@ import json
 import os.path
 import sys
 import typing as t
+from pathlib import Path
 
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets import Bool, Dict, Instance, List, Unicode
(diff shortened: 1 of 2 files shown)


jupyter_client/kernelspecapp.py
```diff
@@ -337,5 +369,20 @@ def start(self) -> None:
...
 
+def _limit_to_missing(
+    paths: dict[str, str], specs: dict[str, t.Any]
+) -> tuple[dict[str, str], dict[str, t.Any]]:
+    specs_: dict[str, t.Any] = {
+        k: v
+        for k, v in specs.items()
+        # If have kernel installed from current environment
+        # Probably not the best way to do this, but it works in edge cases I've run into.
+        if (prog := v["spec"]["argv"][0]) != "python" and not Path(prog).exists()
```
minrk, line 380: This should use `shutil.which` to resolve things on $PATH, since it is not a requirement that `argv[0]` is an absolute path.

minrk (commented):
Nice! This sounds like a useful feature. I think we should should have it.

Can you add a test that makes sure a kernel with e.g. `"bash"` (no path) as `argv[0]` doesn't get deleted?

jupyter_client/kernelspecapp.py
```diff
@@ -337,5 +369,22 @@ def start(self) -> None:
...
+) -> tuple[dict[str, str], dict[str, t.Any]]:
+    from shutil import which
+
+    missing: dict[str, t.Any] = {}
+    for name, data in specs.items():
+        exe = data["spec"]["argv"][0]
+        # if exe exists or is on the path, keep it
+        if Path(exe).exists() or which(exe):
+            continue
+        missing[name] = data
```
wpk-nist-gov, line 383: This should address if the path exists or is on the path.

PR #1059, opened 2025-04-08
Fix blocking issue in Voila (nbclient case)

martinRenou:


martinRenou:
Closing as I was just curious to see how the CI behaves with this change

martinRenou:
@davidbrochart @JohanMabille @minrk friendly ping, would you have any thought on this?

jupyter_client/session.py
```diff
@@ -810,10 +810,6 @@ def send(
             # ZMQStreams and dummy sockets do not support tracking.
             track = False
 
-        if isinstance(stream, zmq.asyncio.Socket):
-            assert stream is not None  # type:ignore[unreachable]
-            stream = zmq.Socket.shadow(stream.underlying)
```
minrk, line 815: without this, it's not actually guaranteed the send will happen, since it's not awaited (`zmq.asyncio` does _currently_ immediately schedule sends, but it doesn't guarantee that it will continue to do so), and the type of `tracker` will be incorrect `Awaitable[Event]` instead of `Event` (and a different type when `copy=True` vs `copy=False`.

We could add async polling, either here or (perhaps easier) in nbclient, so you can avoid calling this when it would block:

```python
events = await async_socket.poll(timeout_ms, zmq.POLLOUT)
if events & zmq.POLLOUT:
    session.send(async_socket) # won't block
else:
    # ????
```

JohanMabille: I might be missing some bits here, but would it be possible to handle that in the ` AsyncZMQSocketChannel`  instead? I.e. overwrite the send method, do the proper stuff, and then call session.send? This would avoid the Socket implementation to leak in the Session class.

minrk: Yeah, that's probably where it belongs

PR #1060, opened 2025-04-08
Support date serialization in messages

### LINKED ISSUE #1058: Cannot handle message data with date type
Example using plotly 6:

```python
from datetime import date, datetime
import polars as pl
import plotly.express as px
from plotly.graph_objs import FigureWidget

df=pl.DataFrame({'x': [date(2025, 1, 1), date(2025, 1, 2)], 'y': [1, 2]})
fig=px.line(df, x='x', y='y')

fig.show() # renders correctly

FigureWidget(fig).show() # fails
```

The last line results in:

```
ValueError: Can't clean for JSON: datetime.date(2025, 1, 1)
```

coming from
```
File ~/dev/jupyter_client/jupyter_client/jsonutil.py:195, in json_clean(obj)
    192     return obj.strftime(ISO8601)
    194 # we don't understand it, it's probably an unserializable object
--> 195 raise ValueError("Can't clean for JSON: %r" % obj)
```

sebwills:
Resolves #1058.

fwiw the (original?) copy of json_clean in https://github.com/ipython/ipykernel/blob/2c212648775646f1c92e8f3e977806dea14186da/ipykernel/jsonutil.py#L160 has date support.

### DIFF
--- a/jupyter_client/jsonutil.py
+++ b/jupyter_client/jsonutil.py
@@ -8,7 +8,7 @@ import types
 import warnings
 from binascii import b2a_base64
 from collections.abc import Iterable
-from datetime import datetime
+from datetime import datetime, date
 from typing import Any, Optional, Union
 
 from dateutil.parser import isoparse as _dateutil_parse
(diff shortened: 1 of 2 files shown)


sebwills:
Hi @minrk, If you have a moment to see if you think this is an acceptable solution to the problem in #1058 that would be much appreciated. Thanks!

minrk (approved):
Looks okay, thank you!

sebwills:
@minrk Great, thanks. One of the PR checks failed but I don't think it's due to my change (and I can't rerun it).

sebwills:
@minrk let me know if I need to do anything to get this green and merged.

minrk:
Thanks!

PR #1069, opened 2025-08-19
Bugfix / Correct stopping async cell execution

truongle1501:
Regarding document:
https://github.com/jupyter/jupyter_client/blob/eb96748e0afb970c806653824540dbadc83ffc84/docs/messaging.rst?plain=1#L1634-L1639

Context:
The async cell execution stopped too early because it hit the below condition while the cell is being executed. This also means there was a moment when iopub message satisfied both `msg["parent_header"].get("msg_id") == msg_id` and `execution_state='idle'`
https://github.com/jupyter/jupyter_client/blob/eb96748e0afb970c806653824540dbadc83ffc84/jupyter_client/client.py#L560-L572

This PR aims to set a flag to allow stopping the cell execution only when it saw 'busy' state, which forces the stop condition to run after cell execution starts.

### DIFF
--- a/jupyter_client/client.py
+++ b/jupyter_client/client.py
@@ -572,10 +572,10 @@ class KernelClient(ConnectionFileMixin):
             if msg["header"]["msg_type"] == "status":
                 state = msg["content"]["execution_state"]
                 # allow to stop
-                if state == 'busy':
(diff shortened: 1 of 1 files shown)


Carreau:
(rebased)

PR #1074, opened 2025-09-29
Add Kernel Lifecycle State Management Mixin

Zsailer:
Addresses #763. 

This PR introduces a state machine for tracking kernel lifecycle states using a mixin approach that can be easily adopted by any existing kernel manager implementation.

  The `KernelManagerStateMixin` design allows developers to add state tracking to **any existing kernel
  manager** by simply including it in the inheritance chain - no configuration or code changes required.
  This opt-in approach means:

  - Existing implementations continue working unchanged
  - Developers can choose when and where to add state tracking
  - Zero risk of breaking current functionality
  - Easy to test and evaluate before wider adoption

  I intentionally did not add this to the default kernel managers yet, allowing the community to
  evaluate the approach first.

  ### Usage

  ```python
  from jupyter_client import KernelManagerStateMixin

  class MyKernelManager(KernelManagerStateMixin, ExistingKernelManager):
      pass  # No changes needed to existing logic

  manager = MyKernelManager()
  manager.start_kernel()
  print(manager.is_started)  # True
  ```

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -10,8 +10,10 @@ To add lifecycle state tracking to a kernel manager, simply inherit from `Kernel
 from jupyter_client.lifecycle import KernelManagerStateMixin
 from jupyter_client.manager import KernelManager
 
+
 class StatefulKernelManager(KernelManagerStateMixin, KernelManager):
     pass
+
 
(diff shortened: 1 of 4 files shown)


krassowski:
Is this still relevant?

PR #1076, opened 2025-10-08
Start testing on 3.14 and 3.14-free-threaded

Carreau:


Carreau:
Sure, and feel free to push on my branch. I'm debugging some issues on pytest-jupyter as well. 

It might be due to the Event Loop policies deprecation warnings that are the source of the issue.

minrk:
There are definitely once again sockets and contexts being created and not closed in the tests, and we need to track those down.

Carreau:
should all those non-closed sockets and open/close pairs become context manager (in a perfect world).

minrk:
> should all those non-closed sockets and open/close pairs become context manager (in a perfect world).

Yes, KernelManager and KernelClient should both be context managers.

zmq sockets are already context managers and have been for some time

tests/test_multikernelmanager.py
```diff
@@ -613,8 +640,9 @@ async def test_bad_kernelspec_pending(self):
         assert kernel_id in km.list_kernel_ids()
         await ensure_future(km.shutdown_kernel(kernel_id))
         assert kernel_id not in km.list_kernel_ids()
+        km.context.destroy(linger=0)
 
-    @gen_test
+    @gen_test(timeout=30)
```
Carreau:
```suggestion
    @gen_test(timeout=TIMEOUT)
```
Maybe ? `TIMEOUT=90` at top of file

tests/test_multikernelmanager.py
```diff
@@ -643,4 +671,6 @@ def record_activity(msg_list):
             await asyncio.sleep(0.1)
 
         client.stop_channels()
+        stream.close()
```
Carreau: Should these be in try/finally in case of errors ?

Carreau (commented):
+1, a couple of questions, but good as is.

.github/workflows/main.yml
```diff
@@ -53,18 +57,28 @@ jobs:
...
-      - name: Run the tests
-        if: ${{ !startsWith(matrix.python-version, 'pypy') && !startsWith(matrix.os, 'windows') }}
+
+      - name: disable warnings on pypy or windows
+        if: ${{ startsWith(matrix.python-version, 'pypy') || startsWith(matrix.os, 'windows') }}
         run: |
-          hatch run cov:test --cov-fail-under 75 || hatch run test:test --lf
-      - name: Run the tests on pypy
+          echo "PYTEST_ADDOPTS=${PYTEST_ADDOPTS:-} -W default" >> $GITHUB_ENV
+
```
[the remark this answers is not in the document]
minrk: Yes, this is not a change, it simplifies the 3 different test invocations below by setting the options before using the same pytest invocation.

Issue #937, opened 2023-03-18
Document cell_id inside docs/wrapperkernels.rst


lucafabbian:
Hi!
I tried to implement my own kernel, but I struggled to find a way to identify cells.

That identifier actually exists, but is not documented anywhere inside [this page](https://jupyter-client.readthedocs.io/en/stable/wrapperkernels.html).

Maybe you may change the `do_execute` signature in the echo example and use this one instead?

```python
    def do_execute(self, code, silent, store_history=True, user_expressions=None,
                   allow_stdin=False, *, cell_id=None):

```

I think it would be really useful to document better such an important feature.

c-h-russell-walker:
Hello, I am also hoping to use the `cell_id` parameter as seen here as part of the ipython `8.3.0` release:
https://ipython.readthedocs.io/en/stable/whatsnew/version8.html#ipython-8-3-0

Is there a chance this could be added to this library?
Thanks so much.

PR #1102, opened 2026-01-23
docs: document cell_id parameter in wrapperkernels.rst

anny0811:
This PR implements the documentation for the cell_id parameter as requested in #937. It updates the do_execute method signature in wrapperkernels.py and provides the corresponding reference in wrapperkernels.rst.

Key Changes:
~ Added cell_id to the do_execute method signature.
~ Updated reStructuredText documentation to include parameter details.

### DIFF
--- a/docs/wrapperkernels.rst
+++ b/docs/wrapperkernels.rst
@@ -50,7 +50,7 @@ following methods and attributes:
 
      Other keys may be added to this later.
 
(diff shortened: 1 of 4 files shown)


minrk:
this doesn't address #937 because while it adds an argument, it is not used (and cannot be). It is also not a docs change, it is an (incomplete) API change.

PR #1104, opened 2026-02-06
test: ensure async recv reply raises TimeoutError on timeout

CoderLink06:
Adds a focused async test verifying that KernelClient._async_recv_reply raises TimeoutError when no reply is received before the timeout.

The test uses a minimal fake channel to simulate a non-responsive kernel and asserts the timeout behavior.


Carreau:
Same as other PRs, please focus on one fix and do it properly.

PR #1105, opened 2026-02-07
test: ensure execute fails when channels are stopped

CoderLink06:
Adds a test verifying that executing code after stopping kernel channels raises an AssertionError. This documents and protects the current failure behavior when a client attempts to send messages without active channels.


Carreau:
This looks spammy like a few of your other PRs.

PR #1107, opened 2026-02-11
fix: allow get_msgs timeout to prevent blocing behavior

CoderLink06:
## Summary

This change adds a timeout parameter to `get_msgs()` so callers can control blocking behavior when polling messages.

Previously, `get_msgs()` repeatedly called `get_msg()` without timeout control, which could lead to unintended blocking depending on usage.

## Changes

- Added `timeout` argument to `get_msgs()` (sync and async)
- Forwarded timeout to `get_msg()`
- Preserves existing behavior with default non-blocking polling

## Why

This gives callers explicit control over message polling and prevents accidental blocking loops.

## Testing

Existing test suite passes locally.


Carreau:
Same comment as other PR, please pay attention to what you are submitting.

PR #1108, opened 2026-02-20
Added documentation for the handshake pattern

JohanMabille:


Carreau:
rebased on main, I'm leanign toward merging even if test are failing.

PR #1109, opened 2026-02-20
Ci fixes

JohanMabille:


krassowski:
I think this is no longer needed?

JohanMabille:
Indeed.

PR #1110, opened 2026-05-05
Add ZMQ Curve for transport encryption

### LINKED ISSUE #808: Support encryption for the kernel protocol over ZMQ
Kernel messages are currently [signed](https://github.com/jupyter/jupyter_client/blob/7d05cfb640a05bd80abc11db815df04b8c3ce25c/jupyter_client/session.py#L744-L745), preventing e.g. from running code that doesn't originate from a trusted source, but the code of an execution request is still sent unencrypted over the wire. Even if it's bad practice, passwords or sensitive information could be intercepted.
Would it make sense to support encryption for the kernel protocol over ZMQ? I'm thinking about remote kernels that wouldn't run on the same machine as the Jupyter server.

krassowski:
This PR adds CurveZMQ transport encryption to jupyter_client, providing the foundational layer for encrypted kernel communication. It is the upstream dependency for ipykernel#1515 and jupyter-server#1638.

What's changed

- local_provisioner.py: Generates a CurveZMQ keypair and writes curve_publickey / curve_secretkey into the connection file. The client owns key generation.
- connect.py: Reads curve keys from connection files and applies them when setting up ZMQ sockets.
- manager.py: Introduces a transport_encryption traitlet with three modes:

  - 'disabled' (default) — no encryption, existing behavior unchanged
  - 'auto' — enables encryption only for kernelspecs that advertise metadata.supported_encryption: 'curve'
  - 'required' — encryption is mandatory; kernel startup fails if the kernelspec doesn't declare support
- channels.py / client.py: Heartbeat channel configured as a CurveZMQ client using the server's public key.
- Validation: Checks for libsodium support in pyzmq at startup and raises a clear error if missing.

Design notes

- Named `transport_encryption` (not a boolean) to leave room for future encryption backends beyond CurveZMQ.
- Fully backward-compatible: 'disabled' is the default, so nothing changes for existing kernels.
- Kernel capability discovery is via kernelspec metadata (supported_encryption: 'curve'), handled on the kernel side in ipykernel#1515.
- Guards against pyzmq builds without libsodium, with graceful degradation or a clear error depending on the configured mode.

### References

- Closes https://github.com/jupyter/jupyter_client/issues/808
- PoC for https://github.com/jupyter/enhancement-proposals/issues/75
- Sibling PR https://github.com/ipython/ipykernel/pull/1515
- Toggle on server-level: https://github.com/jupyter-server/jupyter_server/pull/1638

### Code changes

Follows ipyparallel approach (https://github.com/ipython/ipyparallel/pull/553)

- [x] Tests
- [x] Add `curve_serverkey` handling to heartbeat channel
- [x] Pass curve keys from connection file to the socket (implemented at mixing level)

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -1,0 +1,162 @@
+.. _security:
+
+================================
+Transport security for kernels
+================================
+
+.. versionadded:: 8.9
+
+By default, the ZMQ sockets used to communicate with kernels (shell, IOPub,
+stdin, control, heartbeat) are bound to local TCP ports with no
+transport-level encryption. Any process on the same host that can reach
+those ports can connect and read messages, including all IOPub output.
+
+`CurveZMQ <https://rfc.zeromq.org/spec/26/>`_ adds elliptic-curve
+encryption and authentication at the ZMQ transport layer. When enabled, the
+``KernelManager`` generates a keypair, writes it into the kernel's connection
+file, and configures all sockets as a CurveZMQ server. Clients must present
+the correct server public key to connect; unauthenticated connections are
+silently dropped before any data is delivered.
+
+.. note::
+
+    CurveZMQ is only available when pyzmq was compiled against a libzmq that
+    includes libsodium. You can verify this with::
+
+        python -c "import zmq; print(zmq.has('curve'))"
+
+    If this prints ``False``, the ``transport_encryption`` setting has no
+    effect and attempts to set it to ``'auto'`` or ``'required'`` will raise
+    a :exc:`traitlets.TraitError`.
+
+.. note::
+
+    ``transport_encryption`` applies to TCP transport only. IPC sockets
+    already rely on filesystem permissions for access control and do not
+    support CurveZMQ.
+
+
+The ``transport_encryption`` setting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Transport encryption is controlled by the
+``KernelManager.transport_encryption`` traitlet, which accepts three values:
+
+``'disabled'`` (default)
+    No CurveZMQ keys are generated. All kernel sockets are unencrypted.
+
+``'auto'``
+    Keys are generated **only when the kernelspec declares support** via
+    ``metadata.supported_encryption: 'curve'``. Kernelspecs that do not
+    declare this field are started without encryption, so the setting is safe
+    to enable globally without breaking existing kernels.
+
+``'required'``
+    Keys are always generated. Startup fails with a :exc:`RuntimeError` if
+    the kernelspec does not declare ``metadata.supported_encryption: 'curve'``,
+    so kernels that have not been updated to handle the connection-file keys
+    are never started unencrypted.
+
+To enable encryption for all kernels that support it, add the following to
+your configuration:
+
+.. code:: python
+
+    c.KernelManager.transport_encryption = "auto"
+
+To enforce encryption and refuse to start kernels that do not declare support:
+
+.. code:: python
+
+    c.KernelManager.transport_encryption = "required"
+
+
+Kernelspec requirements
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A kernel must declare CurveZMQ support in its ``kernel.json`` before the
+``KernelManager`` will provision keys for it:
+
+.. code:: JSON
+
+    {
+        "argv": ["python", "-m", "ipykernel_launcher", "-f", "{connection_file}"],
+        "display_name": "Python 3",
+        "language": "python",
+        "metadata": {
+            "supported_encryption": "curve"
+        }
+    }
+
+When ``transport_encryption`` is ``'auto'``, kernelspecs without this field
+are started normally without encryption. When it is ``'required'``, their
+startup is refused.
+
+.. note::
+
+    When updating a previously installed kernel to a version that supports
+    encryption you may need to re-install the kernelspec or manually add the
+    ``supported_encryption`` metadata field. If you subsequently decide to
+    downgrade, you will need to remove this field as otherwise the kernel will
+    silently fail to connect.
+
+
+Connection file fields
+~~~~~~~~~~~~~~~~~~~~~~
+
+When ``transport_encryption`` is active the connection file written to disk
+will contain two additional fields alongside the usual port and key fields:
+
+``curve_publickey``
+    Z85-encoded 40-character ASCII string holding the server's CurveZMQ
+    public key.
+
+``curve_secretkey``
+    Z85-encoded 40-character ASCII string holding the server's CurveZMQ
+    secret key.
+
+Kernel implementations must read these fields from the connection file and
+apply them to their ZMQ sockets before binding. See :ref:`kernels` for the
+full description of the connection file format.
+
+
+Implementing curve support in a kernel
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A kernel that wants to be compatible with ``transport_encryption`` must apply
+the keypair to every socket it binds. In Python, using pyzmq, that looks like:
+
+.. code:: python
+
+    import json
+    import zmq
+
+    with open(connection_file_path) as f:
+        cfg = json.load(f)
+
+    ctx = zmq.Context()
+    shell = ctx.socket(zmq.ROUTER)
+
+    if "curve_publickey" in cfg and "curve_secretkey" in cfg:
+        shell.curve_secretkey = cfg["curve_secretkey"].encode()
+        shell.curve_publickey = cfg["curve_publickey"].encode()
+        shell.curve_server = True
+
+    shell.bind(f"tcp://{cfg['ip']}:{cfg['shell_port']}")
+
+The same pattern applies to the IOPub, stdin, control, and heartbeat sockets.
+Setting ``curve_server = True`` on a bound socket causes ZMQ to require
+CurveZMQ authentication on every incoming connection; unauthenticated clients
+are rejected automatically.
+
+Kernels based on `ipykernel <https://ipykernel.readthedocs.io/>`_ v7.3+ handle
+this automatically when the connection file contains the curve fields.
+
+.. seealso::
+
+   :ref:`kernels`
+     Connection file format and kernelspec reference.
+
+   :ref:`provisioning`
+     How ``LocalProvisioner`` generates and injects curve keys during
+     kernel startup.
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -69,7 +69,7 @@ class HBChannel(Thread):
         address : zmq url
             Standard (ip, port) tuple that the kernel is listening on.
         curve_serverkey : bytes, optional
-            CurveZMQ server public key (Z85).  When provided, the
+            CurveZMQ server public key (Z85). When provided, the
             heartbeat REQ socket is configured as a CurveZMQ client so it
             can communicate with a CurveZMQ-enabled kernel.
         """
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -55,6 +55,8 @@ class HBChannel(Thread):
         context: zmq.Context | None = None,
         session: Session | None = None,
         address: t.Union[t.Tuple[str, int], str] = "",
+        *,
+        curve_serverkey: bytes | None = None,
     ) -> None:
         """Create the heartbeat monitor thread.
 
@@ -66,12 +68,17 @@ class HBChannel(Thread):
             The session to use.
         address : zmq url
             Standard (ip, port) tuple that the kernel is listening on.
+        curve_serverkey : bytes, optional
+            Z85-encoded CurveZMQ server public key.  When provided, the
+            heartbeat REQ socket is configured as a CurveZMQ client so it
+            can communicate with a CurveZMQ-enabled kernel.
         """
         super().__init__()
         self.daemon = True
 
         self.context = context
         self.session = session
+        self.curve_serverkey = curve_serverkey
         if isinstance(address, tuple):
             if address[1] == 0:
                 message = "The port number for a channel cannot be 0."
@@ -104,6 +111,13 @@ class HBChannel(Thread):
         assert self.context is not None
         self.socket = self.context.socket(zmq.REQ)
         self.socket.linger = 1000
+        if self.curve_serverkey is not None:
+            # Generate a fresh ephemeral keypair for each socket; only the
+            # server public key (curve_serverkey) is needed for authentication.
+            client_pub, client_sec = zmq.curve_keypair()
+            self.socket.curve_secretkey = client_sec
+            self.socket.curve_publickey = client_pub
+            self.socket.curve_serverkey = self.curve_serverkey
         assert self.address is not None
         self.socket.connect(self.address)
 
--- a/jupyter_client/channels.py
+++ b/jupyter_client/channels.py
@@ -69,7 +69,7 @@ class HBChannel(Thread):
         address : zmq url
             Standard (ip, port) tuple that the kernel is listening on.
         curve_serverkey : bytes, optional
-            Z85-encoded CurveZMQ server public key.  When provided, the
+            CurveZMQ server public key (Z85).  When provided, the
             heartbeat REQ socket is configured as a CurveZMQ client so it
             can communicate with a CurveZMQ-enabled kernel.
         """
--- a/jupyter_client/client.py
+++ b/jupyter_client/client.py
@@ -393,8 +393,8 @@ class KernelClient(ConnectionFileMixin):
             url = self._make_url("hb")
             self.log.debug("connecting heartbeat channel to %s", url)
             hb_kwargs = {}
-            if self._curve_publickey:
-                hb_kwargs["curve_serverkey"] = self._curve_publickey
+            if self.curve_publickey:
+                hb_kwargs["curve_serverkey"] = self.curve_publickey
             try:
                 self._hb_channel = self.hb_channel_class(  # type:ignore[call-arg,abstract]
                     self.context,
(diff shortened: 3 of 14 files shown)


jupyter_client/client.py
```diff
@@ -393,7 +393,10 @@ def hb_channel(self) -> t.Any:
             url = self._make_url("hb")
             self.log.debug("connecting heartbeat channel to %s", url)
             self._hb_channel = self.hb_channel_class(  # type:ignore[call-arg,abstract]
-                self.context, self.session, url
+                self.context,
+                self.session,
+                url,
+                curve_serverkey=self._curve_publickey,
```
Carreau: Should we pass it conditionally, or using something like https://pypi.org/project/backcall/ , because we _risk_ passing an extra parameter to a custom class (I'm fine saying it is mandatory or will become mandatory), but we may want to be careful here.

jupyter_client/connect.py
```diff
@@ -565,6 +570,13 @@ def load_connection_info(self, info: KernelConnectionInfo) -> None:
             self.session.key = key
         if "signature_scheme" in info:
             self.session.signature_scheme = info["signature_scheme"]
+        if "curve_publickey" in info and "curve_secretkey" in info:
+            pub = info["curve_publickey"]
+            sec = info["curve_secretkey"]
+            assert isinstance(pub, (str, bytes))
+            assert isinstance(sec, (str, bytes))
+            self._curve_publickey = pub.encode() if isinstance(pub, str) else pub
+            self._curve_secretkey = sec.encode() if isinstance(sec, str) else sec
```
Carreau: I don't like the `str|bytes` especially if we may have some z85 around, the encode/decode become ambiguous.

Carreau (commented):
Most failing test seem to be an issue with history DB on IPython, i'm not sure they are related.

Done only a cursory review, I need to wrap my head around everything.

jupyter_client/channels.py
```diff
@@ -66,12 +68,17 @@ def __init__(
             The session to use.
         address : zmq url
             Standard (ip, port) tuple that the kernel is listening on.
+        curve_serverkey : bytes, optional
+            Z85-encoded CurveZMQ server public key.  When provided, the
+            heartbeat REQ socket is configured as a CurveZMQ client so it
+            can communicate with a CurveZMQ-enabled kernel.
```
Carreau, line 74: I don't think we should mention it's Z85 encoded first, maybe in notes, because I got confused if the user/dev need to handled Z85 themseleves. I think it should be treated as raw opaque bytes that _happen_ to be safely .encoded() and decoded() to ascci/utf8.

jupyter_client/connect.py
Carreau, line 574: We may want

```
from typing_extensions import TypedDict

class KernelConnectionInfo(TypedDict, extra_items=str | bytes | int):
    curve_publickey: bytes # or str, I don't care, but likely stay consistent.
    curve_secretkey: bytes
```

right now it's 

```
KernelConnectionInfo = dict[str, Union[int, str, bytes]]
```

This should be backward compatible and keep us with strong typing of fields

tests/test_transport_security.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+@pytest.mark.parametrize(
+    ("enable_curve", "expect_plaintext_visible"),
+    [
+        (False, True),
+        (True, False),
+    ],
+)
+def test_iopub_plaintext_visibility_depends_on_curve(enable_curve, expect_plaintext_visible):
```
Carreau: so `expect_plaintext_visible = not enable_curve` ? Maybe drop one of the params ?
Unless you want to test the cartesian product but I don't think this is what you are doing here.

jupyter_client/client.py
krassowski: We can sniff (I think sniffing became easier in recent Python versions), not sure if it is worth pulling `backcall` out of the grave - maybe, maybe not.

tests/test_transport_security.py
```diff
@@ -0,0 +1,161 @@
...
+                "connect_shell() did not produce a working Curve-authenticated socket"
+            )
+        finally:
+            client_sock.close(linger=0)
+    finally:
+        server.close(linger=0)
+        ctx.term()
+
+
+def test_connect_shell_to_curve_server_without_curve_keys_is_rejected():
```
Carreau, line 324: should we add a test with a mutated key ? And not just absence ?

Carreau:
+1 in general, this is a relatively small patch I was expecting more.

Carreau:
314t failures are due to https://github.com/msgpack/msgpack-python/issues/613, msgpack is not free-threaded compatible yet, I'll send a PR that remove 314t testing.

jupyter_client/client.py
krassowski: Thinking more about it, we should not sniff on parameters but just avoid passing the argument if `self._curve_publickey` is `None`, because if someone enables curve, but heartbeat class does not support they would end up in a broken state. So maybe we should just explicitly raise here if `self._curve_publickey is not None` and sniffing points to no support for `curve_serverkey`.

Carreau: Yeah; i agree. Sniffing signature also just to say that it _may_ become a mandatory parameter later

jupyter_client/connect.py
krassowski: Not trivial to achieve
- `json` supports strings only
- we serialize/deserialize to json on write/read to connection file
- there is precedent on this with `key` which is expected to come in as bytes and come out as string for serialization
- we need to decode the string back to bytes on reads as in:
https://github.com/jupyter/jupyter_client/blob/d2a7b2b168f0f9fc517a07703f3a41ef3e61c9cb/jupyter_client/connect.py#L601-L603

Currently `KernelConnectionInfo` represents both forms: serialization-ready configuration (`key is str`), and the runtime values (`key is bytes`).

krassowski: I suppose it is reasonable to split this into two TypedDicts?

Carreau:
(rebased on main to show less failures)

jupyter_client/connect.py
krassowski: Maybe we don't need two.

jupyter_client/channels.py
krassowski: Reworded in 42420c6

jupyter_client/connect.py
krassowski: Resolved by f73224d

jupyter_client/connect.py
krassowski: Done in f73224d

tests/test_transport_security.py
krassowski: Done in fb28ce5

jupyter_client/client.py
krassowski: See 683aa8e let me know if you think we should do more.

tests/test_transport_security.py
krassowski: :+1: 81a2daa

jupyter_client/connect.py
krassowski: Ah, mypy does not support `extra_items` yet (https://github.com/python/mypy/issues/18176). I will do second best things (`total=False`)

Carreau: If types are too complicated, I think we can fallback to assert and just make the codebase strict at runtime, don't spend an eternity on it.

jupyter_client/client.py
```diff
@@ -392,8 +392,21 @@ def hb_channel(self) -> t.Any:
...
+            hb_supports_curve = (
+                "curve_serverkey" in inspect.signature(self.hb_channel_class.__init__).parameters
+            )
+            if self._curve_publickey is not None and not hb_supports_curve:
+                msg = (
+                    f"{self.hb_channel_class.__name__} does not support the "
+                    "'curve_serverkey' parameter. Upgrade the heartbeat channel "
+                    "class or disable CurveZMQ encryption."
+                )
+                raise RuntimeError(msg)
```
Carreau: :+1:

jupyter_client/client.py
```diff
@@ -392,8 +392,21 @@ def hb_channel(self) -> t.Any:
...
             self._hb_channel = self.hb_channel_class(  # type:ignore[call-arg,abstract]
-                self.context, self.session, url
+                self.context,
+                self.session,
+                url,
+                **({"curve_serverkey": self._curve_publickey} if hb_supports_curve else {}),
```
Carreau: That triggers me to want to resume working on a pep for `undefined`/`void` parameter that is striped when calling function.

jupyter_client/client.py
minrk: I don't think we need to inspect for `hb_supports_curve`, since we can pass the argument if it's required and let the standard unsupported argument error raise:

```python
hb_kwargs = {}
if self._curve_publickey:
    hb_kwargs["curve_serverkey"] = self._curve_publickey
...
hb_channel_class(...**hb_kwargs)
```

but fine if you want to keep the more detailed error. But if you hit that error, a lot of other things are not going to work before we get to the hb channel, I think.

Carreau: I'm fine with both; as long as we know there are some backward compatibility story here.

krassowski: Does 8b5d6d68dfc1f9bc84c772b28ed7526206e6b554 work?

jupyter_client/provisioning/local_provisioner.py
```diff
@@ -174,6 +176,11 @@ async def pre_launch(self, **kwargs: Any) -> dict[str, Any]:
         # This should be considered temporary until a better division of labor can be defined.
         km = self.parent
         if km:
+            transport_encryption = bool(
+                kwargs.pop("transport_encryption", getattr(km, "transport_encryption", False))
+            )
```
Carreau: Thinking for myself, 

It that bool for type annotation to not fail ? 
Should be more conservative and check that it actually is a bool (or None), we get back, and not any other non-falsy vallue ? Or are we thinking the ks.transport_encryption could one day become Enums of different type of encryption ?

jupyter_client/provisioning/local_provisioner.py
```diff
@@ -196,11 +203,23 @@ async def pre_launch(self, **kwargs: Any) -> dict[str, Any]:
...
+                    jupyter_session=jupyter_session,
+                    curve_publickey=curve_publickey,
+                    curve_secretkey=curve_secretkey,
+                )
             else:
-                km.write_connection_file()
+                km.write_connection_file(
+                    curve_publickey=curve_publickey,
+                    curve_secretkey=curve_secretkey,
+                )
```
Carreau: Seem like the provisioner is looking at an attribute on km to set other attribute on KM... I don't have a better way of doing this, but feel like a code smell and that all think should be  in KM itself. Not a request for change; but a note for potential future refactor.

jupyter_client/client.py
```diff
@@ -392,9 +392,26 @@ def hb_channel(self) -> t.Any:
...
+            except TypeError as e:
+                if "curve_serverkey" in str(e):
+                    msg = (
+                        f"{self.hb_channel_class.__name__} does not support the "
+                        "'curve_serverkey' parameter. Upgrade the heartbeat channel "
+                        "class or disable CurveZMQ encryption."
+                    )
+                    raise RuntimeError(msg) from e
+                else:
+                    raise
```
Carreau, line 414: +1 to all this.

jupyter_client/connect.py
```diff
@@ -33,7 +34,22 @@
...
+    control_port: int
+    hb_port: int
+    ip: str
+    key: str
+    transport: str
+    signature_scheme: str
+    kernel_name: str
+    session: Session
+    curve_publickey: str
+    curve_secretkey: str
```
Carreau, line 52: +1 awesome it was more complicated than I thought, but I think it will help is in the long term.

jupyter_client/connect.py
```diff
@@ -657,6 +697,14 @@ def _create_connected_socket(
...
         if identity:
             sock.identity = identity
+        if self.curve_publickey is not None:
+            # The connection file already carries this keypair, so reusing it
+            # avoids introducing an additional key-distribution mechanism here.
+            # curve_serverkey authenticates the server; the keypair configures
+            # encrypted communication for the client socket.
+            sock.curve_secretkey = self.curve_secretkey
+            sock.curve_publickey = self.curve_publickey
+            sock.curve_serverkey = self.curve_publickey
```
Carreau, line 707: public_key twice ?

Obviously you spent more time then me we actually writing the code, so genuine question.

Does the secret/public/server and manipulating each argument separately become prone with error ? And actually thinking at loud; outside of the connection file; would it make more sens to manipulate a `keypair: (bytes, bytes)|None` everywhere ? and split it only when we actually need each components ?

jupyter_client/manager.py
```diff
@@ -393,6 +403,9 @@ async def _async_pre_start_kernel(
              and launching the kernel (e.g. Popen kwargs).
         """
         self.shutting_down = False
+        # Consume manager-level curve toggle so it does not leak into subprocess kwargs.
+        # Provisioners can read this from the manager instance.
+        self.transport_encryption = bool(kw.pop("transport_encryption", self.transport_encryption))
```
Carreau: here I think I woudl also fail on non `bool|None kwargs`.

I'm also wondering if 

```
 async def _async_pre_start_kernel(
        self, *, transport_encryption:bool|None=None, **kw: t.Any
    ) -> t.Tuple[t.List[str], t.Dict[str, t.Any]]:
```

Might be worth it, but I see how ot can be easier to deal with the kwarg not existing than having to deal with None or a sentinel.

Carreau:
I can help trying and writing some of the suggestion if you like to, I'm just doing a quick review on Sunday between two other things.

jupyter_client/provisioning/local_provisioner.py
krassowski: > Or are we thinking the ks.transport_encryption could one day become Enums of different type of encryption

Yes, this is exactly why I called it `transport_encryption` (rather than `enable_transport_encryption` or `transport_encryption_on`).

> Should be more conservative and check that it actually is a bool (or None), we get back, and not any other non-falsy vallue

Maybe for now? I do not have a strong opinion but happy to make this change.

jupyter_client/provisioning/local_provisioner.py
krassowski: What I am not sure of, is how other provisioners would generate keys. Would the responsibility always lie with the `KernelManager`, or is it ever on the provisioner? Since the provisioner write the connection file I lean towards being ambiguous here hence this pattern makes sense to me.

jupyter_client/connect.py
krassowski: Yes, server key is the public key as per Min's example in https://github.com/jupyter/enhancement-proposals/issues/75. 

> keypair

A tuple for key pair would be dangerous (one could unpack in wrong order), but having a dataclass to pass keys around could be tempting.

krassowski: Thinking more about this, collapsing it to a dataclass does not seem worth it. We are bound on either side by traitlets input format or zmq socket format.

jupyter_client/manager.py
krassowski: It might be worth it if the signature was preserved for typing purposes.

krassowski: 242c6b77d7ba5a4518c380b283757e373a651008

jupyter_client/connect.py
Carreau: Ok, no worries, You coudl also have a pure static new type and have a

SecretKey = NewType('SecretKey', bytes)
PublicKey = NewType('PuvlicKey', bytes)
KeyPair = tuple[SecretKey, PublicKey]


But it does end up being heavy indeed.

minrk (approved):
I think this implementation is solid. I think the missing piece in this repo is discovery of whether the kernel supports curve, which I suppose is a KernelSpec feature?

krassowski:
I will push a commit to avoid failing if using zmq build without libsodium, although I think this is unlikely to happen for supported installation methods - I think conda-forge and PyPI ships with libsodium by default. I see https://github.com/conda-forge/pyzmq-feedstock/commit/f492894e66064b02713e78305937b1923b4ee8be removed `libsodium` from `host` requirements but I suppose this gets pulled via zeromq https://github.com/conda-forge/zeromq-feedstock/blob/69c6d89e75cab3a1b3f5c325e2709c4daa49cbdf/recipe/recipe.yaml#L62

krassowski:
> I will push a commit to avoid failing if using zmq build without libsodium

Well failing would be ok, because `zmq.curve_keypair()` is only called when transport encryption is `enabled` or `required`; however we should fail as soon as possible, which is earlier than the first kernel startup as this is not a user error but operator error if they custom built `zeromq` without `libsodium` support. So instead, I will add that as a validator.

krassowski:
Can we merge this one?

(ipykernel failure was addressed downstream in https://github.com/ipython/ipykernel/pull/1524 but not yet released)

docs/security.rst
```diff
@@ -0,0 +1,162 @@
...
+are started normally without encryption. When it is ``'required'``, their
+startup is refused.
+
+.. note::
+
+    When updating a previously installed kernel to a version that supports
+    encryption you may need to re-install the kernelspec or manually add the
+    ``supported_encryption`` metadata field. If you subsequently decide to
+    downgrade, you will need to remove this field as otherwise the kernel will
+    silently fail to connect.
```
Carreau, line 101: :+1:

Carreau (approved):

fleming79:
I was trying this out and think there might be a problem when restarting a kernel with encryption enabled. 

The associated log looks like this:

```console
[IPKernelApp] WARNING | Kernel is running over TCP without encryption. All communication (including code and outputs) is sent in plain text and is susceptible to eavesdropping. Use IPC transport or launch with kernel manager-provisioned CurveZMQ keys to enable transport encryption.
[W 2026-06-08 07:58:05.688 ServerApp] Nudge: attempt 10 on kernel 690e733c-693e-4976-b798-5e0f63
```

krassowski:
^ https://github.com/jupyter/jupyter_client/pull/1124 will fix that

Issue #1111, opened 2026-05-05
Docs build is failing on CI


krassowski:
It appears to have the same underlying issue as https://github.com/jupyter-server/jupyter_server/issues/1595

krassowski:
I suppose we need to release https://github.com/ipython/traitlets/pull/935

krassowski:
Closing as fixed with https://github.com/ipython/traitlets/releases/tag/v5.15.0 release

PR #1112, opened 2026-05-05
Log environment information for docs build on CI, update comment on workaround 

krassowski:
To help with issues like #1111


Carreau (approved):

Carreau:
Force merging as anyway test will fail.

PR #1113, opened 2026-05-06
Remove `3.14t` testing until `msgpack` supports it

Carreau:
Right now this is just CI noise.

See https://github.com/msgpack/msgpack-python/issues/613

I'll open a tracking issue to reenable it.


krassowski (approved):
Thanks!

PR #1116, opened 2026-05-06
CI: fix testing of nbconvert

Carreau:
https://github.com/jupyterlab/maintainer-tools/pull/275 introduce extra quoting of package-spec so `-e` does not worl anymore; fallback to non-editabel install


krassowski (approved):
Thank you!

krassowski:
(I cannot merge here since the required check is failing)

PR #1123, opened 2026-06-08
Fix kernel restart for curve encryption.

fleming79:
Previously if a kernel was started with curve encryption it could not be restarted because the encryption settings were wrong for the restarted kernel.

This PR ensures the curve details are written to the connection file and also avoids changing settings on the kernel if the connection file still exists.

### DIFF
--- a/jupyter_client/provisioning/local_provisioner.py
+++ b/jupyter_client/provisioning/local_provisioner.py
@@ -16,6 +16,9 @@ from ..launcher import launch_kernel
 from ..localinterfaces import is_local_ip, local_ips
 from .provisioner_base import KernelProvisionerBase
 
+if TYPE_CHECKING:
+    from jupyter_client.manager import KernelManager
+
 
 class LocalProvisioner(KernelProvisionerBase):
     """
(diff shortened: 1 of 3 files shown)


jupyter_client/provisioning/local_provisioner.py
```diff
@@ -174,75 +177,64 @@ async def pre_launch(self, **kwargs: Any) -> dict[str, Any]:
...
-                    curve_publickey, curve_secretkey = zmq.curve_keypair()
-                    km.curve_publickey = curve_publickey
-                    km.curve_secretkey = curve_secretkey
-            if "env" in kwargs:
-                jupyter_session = kwargs["env"].get("JPY_SESSION_NAME", "")
-                km.write_connection_file(
-                    jupyter_session=jupyter_session,
-                    curve_publickey=curve_publickey,
-                    curve_secretkey=curve_secretkey,
+            if not os.path.exists(km.connection_file):
```
krassowski: I am worried that this is a big change in behaviour impacting more than just encryption; the change proposed in https://github.com/jupyter/jupyter_client/pull/1124 is contained to reusing encryption only and still writing other things that changed.

fleming79: `km.write_connection_file` only writes the connection file if it [doesn't already exist](https://github.com/jupyter/jupyter_client/blob/94d70709a4e8de94835aa8e702bf3e6fc995cb10/jupyter_client/connect.py#L534).

We could delete the connection file instead, but I think it might be safer to leave it as is.

PR #1124, opened 2026-06-08
Fix restart with curve encryption enabled

krassowski:
- Addresses https://github.com/jupyter/jupyter_client/pull/1110#issuecomment-4644227584
- Alternative to https://github.com/jupyter/jupyter_client/pull/1123

What do you think @fleming79?

### DIFF
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -546,6 +546,8 @@ class ConnectionFileMixin(LoggingConfigurable):
             control_port=self.control_port,
             signature_scheme=self.session.signature_scheme,
             kernel_name=self.kernel_name,
+            curve_publickey=self.curve_publickey,
+            curve_secretkey=self.curve_secretkey,
             **kwargs,
         )
         # write_connection_file also sets default ports:
--- a/jupyter_client/provisioning/local_provisioner.py
+++ b/jupyter_client/provisioning/local_provisioner.py
@@ -220,9 +220,16 @@ class LocalProvisioner(KernelProvisionerBase):
                     and km._kernel_supports_curve_encryption()
                 )
                 if kernel_curve_ok:
-                    curve_publickey, curve_secretkey = zmq.curve_keypair()
-                    km.curve_publickey = curve_publickey
-                    km.curve_secretkey = curve_secretkey
+                    if km.curve_publickey is None:
+                        curve_publickey, curve_secretkey = zmq.curve_keypair()
+                        km.curve_publickey = curve_publickey
+                        km.curve_secretkey = curve_secretkey
+                    else:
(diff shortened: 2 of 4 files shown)


fleming79:
My PR also changes ConnectionFileMixin to write `curve_publickey`, `curve_secretkey` instead of having them passed as kwargs which I think should also be fixed.

krassowski:
> My PR also changes ConnectionFileMixin to write `curve_publickey`, `curve_secretkey` instead of having them passed as kwargs which I think should also be fixed.

I have no objection as in this seems more in line with the existing pattern, but I am curious if this is motivated by any issue or just aligning the code patterns.

Do you think we should merge this PR over https://github.com/jupyter/jupyter_client/pull/1123 and extract the kwargs change from https://github.com/jupyter/jupyter_client/pull/1123 PR (or change it to only address that), or do you think the proposal in https://github.com/jupyter/jupyter_client/pull/1123 is better for another reason?

fleming79:
> I have no objection as in this seems more in line with the existing pattern, but I am curious if this is motivated by any issue or just aligning the code patterns.

It is an issue if you want to call the method `write_connection_file` and have those settings included automatically (I was using it when writing some tests). 

### Example
```python

    client = AsyncKernelClient(
        connection_file=str(tmpdir.joinpath(f"kernel-{os.getpid()}.json")),
        curve_publickey=curve_publickey,
        curve_secretkey=curve_secretkey,
        transport=transport,
        kernel_name=name,
    )
    client.write_connection_file()
    client.start_channels()
```


> Do you think we should merge this PR over #1123 and extract the kwargs change from #1123 PR (or change it to only address that), or do you think the proposal in #1123 is better for another reason?

I like #1123 because it respects a pre-existing connection file. `write_connection_file` is a bit of a gotcha, which does nothing if the file is already written by the kernel manager (such as when restarting the kernel), so could end up with other sync issues between the manager and kernel, not just the curve encryption settings.

I don't mind if you want to merge the PRs.

Edit: Please see #1125

krassowski:
> It is an issue if you want to call the method write_connection_file and have those settings included automatically (I was using it when writing some tests).

Right. I think we should merge that fix then. I can push it as a commit in this PR and will add you as a co-author.

> so could end up with other sync issues between the manager and kernel, not just the curve encryption settings.

Acknowledging this. I mentioned this in https://github.com/jupyter/jupyter_client/pull/1125#issuecomment-4658141513 but I will call out that this could be addressed in a separate PR narrowly addressing that issue alone.

Carreau (approved):
+1, @krassowski do you want to merge and do a patch release immediately ? Otherwise I can try to do in the next few days, and yank the previous release.

PR #1125, opened 2026-06-08
Fix ConnectionFileMixin.write_connection_file and restart for curve encryption.

fleming79:
`ConnectionFileMixin.write_connection_file` will now update an existing file with the current settings. The method can be called multiple times with extra settings to add. Previously the behavior was ambiguous, possibly with nothing written, possibly resulting in settings mismatch. This became more apparent with the recent addition of curve encryption support.

Replaces #1123

This also adds `__del__` to delete the connection file if it was written by the ConnectionFileMixin.

### DIFF
--- a/jupyter_client/client.py
+++ b/jupyter_client/client.py
@@ -128,12 +128,7 @@ class KernelClient(ConnectionFileMixin):
                 if self.log:
                     self.log.debug("Destroying zmq context for %s", self)
                 self.context.destroy(linger=100)
-        try:
-            super_del = super().__del__  # type:ignore[misc]
-        except AttributeError:
-            pass
-        else:
-            super_del()
+        super().__del__()
 
     # --------------------------------------------------------------------------
     # Channel proxy methods
--- a/jupyter_client/connect.py
+++ b/jupyter_client/connect.py
@@ -542,6 +542,8 @@ class ConnectionFileMixin(LoggingConfigurable):
     def write_connection_file(self, **kwargs: Any) -> None:
         """Write connection info to JSON dict in self.connection_file."""
         cfg_: Any = self.get_connection_info()
+        assert not set(cfg_).intersection(kwargs), "Overwriting connection info is not allowed!"
+
         if os.path.exists(self.connection_file):
             with contextlib.suppress(Exception), open(self.connection_file, "rb") as f:
                 cfg_ = json.load(f) | cfg_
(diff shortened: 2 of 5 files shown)


fleming79:
@krassowski - I think this address the concern you had [here](https://github.com/jupyter/jupyter_client/pull/1124#issuecomment-4648190671).

comment: It'd be good to also clean up ipc files if the connection file was written too, maybe a separate pr?

krassowski:
> `ConnectionFileMixin.write_connection_file` will now update an existing file with the current settings. The method can be called multiple times with extra settings to add. Previously the behavior was ambiguous, possibly with nothing written, possibly resulting in settings mismatch.

To clarify, previously `write_connection_file` and `load_connection_file` of `ConnectionFileMixin` were called multiple times by two entities (kernel and client) to exchange information; for each initialization the write happened only once. It was a bit annoying because this was implicit and only documented outside of the code.

This PR proposes to make `write_connection_file` also include a separate loading mechanism and enforce that state does reconcile, allowing multiple writes. I think this makes reasoning about the information flow even harder, and makes typing looser (e.g. the required change to accept `str` in addition to `bytes` for keys now).

For now, I think we should keep the flow as simple as possible and minimize changes. I am _not_ strongly opposed to the idea of the proposed change, I just think this should be a separate PR from one that fixes the curve thing and possible released in a separate release because I am worried that a change in behaviour might break downstreams (we only test a subset) and I won't have resources to fix things up.

I think there were some good ideas on tidying up things in https://github.com/jupyter/jupyter_client/pull/1123 with respect to when `curve_publickey` is passed (as we discussed in https://github.com/jupyter/jupyter_client/pull/1124#issuecomment-4648583413). Again, that could be a separate PR.

fleming79:
> To clarify, previously `write_connection_file` and `load_connection_file` of `ConnectionFileMixin` were called multiple times by two entities (kernel and client) to exchange information; for each initialization the write happened only once. It was a bit annoying because this was implicit and only documented outside of the code.

I had no idea about this. Is this still required? Or can it be deprecated? I think it would be simplest if the connection file was only written once and attempting to overwrite it should cause an error/deprecation warning.

 

> This PR proposes to make `write_connection_file` also include a separate loading mechanism and enforce that state does reconcile, allowing multiple writes. I think this makes reasoning about the information flow even harder, and makes typing looser (e.g. the required change to accept `str` in addition to `bytes` for keys now).

The typing was simply due to json serialization, but point taken. The idea was thinking about subclassing of the provisioner. But I think it'd be best to have a write once policy if at all possible.

PR #1127, opened 2026-06-11
Align `supported_encryption` format with JEP in docs

krassowski:
Same as https://github.com/ipython/ipykernel/pull/1528 to align with the latest state in https://github.com/jupyter/enhancement-proposals/pull/145

### DIFF
--- a/jupyter_client/manager.py
+++ b/jupyter_client/manager.py
@@ -476,7 +476,7 @@ class KernelManager(ConnectionFileMixin):
         ):
             msg = (
                 "transport_encryption='required' but kernelspec does not declare "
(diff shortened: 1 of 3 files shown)


jupyter_client/manager.py
```diff
@@ -476,7 +476,7 @@ async def _async_pre_start_kernel(
         ):
             msg = (
                 "transport_encryption='required' but kernelspec does not declare "
-                "metadata.supported_encryption='curve'."
+                "'curve' in `metadata.supported_encryption`."
```
Carreau, line 479: +1

Carreau (approved):

PR #1128, opened 2026-06-14
support fixed ports from env when kernel using reverse proxy

956237586:
support fixed ports from env when kernel using reverse proxy

### DIFF
--- a/jupyter_client/_version.py
+++ b/jupyter_client/_version.py
@@ -3,7 +3,7 @@
 import re
 from typing import Union
 
-__version__ = "8.9.1"
+__version__ = "8.9.2"
 
 # Build up version_info tuple for backwards compatibility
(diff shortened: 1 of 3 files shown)


krassowski:
Hello :wave: is there an issue explaining what this PR is trying to achieve? This looks a bit short on description side, so it's hard to judge. A new test case would probably help too.

956237586:
> Hello 👋 is there an issue explaining what this PR is trying to achieve? This looks a bit short on description side, so it's hard to judge. A new test case would probably help too.

thanks for your reply
i am developing an [IDEA plugin](https://github.com/dev-assistant/skykoma-plugin-idea)
it will regist IDEA JVM as an custom jupyter kernel, receiving dynamic kotlin code snippets and executed in JVM process, 
when kernel created by jupyter, real process is already running (kotlin jupyter core embedded in JVM plugin process)
JVM process will running in pod/container, so i need know some FIXED port before kernel running to write port mapping config params or reverse proxy config

PR #1135, opened 2026-07-27
Add an example kernel provisioner (Tenki Sandbox microVMs)

fichie23:
## Add an example kernel provisioner (Tenki Sandbox microVMs)

This adds `examples/tenki_provisioner/` — a small but complete
`KernelProvisionerBase` implementation that launches each kernel inside a
disposable [Tenki Sandbox](https://tenki.cloud/docs/sandbox) Linux microVM
instead of on the local machine, and references it from `docs/provisioning.rst`
as a worked remote-provisioner example.

The docs currently describe provisioners with YARN/RBAC fragments; this provides
a runnable, tested end-to-end example of the pieces a real *remote* provisioner
needs (off-box launch, connection bridging, full process-control surface,
entry-point registration, kernelspec installer).

### How it works

A kernel exposes five ZeroMQ channels. Because the kernel runs in a microVM, the
provisioner drives it over the ZeroMQ **`ipc` transport**, so each channel is a
Unix-domain socket the kernel *binds* inside the guest. It stages two connection
files that are identical except for the socket-path prefix (client-side temp dir
vs. guest home), so the HMAC key and channel indices line up on both ends, then
bridges each local socket to the matching guest socket over the vendor SDK's
raw byte-stream primitive. No ssh, CLI, or inbound networking is required.

| `KernelProvisionerBase` method | Action |
| --- | --- |
| `pre_launch` | switch to `ipc`; stage local + guest connection files (restart reuses them) |
| `launch_kernel` | create microVM → install `ipykernel` → launch → bridge sockets (failure-atomic) |
| `poll` / `wait` | track the guest process handle |
| `send_signal` / `kill` / `terminate` | forward signals to the guest process |
| `cleanup` | close the bridge and terminate the microVM (retried; raised on failure) |

### What's included

- `tenki_provisioner/` — the provisioner, the socket bridge, and a
  `python -m tenki_provisioner.kernelspec install` helper
- `pyproject.toml` registering the `jupyter_client.kernel_provisioners`
  entry point, with a nested Ruff config that inherits the repo config
- unit tests (the remote backend is mocked) and a live end-to-end script
- a reference to the example from `docs/provisioning.rst`
- `examples/` added to the `interrogate` exclude list and a scoped CI workflow,
  since the example is not part of the shipped library

### Notes / scope

- The `ipc` transport uses Unix-domain sockets, so the machine running
  `jupyter_client` must be Linux/macOS (documented); the kernel is always Linux
  inside the microVM.
- Restart provisions a fresh microVM (guest state is not preserved). Kernel
  lifecycle is failure-atomic and has a configurable hard-duration backstop so a
  cancelled or crashed launch cannot leak a running VM.
- Cross-process reattach is intentionally not implemented.

### Testing

- Unit tests (SDK mocked): 11/11. `jupyter kernelspec provisioners` lists the
  provisioner.
- Live end-to-end against real infrastructure: a kernel starts in a microVM,
  executes code remotely (verified the hostname differs from the client and that
  `start_kernel(env=...)` values reach the guest), and the microVM is confirmed
  `TERMINATED` on shutdown.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -53,6 +53,9 @@ jobs:
       - uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: jupyter_server
+          # Skip test_check_version: newer `packaging` rejects the "1.0" literal
+          # used by jupyter_server's own test - unrelated to jupyter_client.
+          test_command: "pytest -vv -raXxs -W default --durations 10 --color=yes --deselect tests/test_utils.py::test_check_version"
 
   jupyter_kernel_test:
     runs-on: ubuntu-latest
--- a/.github/workflows/tenki-provisioner-example.yml
+++ b/.github/workflows/tenki-provisioner-example.yml
@@ -1,0 +1,45 @@
+name: tenki_provisioner example
+
+# The repository's main test workflow does not exercise examples/, so this job
+# lints, format-checks, and unit-tests the Tenki Sandbox provisioner example on
+# its own. The unit tests mock the SDK; the live end-to-end check is not run in
+# CI (it needs Tenki credentials and provisions a real microVM).
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - "examples/tenki_provisioner/**"
+      - ".github/workflows/tenki-provisioner-example.yml"
+  pull_request:
+    paths:
+      - "examples/tenki_provisioner/**"
+      - ".github/workflows/tenki-provisioner-example.yml"
+
+concurrency:
+  group: tenki-provisioner-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    defaults:
+      run:
+        working-directory: examples/tenki_provisioner
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install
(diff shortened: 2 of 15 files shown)


krassowski:
Hi @fichie23, thanks for proposing this addition, and sorry it took a while to review.

This looks like vendor promotion, which I am not against in principle, it just does not fit the current scope of the repo. There is no `examples/` directory here at all, and no example for any other provisioner, so this would create one for a single vendor.

What I would suggest instead is publishing the provisioner as its own package on PyPI, and then opening a PR that adds a list of third-party provisioners to the documentation, with yours as one entry. Such a list is useful for users and vendors alike, but it has to be fair and include the alternatives, and this PR does not do that as of now. There are around twenty packages registering the `jupyter_client.kernel_provisioners` entry point today and none of them are documented here. Some to start from: gateway_provisioners, pyproject-local-kernel, [sshpyk](https://github.com/casangi/sshpyk), pixi-kernel, jupyter-kernel-provisioner-proxy, jupyter_slurm_ssh_provisioner, [jupyter_libyt](https://github.com/yt-project/jupyter_libyt).

For entries in that list I would ask for: installable from PyPI, source publicly available under an OSI-approved license, and one line per entry with a link and a short neutral description rather than marketing copy.

The PR also carries three changes unrelated to the example: the mypy fix in `client.py`, the `test_check_version` deselect in the downstream workflow, and the `interrogate` exclude. The first is now covered by #1142 and the second is being handled separately, so nothing is needed from you there.

I am closing this one, but a PR adding the documentation list would be very welcome.

PR #1136, opened 2026-08-09
Preserve Curve key bytes in `KernelManager.client`

### LINKED ISSUE #1131: KernelManager.client() raises TraitError when Curve keys are provisioned
With `transport_encryption='auto'` (or `'required'`) and a kernelspec that declares `supported_encryption: "curve"`, `KernelManager.client()` fails:

```python
import asyncio
from jupyter_client import AsyncKernelManager

async def main():
    km = AsyncKernelManager(transport_encryption='auto')
    await km.start_kernel()  # ipykernel >= 7.3, so curve keys are provisioned
    kc = km.client()

asyncio.run(main())
```

```
traitlets.traitlets.TraitError: The 'curve_publickey' trait of an AsyncKernelClient instance
expected a bytes object, not the str 'cQ!>=@Bkp:b7v)*A4E+[p]bk#mr*IWq</nM1*GE?'
```

Cause: `get_connection_info()` returns the curve keys `.decode()`d to str (connect.py), but `client()` passes that dict straight to `client_factory(**kw)`, and the client's `curve_publickey`/`curve_secretkey` traits are `Bytes`. So manager-provisioned transport encryption can't be used with the manager's own `client()` method. Seen on 8.9.1 and current main.

Proposed fix: coerce in `client()` before instantiating, mirroring what `load_connection_info` already does for the reverse direction:

```python
for k in ('curve_publickey', 'curve_secretkey'):
    if isinstance(kw.get(k), str): kw[k] = kw[k].encode('ascii')
```

Alternatively, a `@validate` cross-validator on the two traits accepting str would fix all entry points at once.

aryansk:
Fixes #1131

`ConnectionFileMixin.get_connection_info()` exposes CurveZMQ keys as strings for connection metadata, while the client traits require bytes. Convert the two Curve key values back to ASCII bytes in `KernelManager.client()` before constructing the client, and add a regression test covering provisioned keys.


krassowski:
Hey any reason this is in draft?

krassowski (approved):
Thanks!

PR #1138, opened 2026-08-11
Fix kernel_dirs help text to match first-wins precedence

### LINKED ISSUE #1137: `KernelSpecManager.kernel_dirs` help text contradicts `find_kernel_specs()` precedence
## Summary

The `kernel_dirs` trait documents later-wins precedence, but `find_kernel_specs()` implements first-wins:

https://github.com/jupyter/jupyter_client/blob/8d760b1efeff2ac7aa2b4b75c44023e59c6945f8/jupyter_client/kernelspec.py#L166-L168

https://github.com/jupyter/jupyter_client/blob/8d760b1efeff2ac7aa2b4b75c44023e59c6945f8/jupyter_client/kernelspec.py#L205-L213

Configuring `kernel_dirs` based on the help text therefore produces the opposite resolution order.

### Note
I believe the documentation is wrong, not the code.

The default `kernel_dirs` comes from `jupyter_path("kernels")`, whose own docstring states that `$JUPYTER_PATH` is "always highest priority", followed by user, then environment, then system directories. First-wins is what makes a user kernelspec override a system one, which is the long-standing and expected behaviour.

Changing `find_kernel_specs()` to last-wins would invert that and silently make system kernelspecs shadow user ones.

## Impact

This is not merely cosmetic for downstream projects that set `kernel_dirs` explicitly to control which kernelspecs win. Following the documented order places the intended override directory last, where it is silently ignored, and users end up on a different kernel than the one they configured.

## Environment
- commit 8d760b1efeff2ac7aa2b4b75c44023e59c6945f8

aryansk:
## Problem

`KernelSpecManager.kernel_dirs` help text says "Later ones take priority over earlier", but `find_kernel_specs()` implements first-wins (the first directory containing a kernel name wins, via `if kname not in d`). Configuring `kernel_dirs` based on the help text therefore produces the opposite resolution order.

## Change

Update the `kernel_dirs` help text to describe the actual earlier-wins precedence, and add a regression test that ties the help text to the implemented behavior.

Closes #1137


krassowski:
Closing in favour of https://github.com/jupyter/jupyter_client/pull/1139. I don't think the test addition here is useful as proposed

